PR #4, opened 2024-12-06
feat: Config-Server/Shard relation

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

Add the managers and event handlers for config-server/shard relation.
Since the former model was using DatabaseRequire/DatabaseProvides, I'm keeping this for now, but making it already compatible with `data_interfaces` v40.
To note that we were using the interfaces but incorrectly, so it will be important to ensure that upgrades still work because there are now some more fields in the databag (which were left empty before because we were not using the whole power of data interfaces, only to be able to request secrets).

## 📑 Motivation and Context

Bring more features to the single kernel charm, in this case bring the sharding capabilities.

This has for now only been tested manually, I'll add some unit testing later.

## ✅ Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.10"
-poetry-core = "^1.9.0"
+poetry-core = "<2.0"
 ops = "~2.15.0"
 overrides = "^7.7.0"
 pydantic = "~2.9.0"
@@ -44,7 +44,7 @@ dacite = "^1.8.0"
 
 
 [build-system]
-requires = ["poetry-core>=1.9.0"]
+requires = ["poetry-core<2.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.group.charm-libs.dependencies]
--- a/single_kernel_mongo/abstract_charm.py
+++ b/single_kernel_mongo/abstract_charm.py
@@ -31,7 +31,7 @@ from ops.charm import CharmBase
 from single_kernel_mongo.config.literals import Substrates
 from single_kernel_mongo.config.relations import PeerRelationNames
 from single_kernel_mongo.core.operator import OperatorProtocol
-from single_kernel_mongo.core.structured_config import MongoConfigModel
+from single_kernel_mongo.core.structured_config import MongoConfigModel, MongoDBRoles
 from single_kernel_mongo.events.lifecycle import LifecycleEventsHandler
 from single_kernel_mongo.status import StatusManager
 
@@ -97,7 +97,6 @@ class AbstractMongoCharm(Generic[T, U], CharmBase):
 
     def on_leader_elected(self, _):
         """First leader elected handler."""
-        # FIXME: Check role status in databag first.
-
         # Sets the role in the databag.
-        self.operator.state.app_peer_data.role = self.parsed_config.role
+        if self.operator.state.app_peer_data.role == MongoDBRoles.UNKNOWN:
+            self.operator.state.app_peer_data.role = self.parsed_config.role
--- a/single_kernel_mongo/abstract_charm.py
+++ b/single_kernel_mongo/abstract_charm.py
@@ -97,6 +97,8 @@ class AbstractMongoCharm(Generic[T, U], CharmBase):
 
     def on_leader_elected(self, _):
         """First leader elected handler."""
-        # Sets the role in the databag.
+        # Sets the role in the databag: when the charm is first created, its
+        # role won't exist in the databag. We save it in the databag because we
+        # don't allow role changing yet.
         if self.operator.state.app_peer_data.role == MongoDBRoles.UNKNOWN:
             self.operator.state.app_peer_data.role = self.parsed_config.role
--- a/single_kernel_mongo/events/sharding.py
+++ b/single_kernel_mongo/events/sharding.py
@@ -22,6 +22,7 @@ from single_kernel_mongo.exceptions import (
     DeferrableFailedHookChecksError,
     FailedToUpdateCredentialsError,
     NonDeferrableFailedHookChecksError,
+    ShardAuthError,
     WaitingForCertificatesError,
     WaitingForSecretsError,
 )
@@ -75,7 +76,7 @@ class ConfigServerEventHandler(Object):
         is_leaving = isinstance(event, RelationBrokenEvent)
         try:
             self.manager.on_relation_event(event.relation, is_leaving)
-        except (DeferrableFailedHookChecksError, ServerSelectionTimeoutError) as e:
+        except (DeferrableFailedHookChecksError, ServerSelectionTimeoutError, ShardAuthError) as e:
             defer_event_with_info_log(logger, event, str(type(event)), str(e))
         except NonDeferrableFailedHookChecksError as e:
             logger.info(f"Skipping {str(type(event))}: {str(e)}")
--- a/single_kernel_mongo/events/sharding.py
+++ b/single_kernel_mongo/events/sharding.py
@@ -138,6 +138,7 @@ class ShardEventHandler(Object):
             DeferrableFailedHookChecksError,
             WaitingForSecretsError,
             WaitingForCertificatesError,
+            NotReadyError,
         ) as e:
             defer_event_with_info_log(logger, event, str(type(event)), str(e))
         except NonDeferrableFailedHookChecksError as e:
--- a/single_kernel_mongo/events/sharding.py
+++ b/single_kernel_mongo/events/sharding.py
@@ -73,7 +73,7 @@ class ConfigServerEventHandler(Object):
         """Handle relation changed and relation broken events."""
         is_leaving = isinstance(event, RelationBrokenEvent)
         try:
-            self.manager.on_relation_event(event.relation, is_leaving)
+            self.manager.reconcile_shards_for_relation(event.relation, is_leaving)
         except (DeferrableFailedHookChecksError, ServerSelectionTimeoutError, ShardAuthError) as e:
             defer_event_with_info_log(logger, event, str(type(event)), str(e))
         except NonDeferrableFailedHookChecksError as e:
@@ -82,7 +82,7 @@ class ConfigServerEventHandler(Object):
     def _on_database_requested(self, event: DatabaseRequestedEvent):
         """Relation joined events."""
         try:
-            self.manager.on_database_requested(event.relation)
+            self.manager.prepare_sharding_config(event.relation)
         except DeferrableFailedHookChecksError as e:
             logger.info("Skipping database requested event: hook checks did not pass.")
             defer_event_with_info_log(logger, event, str(type(event)), str(e))
@@ -125,11 +125,11 @@ class ShardEventHandler(Object):
         )
 
     def _on_relation_created(self, event: RelationCreatedEvent):
-        self.manager.relation_created()
+        self.manager.prepare_to_add_shard()
 
     def _on_database_created(self, event: DatabaseCreatedEvent):
         try:
-            self.manager.on_database_created(event.relation)
+            self.manager.synchronise_cluster_secrets(event.relation)
         except (
             DeferrableFailedHookChecksError,
             WaitingForSecretsError,
@@ -150,7 +150,7 @@ class ShardEventHandler(Object):
 
     def _on_relation_broken(self, event: RelationBrokenEvent):
         try:
-            self.manager.relation_broken(event.relation)
+            self.manager.drain_shard_from_cluster(event.relation)
         except DeferrableFailedHookChecksError as e:
             defer_event_with_info_log(logger, event, str(type(event)), str(e))
         except NonDeferrableFailedHookChecksError as e:
--- a/single_kernel_mongo/events/sharding.py
+++ b/single_kernel_mongo/events/sharding.py
@@ -1,0 +1,149 @@
+#!/usr/bin/env python3
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Manager for sharding and config server events."""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+from ops.charm import (
+    RelationBrokenEvent,
+    RelationChangedEvent,
+    RelationCreatedEvent,
+    RelationJoinedEvent,
+    SecretChangedEvent,
+)
+from ops.framework import Object
+
+from single_kernel_mongo.exceptions import (
+    DeferrableFailedHookChecksError,
+    FailedToUpdateCredentialsError,
+    NonDeferrableFailedHookChecksError,
+    WaitingForSecretsError,
+)
+from single_kernel_mongo.lib.charms.data_platform_libs.v0.data_interfaces import (
+    DatabaseProviderEventHandlers,
+    DatabaseRequirerEventHandlers,
+)
+from single_kernel_mongo.utils.event_helpers import defer_event_with_info_log
+from single_kernel_mongo.utils.mongo_connection import NotReadyError
+
+if TYPE_CHECKING:
+    from single_kernel_mongo.managers.mongodb_operator import MongoDBOperator
+
+
+logger = logging.getLogger(__name__)
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigServerEventHandler(Object):
+    """Event Handler for managing config server side events."""
+
+    def __init__(self, dependent: MongoDBOperator):
+        self.dependent = dependent
+        self.charm = self.dependent.charm
+        self.manager = self.dependent.config_server_manager
+        self.relation_name = self.manager.relation_name
+        super().__init__(parent=self.manager, key=dependent.config_server_manager.relation_name)
+
+        self.database_provider_events = DatabaseProviderEventHandlers(
+            self.charm, self.manager.data_interface
+        )
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_joined, self._on_relation_joined
+        )
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_departed,
+            self.dependent.check_relation_broken_or_scale_down,
+        )
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_changed, self._on_relation_event
+        )
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_broken, self._on_relation_event
+        )
+
+    def _on_relation_event(self, event: RelationChangedEvent):
+        """Handle relation changed and relation broken events."""
+        is_leaving = isinstance(event, RelationBrokenEvent)
+        try:
+            self.manager.on_relation_event(event.relation, is_leaving)
+        except DeferrableFailedHookChecksError as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
+
+    def _on_relation_joined(self, event: RelationJoinedEvent):
+        """Relation joined events."""
+        try:
+            self.manager.on_relation_joined(event.relation)
+        except DeferrableFailedHookChecksError as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
+
+
+class ShardEventHandler(Object):
+    """Event Handler for managing shard side events."""
+
+    def __init__(self, dependent: MongoDBOperator):
+        self.dependent = dependent
+        self.charm = self.dependent.charm
+        self.manager = self.dependent.shard_manager
+        self.relation_name = self.manager.relation_name
+        super().__init__(parent=self.manager, key=dependent.shard_manager.relation_name)
+
+        self.database_require_events = DatabaseRequirerEventHandlers(
+            self.charm, self.manager.data_requirer
+        )
+
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_created, self._on_relation_created
+        )
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_changed, self._on_relation_changed
+        )
+
+        self.framework.observe(
+            getattr(self.charm.on, "secret_changed"), self._handle_changed_secrets
+        )
+
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_departed,
+            self.dependent.check_relation_broken_or_scale_down,
+        )
+
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_broken, self._on_relation_broken
+        )
+
+    def _on_relation_created(self, event: RelationCreatedEvent):
+        self.manager.relation_created()
+
+    def _on_relation_changed(self, event: RelationChangedEvent):
+        try:
+            self.manager.relation_changed(event.relation)
+        except DeferrableFailedHookChecksError as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
+
+    def _handle_changed_secrets(self, event: SecretChangedEvent):
+        try:
+            self.manager.handle_secret_changed(event.secret.label or "")
+        except (NotReadyError, FailedToUpdateCredentialsError):
+            event.defer()
+        except WaitingForSecretsError:
+            logger.info("Missing secrets, ignoring")
+
+    def _on_relation_broken(self, event: RelationBrokenEvent):
+        try:
+            self.manager.relation_broken(event.relation)
+        except DeferrableFailedHookChecksError as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
--- a/single_kernel_mongo/events/sharding.py
+++ b/single_kernel_mongo/events/sharding.py
@@ -41,8 +41,6 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-logger = logging.getLogger(__name__)
-
 
 class ConfigServerEventHandler(Object):
     """Event Handler for managing config server side events."""
@@ -110,10 +108,7 @@ class ShardEventHandler(Object):
             self.charm.on[self.relation_name].relation_created, self._on_relation_created
         )
         self.framework.observe(
-            self.database_require_events.on.database_created, self._on_relation_changed
-        )
-        self.framework.observe(
-            self.charm.on[self.relation_name].relation_changed, self._on_relation_changed
+            self.database_require_events.on.database_created, self._on_database_created
         )
 
         self.framework.observe(
@@ -132,9 +127,9 @@ class ShardEventHandler(Object):
     def _on_relation_created(self, event: RelationCreatedEvent):
         self.manager.relation_created()
 
-    def _on_relation_changed(self, event: RelationChangedEvent | DatabaseCreatedEvent):
+    def _on_database_created(self, event: DatabaseCreatedEvent):
         try:
-            self.manager.relation_changed(event.relation)
+            self.manager.on_database_created(event.relation)
         except (
             DeferrableFailedHookChecksError,
             WaitingForSecretsError,
--- a/single_kernel_mongo/events/sharding.py
+++ b/single_kernel_mongo/events/sharding.py
@@ -13,7 +13,6 @@ from ops.charm import (
     RelationBrokenEvent,
     RelationChangedEvent,
     RelationCreatedEvent,
-    RelationJoinedEvent,
     SecretChangedEvent,
 )
 from ops.framework import Object
@@ -26,7 +25,9 @@ from single_kernel_mongo.exceptions import (
     WaitingForSecretsError,
 )
 from single_kernel_mongo.lib.charms.data_platform_libs.v0.data_interfaces import (
+    DatabaseCreatedEvent,
     DatabaseProviderEventHandlers,
+    DatabaseRequestedEvent,
     DatabaseRequirerEventHandlers,
 )
 from single_kernel_mongo.utils.event_helpers import defer_event_with_info_log
@@ -78,7 +79,7 @@ class ConfigServerEventHandler(Object):
         except NonDeferrableFailedHookChecksError as e:
             logger.info(f"Skipping {str(type(event))}: {str(e)}")
 
-    def _on_database_requested(self, event: RelationJoinedEvent):
+    def _on_database_requested(self, event: DatabaseRequestedEvent):
         """Relation joined events."""
         try:
             self.manager.on_database_requested(event.relation)
@@ -107,6 +108,9 @@ class ShardEventHandler(Object):
             self.charm.on[self.relation_name].relation_created, self._on_relation_created
         )
         self.framework.observe(
+            self.database_require_events.on.database_created, self._on_relation_changed
+        )
+        self.framework.observe(
             self.charm.on[self.relation_name].relation_changed, self._on_relation_changed
         )
 
@@ -126,7 +130,7 @@ class ShardEventHandler(Object):
     def _on_relation_created(self, event: RelationCreatedEvent):
         self.manager.relation_created()
 
-    def _on_relation_changed(self, event: RelationChangedEvent):
+    def _on_relation_changed(self, event: RelationChangedEvent | DatabaseCreatedEvent):
         try:
             self.manager.relation_changed(event.relation)
         except (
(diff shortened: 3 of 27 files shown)


single_kernel_mongo/abstract_charm.py
```diff
@@ -97,7 +97,6 @@ def on_install(self, _):
 
     def on_leader_elected(self, _):
         """First leader elected handler."""
-        # FIXME: Check role status in databag first.
-
         # Sets the role in the databag.
-        self.operator.state.app_peer_data.role = self.parsed_config.role
+        if self.operator.state.app_peer_data.role == MongoDBRoles.UNKNOWN:
```
MiaAltieri, line 103:
```suggestion
        # when the charm starts for the first time its role won't exist in the databag. We save the role in the databag
        # since role migration is not yet supported
        if self.operator.state.app_peer_data.role == MongoDBRoles.UNKNOWN:
```

single_kernel_mongo/events/sharding.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def _on_relation_event(self, event: RelationChangedEvent):
+        """Handle relation changed and relation broken events."""
+        is_leaving = isinstance(event, RelationBrokenEvent)
+        try:
+            self.manager.on_relation_event(event.relation, is_leaving)
+        except (DeferrableFailedHookChecksError, ServerSelectionTimeoutError, ShardAuthError) as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
```
MiaAltieri, line 80: i. love. this.

single_kernel_mongo/managers/config.py
```diff
@@ -100,7 +100,7 @@ def configure_and_restart(self):
                 self.workload.stop()
                 self.set_environment()
                 # Avoid restart errors on PBM.
-                time.sleep(2)
+                time.sleep(5)
```
MiaAltieri: newb q - why 5?  we could also do

time.sleep(PBM_RESTART_DELAY)

single_kernel_mongo/managers/config.py
```diff
@@ -280,7 +282,10 @@ def db_path_argument(self) -> list[str]:
     @property
     def role_parameter(self) -> list[str]:
         """The role parameter."""
-        match self.state.app_peer_data.role:
+        role = self.state.app_peer_data.role
+        if role == MongoDBRoles.UNKNOWN:  # First install we don't have the role in databag yet.
```
MiaAltieri: hmm, I noticed we also did this in the leader_elected event

it certainly doesn't hurt to have it in two places but not very clean

single_kernel_mongo/managers/mongo.py
```diff
@@ -404,6 +404,18 @@ def process_added_units(self) -> None:
                     raise NotReadyError
                 mongo.add_replset_member(member)
 
+    def get_draining_shards(self, config: MongoConfiguration | None = None) -> list[str]:
+        """Returns the shard that is currently draining."""
+        with MongoConnection(config or self.state.mongos_config) as mongo:
+            draining_shards = mongo.get_draining_shards()
+
+            # in theory, this should always be a list of one. But if something has gone wrong we
+            # should take note and log it
```
MiaAltieri, line 418: ty

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -561,7 +600,8 @@ def update_related_hosts(self):
         if self.state.is_role(MongoDBRoles.REPLICATION):
             for relation in self.state.client_relations:
                 self.mongo_manager.update_app_relation_data(relation)
-        # TODO: Update related hosts for config server , cluster.
+        self.config_server_manager.update_mongos_hosts()
+        # TODO: Update related hosts for cluster.
```
MiaAltieri, line 604: remove comment?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -756,3 +796,15 @@ def _initialise_replica_set(self):
...
+    def assert_proceed_on_broken_event(self, relation: Relation):
+        """Runs some checks on broken relation event."""
+        if not self.state.has_departed_run(relation.id):
+            raise DeferrableFailedHookChecksError(
+                "must wait for relation departed hook to decide if relation should be removed"
+            )
+
+        if self.state.is_scaling_down(relation.id):
+            raise NonDeferrableFailedHookChecksError(
+                "Relation broken event occurring during scale down, do not proceed to remove users."
```
MiaAltieri:
```suggestion
                "Relation broken event occurring during scale down, no need to proceed with broken event"
```

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -756,3 +796,15 @@ def _initialise_replica_set(self):
...
+
+    def assert_proceed_on_broken_event(self, relation: Relation):
+        """Runs some checks on broken relation event."""
+        if not self.state.has_departed_run(relation.id):
+            raise DeferrableFailedHookChecksError(
+                "must wait for relation departed hook to decide if relation should be removed"
+            )
+
+        if self.state.is_scaling_down(relation.id):
+            raise NonDeferrableFailedHookChecksError(
```
MiaAltieri, line 808: this is a benign scenario do we really want to raise?

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,704 @@
...
+        self.charm = dependent.charm
+        self.state = state
+        self.workload = workload
+        self.substrate = substrate
+        self.relation_name = relation_name
+        self.data_interface = DatabaseProviderData(
+            self.model, relation_name=self.relation_name.value
+        )
+
+    def on_database_requested(self, relation: Relation):
```
MiaAltieri: is there a reason for this name? Maybe it could be `prep_shard_config`

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,704 @@
...
+        int_tls_ca = self.state.tls.get_secret(internal=True, label_name=SECRET_CA_LABEL)
+        if int_tls_ca:
+            relation_data[ConfigServerKeys.int_ca_secret.value] = int_tls_ca
+
+        self.data_interface.update_relation_data(relation.id, relation_data)
+        self.data_interface.set_credentials(
+            relation.id, "unused", "unused"
+        )  # Triggers the database created event
+
+    def on_relation_event(self, relation: Relation, is_leaving: bool = False):
```
MiaAltieri: should this be called `on_relation_event`? I feel like it blurs the lines that we have between events going in the events dir and managers going in the managers dir. Perhaps instead it could be called `reconcile_shards`

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,704 @@
...
+    def add_shard(self, relation: Relation):
+        """Adds a shard to the cluster."""
+        shard_name = relation.app.name
+
+        hosts = []
+        for unit in relation.units:
+            if self.substrate == "k8s":
+                unit_name = unit.name.split("/")[0]
+                unit_id = unit.name.split("/")[1]
+                host_name = f"{unit_name}-{unit_id}.{unit_name}-endpoints"
```
MiaAltieri: if possbile I would love for things like `                host_name = f"{unit_name}-{unit_id}.{unit_name}-endpoints"`

to go in K8s lib. reasoning, we might want to universally change host names. I know Enrico already has some learnings on this and we might want to standardise across the team

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,704 @@
...
+                )
+            case _:
+                pass
+
+        if not self.is_ca_compatible():
+            raise DeferrableFailedHookChecksError(
+                "Shard is integrated to a different CA than the config server. Please use the same CA for all cluster components.",
+            )
+
+    def relation_created(self):
```
MiaAltieri: same here, event handler like name. Perhaps `prepare_to_add_shard`

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,704 @@
...
+        """Sets status and flags in relation data relevant to sharding."""
+        # if re-using an old shard, re-set flags.
+        self.state.unit_peer_data.drained = False
+        self.charm.status_manager.to_maintenance("Adding shard to config-server")
+
+    def on_database_created(self, relation: Relation, leaving: bool = False):
```
MiaAltieri: rename `sync_cluster_secrets`

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,704 @@
...
+            return
+
+        operator_password = self.state.shard_state.operator_password
+        backup_password = self.state.shard_state.backup_password
+
+        if not operator_password or not backup_password:
+            raise WaitingForSecretsError
+        self.sync_cluster_passwords(operator_password, backup_password)
+
+    def relation_broken(self, relation: Relation) -> None:
```
MiaAltieri: `drain_shard_from_cluster`

single_kernel_mongo/state/charm_state.py
```diff
@@ -287,6 +297,17 @@ def host_port(self) -> int:
             return MongoPorts.MONGOS_PORT
         return MongoPorts.MONGODB_PORT
 
+    @property
+    def shard_state(self):
+        """The shard state."""
+        return ConfigServerState(
```
MiaAltieri: `def shard_state` returns a `ConfigServerState`, which by looking at `ConfigServerState` holds the same information as a shard state, so logically it is not an issue, but I think `ConfigServerState` should be renamed to `ShardingComponentState`

also you're missing a return type here

single_kernel_mongo/state/unit_peer_state.py
```diff
@@ -99,3 +100,14 @@ def node_port(self) -> int:
         K8s-only.
         """
         return self.k8s.get_node_port(MongoPorts.MONGOS_PORT)
+
+    @property
+    def drained(self) -> bool:
+        """Is the shard drained."""
```
MiaAltieri: I believe the reason that this must be a unit state instead of a peer state is because we can then check if a shard is drained *without* receiving a `relation_changed_event`. If that is true would you mind adding that context to the doc strings in these functions?

single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -333,3 +357,261 @@ def is_any_sync(rs_status: dict[str, Any]) -> bool:
...
+                    f"Cannot move primary for database: {database_name}, new shard: {new_shard}",
+                    f"does not have enough space. {db_size} > {avail_space}",
+                )
+                logger.error(no_space_on_new_primary)
+                raise NotEnoughSpaceError(no_space_on_new_primary)
+
+            # From MongoDB Docs: After starting movePrimary, do not perform any read or write
+            # operations against any unsharded collection in that database until the command
+            # completes.
+            logger.info(
```
MiaAltieri:
```suggestion
            logger.warning(
```

single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -333,3 +357,261 @@ def is_any_sync(rs_status: dict[str, Any]) -> bool:
...
+                new_shard,
+            )
+
+    def get_db_size(self, database_name, primary_shard) -> int:
+        """Returns the size of a DB on a given shard in bytes."""
+        database = self.client[database_name]
+        db_stats = database.command("dbStats")
+
+        # sharded databases are spread across multiple shards, find the amount of storage used on
+        # the primary shard
```
MiaAltieri, line 581: I assume this function was copied from the original?

I am forgetful of the logic of why we only want to find the amount of storage used on the primary shard.

Regardless maybe the function name should be changed to get_db_size_on_primary_shard

single_kernel_mongo/managers/config.py
Gu1nness: I found that 2 seconds usually was still not enough to start reliably.
I'll create a literal for the 5 seconds

single_kernel_mongo/managers/config.py
Gu1nness: Ok so there was a reason for that when I did it, but due to other changes it has disappeared so I think I can go back to former behavior.
When I did it, the configuration of workloads was done during install event, so before the leader elected event.
I changed this for other reasons and now it's configured in the start event but still, it's a good thing to go back to the former behavior you're right.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Nope, not in this PR because there is still the updates for cluster (which are coming in a following PR :)

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: As you have noticed, I decided to go for a assert/raise pattern: if all checks are valid, we do nothing, if the check fails we raise and it's caught by the event handler which reacts according to the error raise: Non deferrable means we just return, deferrable means we log and defer, and uncaught exceptions will put into error (because it still needs to happen in some cases).
This makes a simple and straightforward pattern and avoid mixing booleans and error catching :)

single_kernel_mongo/managers/sharding.py
Gu1nness: Mainly because it handles a database requested event.
But you're right I'll rename it :)

single_kernel_mongo/managers/sharding.py
Gu1nness: Changing this to a better solution indeed :)

single_kernel_mongo/state/unit_peer_state.py
Gu1nness: Hum, I'm not sure what you mean by that, it was in the Unit peer databag in the former charm so I have to keep it that way I think and I don't see which difference you make between unit state and peer state in that case.
Anyway, updating our unit peer databag does not generate relation changed events IIRC and reading the drained value should not trigger a relation changed event because we are reading.
Maybe you are talking about writing the drained value in the databag?

single_kernel_mongo/utils/mongo_connection.py
Gu1nness: Yes it was copied.
We need to find the amount of storage used because we cannot move primary to a shard that doesn't have enough available space so we try to find the one which has the most available space and if it's still to small, we raise an error so the user can act accordingly.

I can rename it yes

single_kernel_mongo/state/unit_peer_state.py
MiaAltieri: My apologies my suggestion wasn't clear. Let me make that clearer. Please update this comment to say:
"""
Returns True if the shard is drained.

We check the unit databag rather than the app databag since a draining operation blocks all events from occurring. The unit databag allows us to update and check our draining status in the same event as the draining. Unlike the app databag which triggers requires a `relation_changed_event` to propogate.
"""

single_kernel_mongo/utils/mongo_connection.py
MiaAltieri: Can you please add the context of what you just described in the docstring of this function

single_kernel_mongo/abstract_charm.py
```diff
@@ -97,7 +97,8 @@ def on_install(self, _):
...
         """First leader elected handler."""
-        # FIXME: Check role status in databag first.
-
-        # Sets the role in the databag.
-        self.operator.state.app_peer_data.role = self.parsed_config.role
+        # Sets the role in the databag: when the charm is first created, its
+        # role won't exist in the databag. We save it in the databag because we
+        # don't allow role changing yet.
+        if self.operator.state.app_peer_data.role == MongoDBRoles.UNKNOWN:
+            self.operator.state.app_peer_data.role = self.parsed_config.role
```
Mehdi-Bendriss, line 104: Should this be in the charm or in the core or manager?

single_kernel_mongo/events/sharding.py
```diff
@@ -0,0 +1,157 @@
...
+        try:
+            self.manager.prepare_sharding_config(event.relation)
+        except DeferrableFailedHookChecksError as e:
+            logger.info("Skipping database requested event: hook checks did not pass.")
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
+
+
+class ShardEventHandler(Object):
```
Mehdi-Bendriss, line 93: missing docstrings on methods of this class

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -423,9 +443,21 @@ def on_storage_detaching(self) -> None:
...
         # have 0 members.
-        # TODO: When we have config server and shard managers.
-        # if self._is_removing_last_replica:
-        #    pass
+        if self.is_removing_last_replica:
+            if self.state.is_role(MongoDBRoles.CONFIG_SERVER) and self.state.config_server_relation:
+                current_shards = [
+                    relation.app.name for relation in self.state.config_server_relation
+                ]
+                early_removal_message = f"Cannot remoe config-server, still related to shards {', '.join(current_shards)}"
```
Mehdi-Bendriss:
```suggestion
                early_removal_message = f"Cannot remove config-server, still related to shards {', '.join(current_shards)}"
```

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,699 @@
...
+            )
+        relation_data = {
+            ConfigServerKeys.operator_password.value: self.state.get_user_password(OperatorUser),
+            ConfigServerKeys.backup_password.value: self.state.get_user_password(BackupUser),
+            ConfigServerKeys.key_file.value: self.state.get_keyfile(),
+            ConfigServerKeys.host.value: json.dumps(sorted(self.state.app_hosts)),
+        }
+
+        int_tls_ca = self.state.tls.get_secret(internal=True, label_name=SECRET_CA_LABEL)
+        if int_tls_ca:
```
Mehdi-Bendriss: nit:
```suggestion
        if int_tls_ca := self.state.tls.get_secret(internal=True, label_name=SECRET_CA_LABEL)
```

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,699 @@
...
+            logger.info("Waiting for secrets requested")
+            return
+
+        if not self.data_interface.fetch_relation_field(relation.id, "auth-updated") == "true":
+            logger.info(f"Waiting for shard {relation.app.name} to update its authentication")
+            return
+
+        try:
+            logger.info("Adding/Removing shards not present in cluster.")
+            match is_leaving:
```
Mehdi-Bendriss: in this particular occurrence, I think an if/else is clear and more concise

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,699 @@
...
+
+        uri = f"mongodb://{','.join(self.state.app_hosts)}"
+        if not self.dependent.mongo_manager.mongod_ready(uri):
+            return BlockedStatus("Internal mongos is not running.")
+
+        if not self.cluster_password_synced():
+            return WaitingStatus("Waiting to sync passwords across the cluster")
+
+        shard_draining = self.dependent.mongo_manager.get_draining_shards()
+        if shard_draining:
```
Mehdi-Bendriss: nit: walrus operator

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,699 @@
...
+            draining = ",".join(shard_draining)
+            return MaintenanceStatus(f"Draining shard {draining}")
+
+        if not self.state.config_server_relation:
+            return BlockedStatus("missing relation to shard(s)")
+
+        unreachable_shards = self.get_unreachable_shards()
+
+        if unreachable_shards:
```
Mehdi-Bendriss: nit: walrus operator

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,699 @@
...
+        )
+        try:
+            # check our ability to use connect to mongos
+            with MongoConnection(config) as mongos:
+                mongos.get_shard_members()
+            # check our ability to use connect to mongod
+            with MongoConnection(self.state.mongo_config) as mongod:
+                mongod.get_replset_status()
+        except OperationFailure as e:
+            if e.code in [13, 18]:
```
Mehdi-Bendriss: could you add comments or a log line explaining these code?

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,699 @@
...
+            # use a URI that is not dependent on the operator password, as we are not guaranteed
+            # that the shard has received the password yet.
+            uri = f"mongodb://{','.join(hosts)}"
+            if not self.dependent.mongo_manager.mongod_ready(uri):
+                unreachable_hosts.append(shard_name)
+
+        return unreachable_hosts
+
+
+class ShardManager(Object):
```
Mehdi-Bendriss, line 374: missing return type in most of the methods in this class

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,699 @@
...
+
+        while not drained:
+            try:
+                # no need to continuously check and abuse resources while shard is draining
+                time.sleep(60)
+                drained = self.drained(mongos_hosts, self.charm.app.name)
+                draining_status = (
+                    "Shard is still draining" if not drained else "Shard is fully drained."
+                )
+                self.charm.status_manager.to_maintenance("Draining shard from cluster.")
```
Mehdi-Bendriss, line 655: probably better to put this before the `while`?

single_kernel_mongo/state/app_peer_state.py
```diff
@@ -64,8 +65,10 @@ def role(self) -> MongoDBRoles:
 
         Either from the app databag or from the default from config.
         """
-        databag_role: str = str(self.relation_data.get(AppPeerDataKeys.role.value))
-        if not self.relation or not databag_role:
+        databag_role: str | None = self.relation_data.get(AppPeerDataKeys.role.value)
+        if not databag_role:
```
Mehdi-Bendriss: nit: walrus

single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -333,3 +357,261 @@ def is_any_sync(rs_status: dict[str, Any]) -> bool:
...
+        shard_hosts = ",".join(f"{host}:{shard_port}" for host in shard_hosts)
+        shard_url = f"{shard_name}/{shard_hosts}"
+        if shard_name in self.get_shard_members():
+            logger.info("Skipping adding shard %s, shard is already in cluster", shard_name)
+            return
+
+        logger.info("Adding shard %s", shard_name)
+        self.client.admin.command("addShard", shard_url)
+
+    def pre_remove_checks(self, shard_name: str):
```
Mehdi-Bendriss: Could you add the `shard` keyword in all shards related methods?

single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -333,3 +357,261 @@ def is_any_sync(rs_status: dict[str, Any]) -> bool:
...
+
+    def start_and_wait_for_balancer(self) -> None:
+        """Turns on the balancer and waits for it to be running.
+
+        Starting the balancer doesn't guarantee that is is running, wait until it starts up.
+
+        Raises:
+            BalancerNotEnabledError
+        """
+        self.client.admin.command("balancerStart")
```
Mehdi-Bendriss, line 442: Should we catch an exception?

single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -333,3 +357,261 @@ def is_any_sync(rs_status: dict[str, Any]) -> bool:
...
+        if remaining_chunks:
+            logger.info("Waiting for all chunks to be drained from %s.", shard_name)
+            raise NotDrainedError
+
+    def move_primary_after_draining_shard(self, shard_name: str):
+        """Move primary after the shard was drained and removed from the cluster."""
+        # MongoDB docs says to movePrimary only after all chunks have been drained from the shard.
+        logger.info("All chunks drained from shard: %s", shard_name)
+        databases_using_shard_as_primary = self.get_databases_for_shard(shard_name)
+        if databases_using_shard_as_primary:
```
Mehdi-Bendriss: nit: walrus

Mehdi-Bendriss (commented):
Thanks Neha! great work, no major concerns, I left some nit and a couple of questions

single_kernel_mongo/abstract_charm.py
Gu1nness: I think it need to be in the charm because this is true no matter which substrate or which kind (mongodb, mongos) charm you are using?

single_kernel_mongo/events/sharding.py
Gu1nness: Fixed

single_kernel_mongo/managers/sharding.py
Gu1nness: Indeed yes

single_kernel_mongo/managers/sharding.py
Gu1nness: Yes and I'll open a ticket to make all this better because for now this is ugly

single_kernel_mongo/managers/sharding.py
Gu1nness: Fixed

single_kernel_mongo/managers/sharding.py
Gu1nness: We can put it once there and then leave it also here because draining errors can self resolve.
So yeah I'll add it also before the while.

single_kernel_mongo/utils/mongo_connection.py
Gu1nness: Yes !

single_kernel_mongo/utils/mongo_connection.py
Gu1nness: There is no exception raised by the command in itself
If there are some errors, it can be OperationFailure or ConfigurationError (which is handled before in the call chain). I'll document that it can raise those errors here.

Mehdi-Bendriss (approved):

pyproject.toml
```diff
@@ -44,7 +44,7 @@ dacite = "^1.8.0"
 
 
 [build-system]
-requires = ["poetry-core>=1.9.0"]
+requires = ["poetry-core<2.0"]
```
MiaAltieri, line 47: in the future we will need to update these, but I am sure you already know that :)

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -491,9 +528,11 @@ def on_set_password_action(self, username: str, password: str | None = None) ->
...
             # Update and restart mongodb exporter.
             self.mongodb_exporter_config_manager.configure_and_restart()
-        # TODO: Rotate password.
-        if user in (OperatorUser, BackupUser):
-            pass
+        if user in (OperatorUser, BackupUser) and self.state.is_role(MongoDBRoles.CONFIG_SERVER):
+            self.config_server_manager.update_credentials(
+                user.password_key_name,
+                new_password,
+            )
```
MiaAltieri, line 535: does this function allow shads to rotate passwords? I think there are some scenrios which we don't want to do that [in the previous iteration we check the conidtions](https://github.com/canonical/mongodb-operator/blob/677a6620dd9d59e820793e4a7f658770019eef84/src/charm.py#L714)

single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -333,3 +357,267 @@ def is_any_sync(rs_status: dict[str, Any]) -> bool:
...
+        shard_hosts = ",".join(f"{host}:{shard_port}" for host in shard_hosts)
+        shard_url = f"{shard_name}/{shard_hosts}"
+        if shard_name in self.get_shard_members():
+            logger.info("Skipping adding shard %s, shard is already in cluster", shard_name)
+            return
+
+        logger.info("Adding shard %s", shard_name)
+        self.client.admin.command("addShard", shard_url)
+
+    def pre_remove_shard_checks(self, shard_name: str):
```
MiaAltieri: nit some functions here are missing return type

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,689 @@
...
+        self.charm = dependent.charm
+        self.state = state
+        self.workload = workload
+        self.substrate = substrate
+        self.relation_name = relation_name
+        self.data_interface = DatabaseProviderData(
+            self.model, relation_name=self.relation_name.value
+        )
+
+    def prepare_sharding_config(self, relation: Relation):
```
MiaAltieri: nit many functions missing return type

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,689 @@
...
+        self.data_interface.set_credentials(
+            relation.id, "unused", "unused"
+        )  # Triggers the database created event
+
+    def reconcile_shards_for_relation(self, relation: Relation, is_leaving: bool = False):
+        """Handles adding and removing shards.
+
+        Updating of shards is done automatically via MongoDB change-streams.
+        """
+        logger.info("Running Relation Changed hook.")
```
MiaAltieri: maybe this logger could be moved to event handler

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,689 @@
...
+        )
+        try:
+            # check our ability to use connect to mongos
+            with MongoConnection(config) as mongos:
+                mongos.get_shard_members()
+            # check our ability to use connect to mongod
+            with MongoConnection(self.state.mongo_config) as mongod:
+                mongod.get_replset_status()
+        except OperationFailure as e:
+            if e.code in [13, 18]:  # Unauthorized, Auth Failed
```
MiaAltieri, line 338: I've seen these numbers used elsewhere - it might be nice for these to be constants but it isn't necessary

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,689 @@
...
+
+        operator_password = self.state.shard_state.operator_password
+        backup_password = self.state.shard_state.backup_password
+        if not operator_password or not backup_password:
+            raise WaitingForSecretsError("Missing operator password or backup password")
+
+        self.sync_cluster_passwords(operator_password, backup_password)
+
+        # We have updated our auth, config-server can add the shard.
+        self.data_requirer.update_relation_data(relation.id, {"auth-updated": "true"})
```
MiaAltieri, line 491: interesting

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,689 @@
...
+    def drain_shard_from_cluster(self, relation: Relation) -> None:
+        """Waits for the shard to be fully drained from the cluster."""
+        self.assert_pass_hook_checks(relation, is_leaving=True)
+
+        self.charm.status_manager.to_maintenance("Draining shard from cluster.")
+
+        mongos_hosts = self.state.app_peer_data.mongos_hosts
+
+        self.wait_for_draining(mongos_hosts)
+
```
MiaAltieri, line 534: is this whitespace on purpose?

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,689 @@
...
+                            "Failed changing the password: Not all members healthy or finished initial sync."
+                        )
+                        raise
+                    except PyMongoError as e:
+                        logger.error(f"Failed changing the password: {e}")
+                        raise
+        self.state.set_user_password(user, new_password)
+
+    def _should_request_new_certs(self) -> bool:
+        """Returns if the shard has already requested the certificates for internal-membership."""
```
MiaAltieri:
```suggestion
        """
        Returns if the shard has already requested the certificates for internal-membership.
        
        MongoDB specific: sharded components must have the same subject names in their certs. 
        """
```

pyproject.toml
Gu1nness: Yes I know, I'll do it when everything is merged!
I'll take inspiration on your work on the current charms to do it, but it was easier for now to keep it <2.0

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: The checks are run in the event handler.
You can have a look at `assert_pass_password_hook_checks` in the `mongodb_operator` to see that it does the same checks as it used to.

single_kernel_mongo/managers/sharding.py
Gu1nness: It even should be removed completely, I just used it as debug log tbh.
I'll remove it

single_kernel_mongo/managers/sharding.py
Gu1nness: Mehdi asked the same in another PR and I'll bring an enum of the error codes that we handle in a future PR.

single_kernel_mongo/managers/sharding.py
Gu1nness: I'm trying to compose with the former charm while moving towards a correct use of the data interfaces that we're using (and it's hard).
I'm not fully satisfied about it but it works good enough

single_kernel_mongo/managers/sharding.py
Gu1nness: I don't see any whitespace in this strings?
I'll run the formatter again which removes whitespaces.

MiaAltieri (approved):

PR #5, opened 2024-12-11
feat: Mongos Operator

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

* Mongos operator, completing the MongoDB Operator
* Unification of the two operators
* Add support for mongos VM client integration in the `DatabaseEventHandler.

## 📑 Motivation and Context

 * Be able to deploy mongos charms
 * There are many events which call the same method, this is to be expected as there are way less event handled and many are handled the same way in the peer relationships.
 * Still needs the cluster manager and events to be functional.

None yet, need to be added.

## ✅ Checklist


single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+        self.handle_licenses()
+        # start hooks are fired before relation hooks and `mongos` requires a config-server in
+        # order to start. Wait to receive config-server info from the relation event before
+        # starting `mongos` daemon
+        self.charm.status_manager.to_blocked("Missing relation to config-server.")
+
+    @override
+    def on_secret_changed(self, secret_label: str, secret_id: str) -> None:
+        # Nothing happens in this handler for mongos operators
+        pass
```
MiaAltieri:
```suggestion
        """Nothing happens in this handler for mongos operators."""
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+    @override
+    def on_config_changed(self) -> None:
+        if self.substrate == Substrates.K8S:
+            if self.config.expose_external == ExposeExternal.UNKNOWN:
+                logger.error(
+                    "External configuration: %s for expose-external is not valid, should be one of: %s",
+                    self.charm.config["expose-external"],
+                    "['nodeport', 'none']",
+                )
+                self.charm.status_manager.to_blocked("Config option for expose-external not valid.")
```
MiaAltieri, line 152:
```suggestion
                self.charm.status_manager.to_blocked("Config option for expose-external not valid.")
                return
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+            self.update_external_services()
+
+            self.tls_manager.update_tls_sans()
+            self.share_connection_info()
+
+    @override
+    def on_storage_attached(self) -> None:
+        # Nothing happens in this handler for mongos operators
+        pass
```
MiaAltieri:
```suggestion
        """Nothing happens in this handler for mongos operators."""
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+
+    @override
+    def on_storage_detaching(self) -> None:
+        # Nothing happens in this handler for mongos operators
```
MiaAltieri: same suggestion as above

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+    def on_update_status(self) -> None:
+        if self.substrate == Substrates.K8S:
+            if self.config.expose_external == ExposeExternal.UNKNOWN:
+                logger.error(
+                    "External configuration: %s for expose-external is not valid, should be one of: %s",
+                    self.charm.config["expose-external"],
+                    "['nodeport', 'none']",
+                )
+                self.charm.status_manager.to_blocked("Config option for expose-external not valid.")
+                return
```
MiaAltieri, line 195:
```suggestion
        if self.substrate == Substrates.K8S and self.config.expose_external == ExposeExternal.UNKNOWN:
                logger.error(
                    "External configuration: %s for expose-external is not valid, should be one of: %s",
                    self.charm.config["expose-external"],
                    "['nodeport', 'none']",
                )
                self.charm.status_manager.to_blocked("Config option for expose-external not valid.")
                return
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+            self.charm.status_manager.to_blocked("Missing relation to config-server.")
+            return
+
+        # TODO : Check TLS status
+
+        if not self.is_mongos_running():
+            logger.info("mongos has not started yet")
+            self.charm.status_manager.to_waiting("Waiting for mongos to start.")
+            return
+
```
MiaAltieri, line 209:
```suggestion

# in K8s mongos charms which are exposed externally it is possible for the node port to change. This can invalidate our current certificates.
# when this happens we do not receive any notifications from Juju so we must monitor it and update our sans as necessary
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+        if self.substrate == Substrates.K8S:
+            self.tls_manager.update_tls_sans()
```
MiaAltieri, line 217: I believe it is also necessary to update connection info

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+                )
+            raise
+
+    def _share_configuration(self):
+        """Actually shares the configuration according to the substrate."""
+        match self.substrate:
+            case Substrates.VM:
+                if not self.state.mongos_config.password or not self.state.mongos_config.username:
+                    return
+                for relation in self.state.client_relations:
```
MiaAltieri, line 297: i believe that VM can only have one client relation

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+
+        if new_extra_user_roles != self.state.app_peer_data.extra_user_roles:
+            self.state.app_peer_data.extra_user_roles = new_extra_user_roles
+            if self.state.mongos_cluster_relation:
+                self.state.cluster.extra_user_roles = new_extra_user_roles
+
+        self.state.app_peer_data.external_connectivity = external_connectivity
+
+        if external_connectivity:
+            self.charm.unit.open_port("tcp", MongoPorts.MONGOS_PORT)
```
MiaAltieri, line 338: hmm this seems like a side effect to the function which is supposed to only share credentials, is there a better place to put this?

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+
+    # BEGIN: Helpers
+    def update_external_services(self):
+        """Updates the external service if necessary."""
+        if self.substrate == Substrates.K8S:
```
MiaAltieri:
```suggestion
        if self.substrate == Substrates.VM:
              return
        
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+                self.state.cluster.extra_user_roles = new_extra_user_roles
+
+        self.state.app_peer_data.external_connectivity = external_connectivity
+
+        if external_connectivity:
+            self.charm.unit.open_port("tcp", MongoPorts.MONGOS_PORT)
+
+    # BEGIN: Helpers
+    def update_external_services(self):
+        """Updates the external service if necessary."""
```
MiaAltieri:
```suggestion
        """Updates the external k8s service if necessary."""
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,318 @@
...
+                data_interface.fetch_relation_field(
+                    relation.id,
+                    "extra-user-roles",
+                )
+                or "[]"
+            )
+        )
+        external_connectivity = json.loads(
+            data_interface.fetch_relation_field(relation.id, "external-node-conectivity") or "false"
+        )
```
MiaAltieri: can we retrieve the value for VM for `"external-node-conectivity"` in the state rather than the app databag?

single_kernel_mongo/state/app_peer_state.py
```diff
@@ -165,3 +173,40 @@ def external_connectivity(self, value: bool) -> None:
...
+    @property
+    def expose_external(self) -> ExposeExternal:
+        """The value of the expose-external flag."""
+        if not self.relation:
+            return ExposeExternal.UNKNOWN
+        return ExposeExternal(self.relation_data.get(AppPeerDataKeys.expose_external.value, ""))
+
+    @expose_external.setter
+    def expose_external(self, value: ExposeExternal):
+        self.update({AppPeerDataKeys.expose_external.value: value.value})
```
MiaAltieri, line 211: is this property for both VM + K8s?

seems like it is just for k8s, but I would love it to work for VM too if possible

single_kernel_mongo/state/charm_state.py
```diff
@@ -119,6 +117,8 @@ def peers_units(self) -> set[Unit]:
     @property
     def client_relations(self) -> set[Relation]:
         """The set of client relations."""
```
MiaAltieri:
```suggestion
        """The set of client relations.
        
        Client relations exist on two separate interfaces, one for sharding and one for replication
        """
```

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Fixed

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Nice catch :eyes:

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: I'm not sure this is a good idea because all this code should only run on K8S and there's some code that is still to be run in case of a valid "expose-external" value (We don't handle config changed events in the mongos VM charm because there's no need for it).
I'd rather keep it the way it is now :)

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: I don't think so: `update_tls_sans` will request new certificates which will trigger some more events that will eventually update the connection info when we have the new certificates.

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: I think it's more the function that is poorly named.
It's mainly a proxy function between the client and the cluster relation and maybe that's how it should be named: `proxy_information_to_client`.
As this function is used only once, I could create an `open_port` function but this would duplicate all the code that gets the external connectivity value and such.
I think for now I'll rename the function.

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Hum, I'm not sure tbh.
It's two faces of the same coin but I'd rather "if X: do something" than `if not X: return / else: do something`

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Oh no not here!
This is exactly the moment where it's stored in the app databag.
This function is called when the cluster relation receives a ChangedEvent which will happen when it shares the credentials with mongos (This is VM only).
So in this case we specifically want to store the value returned by the cluster in the databag and the use it.
I agree however that the interface is not very clear here and we do need a better interface which should be stored in the Charm State (and maybe that's what you meant?)

I'll give a try to something and we'll see if that works.

single_kernel_mongo/state/app_peer_state.py
Gu1nness: The value is stored anyway, (using the default value of `expose-external` from the config) and accessible, it can just not be changed by a config changed event in the case of VM (where it's not needed anyway because we are subordinate)

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: I think this is something that I changed in a following PR, if that's not the case I do know that I changed some things over there so let's wait just a bit to change that but we'll definitely use the state to access this!

single_kernel_mongo/core/structured_config.py
```diff
@@ -58,6 +69,14 @@ class MongoConfigModel(BaseConfigModel):
     role: SerializeLiteralAsStr[MongoDBRoles]
     auto_delete: bool = Field(default=False, alias="auto-delete")
 
+    @field_validator("expose_external", mode="before")
+    @classmethod
+    def invalid_expose_to_unknown(cls, v: str) -> ExposeExternal:
+        """If the value is neither none or nodeport, returns unknown."""
+        if v not in ExposeExternal.valid_roles():
+            return ExposeExternal.UNKNOWN
```
Mehdi-Bendriss, line 77: Should we not let the validation error propagate to avoid masking it? or do we handle `UNKNOWN` elsewhere?

single_kernel_mongo/events/database.py
```diff
@@ -68,6 +69,11 @@ def _on_relation_event(self, event: RelationEvent):
         # TODO : Handle the Mongos VM case
         # (https://github.com/canonical/mongos-operator/blob/6/edge/lib/charms/mongos/v0/mongos_client_interface.py)
         # for interface unification.
+        if (
+            self.dependent.substrate == Substrates.VM
+            and self.relation_name == RelationNames.MONGOS_PROXY
+        ):
+            self.dependent.proxy_information_to_client_and_handle_connectivity(event.relation)  # type: ignore[attr-defined]
```
Mehdi-Bendriss: I feel like this method could use some renaming :D

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,324 @@
...
+            self.substrate,
+        )
+        pod_name = self.model.unit.name.replace("/", "-")
+        self.k8s = K8sManager(pod_name, self.model.name)
+
+        self.tls_events = TLSEventsHandler(self)
+        self.client_events = DatabaseEventsHandler(self, RelationNames.MONGOS_PROXY)
+
+    @property
+    def config(self):
```
Mehdi-Bendriss: nit: missing return type

single_kernel_mongo/core/structured_config.py
```diff
@@ -58,6 +69,14 @@ class MongoConfigModel(BaseConfigModel):
     role: SerializeLiteralAsStr[MongoDBRoles]
     auto_delete: bool = Field(default=False, alias="auto-delete")
 
+    @field_validator("expose_external", mode="before")
+    @classmethod
```
Mehdi-Bendriss, line 73: Redundant / safe to delete, as the `field_validator` already returns a `classmethod`

single_kernel_mongo/core/structured_config.py
```diff
@@ -67,11 +86,19 @@ class MongoDBCharmConfig(MongoConfigModel):
 
     role: SerializeLiteralAsStr[MongoDBRoles] = Field(default=MongoDBRoles.REPLICATION)
 
+    @field_validator("role", mode="before")
+    @classmethod
```
Mehdi-Bendriss, line 90: similar comment, safe to delete

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -7,10 +7,10 @@
 from __future__ import annotations
 
 import logging
-from pathlib import Path
+
+# from pathlib import Path
```
Mehdi-Bendriss: should we delete?

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,324 @@
...
+        """Returns the actual config."""
+        return self.charm.parsed_config
+
+    @override
+    def on_install(self) -> None:
```
Mehdi-Bendriss, line 104: Could you explain why we have `on_install` divided between the `AbstractCharm` and `LifecycleEventsHandler`?

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,324 @@
...
+        if not self.workload.workload_present:
```
Mehdi-Bendriss, line 110: will we ever hit this path in the case of `K8s`? I which case, what happens to the subsequent calls?

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,324 @@
...
+)
+from single_kernel_mongo.workload.mongos_workload import MongosWorkload
+
+if TYPE_CHECKING:
+    from single_kernel_mongo.abstract_charm import AbstractMongoCharm  # pragma: nocover
+logger = logging.getLogger(__name__)
+
+
+@final
+class MongosOperator(OperatorProtocol, Object):
```
Mehdi-Bendriss, line 51: missing docstrings in multiple methods of this class

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,324 @@
...
+            raise MissingConfigServerError()
+
+        self.mongos_config_manager.set_environment()
+        self.workload.start()
+
+    @override
+    def is_relation_feasible(self, name: str) -> bool:
+        if name != RelationNames.MONGOS_PROXY:
+            return False
+        return True
```
Mehdi-Bendriss: nit: 
```suggestion
        return name == RelationNames.MONGOS_PROXY
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,324 @@
...
+
+    # BEGIN: Helpers
+    def update_external_services(self):
+        """Updates the kubernetes external service if necessary."""
+        if self.substrate == Substrates.K8S:
+            if self.config.expose_external == ExposeExternal.NODEPORT:
+                service = self.k8s.build_node_port_services(str(MongoPorts.MONGOS_PORT))
+                self.k8s.apply_service(service)
+            else:
+                self.k8s.delete_service()
```
Mehdi-Bendriss: should we delete the service if the value is `UNKNOWN`?

single_kernel_mongo/state/charm_state.py
```diff
@@ -118,7 +116,14 @@ def peers_units(self) -> set[Unit]:
...
     @property
     def client_relations(self) -> set[Relation]:
-        """The set of client relations."""
+        """The set of client relations.
+
+        Client relations exist on two separate interfaces, one for sharding,
+        which is exposed for mongos charms, and one for replication which is
+        exposed for mongodb charms.
+        """
+        if self.charm_role.name == KindEnum.MONGOS:
```
Mehdi-Bendriss, line 125: I believe we should find a better name for this Enum and getting rid of the `Enum` suffix.

Mehdi-Bendriss (commented):
Thanks Neha, great work! Left some comments and questions

single_kernel_mongo/events/database.py
Gu1nness: It was renamed from `share_credentials` and I don't have a better name idea, if you have any, be my guest ;)

single_kernel_mongo/core/structured_config.py
Gu1nness: This is handled elsewhere, in the config-changed event

single_kernel_mongo/core/structured_config.py
Gu1nness: This is advised by [the documentation](https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.field_validator) to specify it as class method (and on top of that if I remove it, the linters complain that it should be named "self" so it doesn't seem to understand it as a classmethod

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Yes, completely forgot to do it.

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: The part in `AbstractCharm` is the common part to all charms (install charmed mongodb) and the rest is kind specific (depends if we are mongodb or mongos)

It's purely a matter of preference between everything in the managers or split between both.
If everything is in the managers, then there's a bit more code duplication but not much

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: This can happen yes because `workload_present` in the case of k8s is `container.can_connect`.
The events handler does defer in this case as we need the container to be present to set the workload version and truncate the files

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: I'll add it, I tried no to because it's already defined in the Operator Protocol but sure :)

single_kernel_mongo/core/structured_config.py
Gu1nness: Similar response

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: You're correct!

single_kernel_mongo/state/charm_state.py
Gu1nness: You're probably right.
I'm trying to think about something that is not "Role" because role means replication vs shard vs config-server vs mongos.
Maybe just `Kind`?

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -96,19 +96,29 @@ def __init__(self, charm: AbstractMongoCharm):
...
         if not self.workload.workload_present:
             raise ContainerNotReadyError
         self.charm.unit.set_workload_version(self.workload.get_version())
         self.mongos_config_manager.set_environment()
 
     @override
     def on_start(self) -> None:
+        """For this case, we don't start any service.
+
+        Mongos charms need to be integrated to its config server before
```
MiaAltieri, line 119: 😍 love this docstring

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -125,6 +135,11 @@ def on_secret_changed(self, secret_label: str, secret_id: str) -> None:
 
     @override
     def on_config_changed(self) -> None:
+        """We check the expose-external value as it can be invalid.
+
+        And then, update external services, update TLS certificates and share
+        connection information with client if it changed.
+        """
```
MiaAltieri:
```
Handle configurations for expose-external.

It is necessary to check that the option is valid, and if it has changed we must update external services, update TLS certificates and share
        connection information with client. This is because when we change our connectivity we update the IP address of mongos. 
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -151,11 +166,17 @@ def on_storage_detaching(self) -> None:
 
     @override
     def on_leader_elected(self) -> None:
-        # Just forward the call, this is for simplicity and typing.
+        """Just forward the call, this is for simplicity and typing."""
```
MiaAltieri:
```
 """
Just forward the call, this is for simplicity and typing.

Leader elected events indicate that a unit may have been removed or lost connectivity. In these cases the hosts can be updated and we must share the most up to date information with the hosts.
""""

File: single_kernel_mongo/managers/mongos_operator.py [hunk 30]
```diff
@@ -190,14 +211,17 @@ def on_update_status(self) -> None:
...
 
     @override
     def on_relation_changed(self) -> None:
+        """Any relation event will just share the connection with the client."""
         self.share_connection_info()
 
     @override
     def on_relation_departed(self, departing_unit: Unit | None) -> None:
+        """Any relation event will just share the connection with the client."""
         self.share_connection_info()
```
MiaAltieri: is this true for departed?

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -206,14 +230,18 @@ def on_stop(self) -> None:
 
     @override
     def start_charm_services(self) -> None:
+        """Star the charm services."""
+        self.mongos_config_manager.set_environment()
         self.workload.start()
 
     @override
     def stop_charm_services(self) -> None:
+        """Star the charm services."""
```
MiaAltieri, line 248: nit `start`

single_kernel_mongo/managers/mongos_operator.py
MiaAltieri: I think it helps the indentation, but up to you

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: This is true for departed as well: Departed here is for the peer relation.

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Fixed

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Yes but this is easier to write and will do at most one round of the loop :)

single_kernel_mongo/state/charm_state.py
Gu1nness: Done in the status PR : https://github.com/canonical/mongo-single-kernel-library/commit/b31de6b60596ca9548425cc5b77607ff99dd9d7a

single_kernel_mongo/managers/mongos_operator.py
Mehdi-Bendriss: Yes, but I thought there is no `_install` hook on k8s, is there?

single_kernel_mongo/events/database.py
Mehdi-Bendriss: what about: `update_proxy_connection`?

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: There is an install hook on k8s as well, it's almost never used but can be (for example kafka uses it)

single_kernel_mongo/events/database.py
Gu1nness: That's a good idea!

single_kernel_mongo/managers/mongos_operator.py
MiaAltieri: if thats the case can you please add a comment so it is clear to future developers that e need to handle this case

single_kernel_mongo/managers/mongos_operator.py
MiaAltieri: Can you add a comment here so it is clear for future developers

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -326,17 +333,23 @@ def proxy_information_to_client_and_handle_connectivity(self, relation: Relation
 
     # BEGIN: Helpers
     def update_external_services(self):
```
MiaAltieri: sorry nit here, but can we rename to `update_k8s_external_services`, I say this because the mongos vm service can be made external, but this function is purely for k8s

Mehdi-Bendriss (approved):

MiaAltieri (approved):

PR #6, opened 2024-12-12
feat: Cluster events and managers

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

* Integrate Cluster management on config server and mongos
* Integrate Cluster events on config server and mongos
* Small bug fixes
* `assert_proceed_on_broken_event` is in the `OperatorProtocol` now.

## 📑 Motivation and Context

* One more step towards feature parity with former charms.
* Be able to test the mongos operator

Manually testing rn.

## ✅ Checklist


single_kernel_mongo/core/operator.py
```diff
@@ -148,6 +152,18 @@ def is_relation_feasible(self, name: str) -> bool:
...
+    def assert_proceed_on_broken_event(self, relation: Relation):
+        """Runs some checks on broken relation event."""
+        if not self.state.has_departed_run(relation.id):
+            raise DeferrableFailedHookChecksError(
+                "must wait for relation departed hook to decide if relation should be removed"
+            )
+
+        if self.state.is_scaling_down(relation.id):
+            raise NonDeferrableFailedHookChecksError(
+                "Relation broken event occurring during scale down, do not proceed to remove users."
```
MiaAltieri, line 164: update log

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,351 @@
...
+            raise NonDeferrableFailedHookChecksError("Not leader")
+
+        if self.state.upgrade_in_progress:
+            raise DeferrableFailedHookChecksError(
+                "Processing mongos applications is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
+            )
+
+    def is_valid_mongos_integration(self):
+        """Returns True if the integration to mongos is valid."""
+        return self.state.is_role(MongoDBRoles.CONFIG_SERVER) or not self.state.cluster_relations
```
MiaAltieri, line 85: I like the simplification here, but the logic was hard to follow ([the original was pretty straightforwarad](https://github.com/canonical/mongodb-operator/blob/62dd700d4e9ddbe25debd2d5bd51e7b4966331ef/lib/charms/mongodb/v0/config_server_interface.py#L113)) maybe a comment can be added

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def on_database_requested(self, relation: Relation):
```
MiaAltieri: same comments as previous PR on the function names

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,351 @@
...
+        if not self.data_interface.fetch_relation_field(relation.id, "database"):
+            logger.info("Database Requested has not run yet, skipping.")
+            return
+
+        self.on_database_requested(relation)
+
+    def on_relation_broken(self, relation: Relation) -> None:
+        """Handles the relation broken event.
+
+        Needs to decide what we do based on the situation.
```
MiaAltieri: improve doc string here

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,351 @@
...
+        for relation in self.state.cluster_relations:
+            self.data_interface.update_relation_data(
+                relation.id,
+                {
+                    ClusterStateKeys.config_server_db.value: config_server_db,
+                },
+            )
+
+    def update_ca_secret(self, new_ca: str | None) -> None:
+        """Updates the new CA for all related shards."""
```
MiaAltieri:
```suggestion
        """Updates the new CA for all related mongos."""
```

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,351 @@
...
+        self.state.secrets.set(AppPeerDataKeys.password.value, password, Scope.APP)
+
+    def relation_changed(self) -> None:
+        """Start/restarts mongos with config server information."""
+        self.assert_pass_hook_checks()
+        key_file_contents = self.state.cluster.keyfile
+        config_server_db_uri = self.state.cluster.config_server_uri
+
+        if not key_file_contents or not config_server_db_uri:
+            raise WaitingForSecretsError("Waiting for keyfile or config server db uri")
```
MiaAltieri, line 244: do we really want to raise this error instead of one of the Deferrable errors?

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,351 @@
...
+        if not self.charm.unit.is_leader():
+            return
+
+        logger.info("Cleaning database and user removed for mongos application")
+        self.state.secrets.remove(Scope.APP, AppPeerDataKeys.username.value)
+        self.state.secrets.remove(Scope.APP, AppPeerDataKeys.password.value)
+
+    def update_users(self):
+        """Updates users after being initialised."""
+        if self.substrate != Substrates.K8S:
```
MiaAltieri, line 289: please add a comment explaining this logic so it will be obvious the differences between k8s vs VM management in mongo users

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,351 @@
...
+        except PyMongoError:
+            raise DeferrableError("Failed to add users on mongos-k8s router.")
+
+    def remove_users(self, relation: Relation):
+        """Handles the removal of all client mongos-k8s users and the mongos-k8s admin user.
+
+        Raises:
+            PyMongoError
+        """
+        if self.substrate != Substrates.K8S:
```
MiaAltieri, line 309: same here

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,351 @@
...
+                data_interface.fetch_relation_field(relation.id, "data") or "{}"
+            )["secret-user"]
+
+            user_secrets = self.charm.model.get_secret(id=secret_id)
+            user_secrets.remove_all_revisions()
+            user_secrets.get_content(refresh=True)
+            relation.data[self.charm.app].clear()
+
+    def is_ca_compatible(self) -> bool:
+        """Returns true if both the mongos and the config-server use the same CA."""
```
MiaAltieri:
```suggestion
        """Returns true if both the mongos and the config-server use the same CA, a requirement for sharded clusters."""
```

single_kernel_mongo/managers/mongo.py
```diff
@@ -326,24 +326,29 @@ def update_app_relation_data_for_config(self, relation: Relation, config: MongoC
...
+        if config.uri != uris:
+            data_interface.set_uris(
+                relation.id,
+                config.uri,
+            )
+        if config.database != database:
+            data_interface.set_database(
+                relation.id,
+                config.database,
+            )
```
MiaAltieri, line 354: whats wrong with setting them no matter what (i.e. removing the if statement). IIRC the relation changed event only gets triggered if the contents are different

single_kernel_mongo/managers/mongo.py
```diff
@@ -441,11 +446,36 @@ def get_draining_shards(self, config: MongoConfiguration | None = None) -> list[
...
             return draining_shards
 
-    # Keep for memory for now.
-    # def get_relation_name(self):
-    #    """Returns the name of the relation to use."""
-    #    if self.state.is_role(MongoDBRoles.CONFIG_SERVER):
-    #        return RelationNames.CLUSTER
-    #    if self.state.is_role(MongoDBRoles.MONGOS):
-    #        return RelationNames.MONGOS_PROXY
-    #    return RelationNames.DATABASE
```
MiaAltieri, line 443: is this really no longer needed?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -12,7 +12,7 @@
 from typing import TYPE_CHECKING, final
 
 from ops.framework import Object
```
MiaAltieri, line 12: in the main PR for single kernel charm I had a comment requesting that `mongodb_operator` and `mongos_operator` be moved to their own directory `~/charms/` or even  `~/single_kernel_mongo/charms/` - just pointing that out here as to not forget

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -171,13 +176,18 @@ def on_update_status(self) -> None:
...
+            logger.info(f"Invalid TLS integration: {status.message}")
+            self.charm.status_manager.set_and_share_status(status)
+            return
 
         if not self.is_mongos_running():
             logger.info("mongos has not started yet")
             self.charm.status_manager.to_waiting("Waiting for mongos to start.")
             return
 
+        self.share_connection_info()
```
MiaAltieri, line 220: comment above this please

single_kernel_mongo/state/charm_state.py
```diff
@@ -278,33 +306,63 @@ def peer_units_data_interfaces(self) -> dict[Unit, DataPeerOtherUnitData]:
             for unit in self.peers_units
         }
 
+    @property
+    def formatted_socket_path(self) -> str:
+        """URL encoded socket path."""
```
MiaAltieri: can you embed some of our expert knowledge here. i.e. this is primarily used for mongos subordinate VM charm in order to improve latency

single_kernel_mongo/state/charm_state.py
```diff
@@ -278,33 +306,63 @@ def peer_units_data_interfaces(self) -> dict[Unit, DataPeerOtherUnitData]:
...
+        if (
+            self.substrate == Substrates.VM
+            and self.charm_role.name == KindEnum.MONGOS
+            and not self.app_peer_data.external_connectivity
+        ):
+            return {self.formatted_socket_path}
         return {unit.host for unit in self.units}
 
     @property
     def internal_hosts(self) -> set[str]:
```
MiaAltieri, line 358: QQ is this function and one above really needed? is it possible to combine them?

single_kernel_mongo/core/operator.py
Gu1nness: What do you mean by that?

single_kernel_mongo/managers/cluster.py
Gu1nness: Nice catch!

single_kernel_mongo/managers/cluster.py
Gu1nness: Hum this is a bit more precise IMHO in case we want to keep track of the code in case of errors.
It could be a DeferrableError indeed, so I don't have a strong opinion here.

single_kernel_mongo/managers/mongo.py
Gu1nness: I'd love to be sure but I'm not.
On top of that, some of those are secrets so setting it will trigger some secret changed event and we now that our events end up in secret changed storms all the time (especially before because we were not sorting hosts in any way so uris would constantly change.

This is mainly to reduce the load on juju and reduce the amount of events and secret changed events.

single_kernel_mongo/managers/mongo.py
Gu1nness: Yes, I kept it just to have a track until I was fully done with updating all this code.
The reason why it's not need is that now it's the caller of `reconcile_mongo_users_and_dbs` which is in charge of providing the correct relation to work on while in the past it was the `oversee_users` which was in charge of selecting the right relation and it lead to some nasty bugs because of that.

single_kernel_mongo/state/charm_state.py
Gu1nness: They are not the same thing and that's why we want to keep it separate:
 * `app_hosts` is for external access
 * `internal_host` is for internal access
 There are some cases where it's the same, some where it's not and we want to keep both.
 What can be done however would be to make a small method for the socket path section.

single_kernel_mongo/events/cluster.py
```diff
@@ -0,0 +1,156 @@
...
+        try:
+            self.manager.update_keyfile_and_hosts_on_mongos(event.relation)
+        except DeferrableFailedHookChecksError as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
+
+    def _on_relation_broken_event(self, event: RelationBrokenEvent):
+        try:
+            self.manager.on_relation_broken(event.relation)
```
Mehdi-Bendriss: `relation_broken` doesn't seem like a method name a manager should have, could you be more explicit / business logic oriented here?

single_kernel_mongo/events/cluster.py
```diff
@@ -0,0 +1,156 @@
...
+from single_kernel_mongo.utils.event_helpers import defer_event_with_info_log
+
+if TYPE_CHECKING:
+    from single_kernel_mongo.managers.mongodb_operator import MongoDBOperator
+    from single_kernel_mongo.managers.mongos_operator import MongosOperator
+
+logger = logging.getLogger(__name__)
+
+
+class ClusterConfigServerEventHandler(Object):
```
Mehdi-Bendriss, line 36: nit, missing return types  and docstrings on this and subsequent methods

single_kernel_mongo/events/cluster.py
```diff
@@ -0,0 +1,156 @@
...
+    def _on_relation_broken_event(self, event: RelationBrokenEvent):
+        try:
+            self.manager.on_relation_broken(event.relation)
+        except DeferrableFailedHookChecksError as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
+
+
+class ClusterMongosEventHandler(Object):
```
Mehdi-Bendriss, line 96: nit, missing return types and docstrings on this and subsequent methods

single_kernel_mongo/events/cluster.py
```diff
@@ -0,0 +1,156 @@
...
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_departed,
+            self.dependent.check_relation_broken_or_scale_down,
+        )
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_broken, self._on_relation_broken
+        )
+
+    def _on_relation_created(self, event: RelationCreatedEvent):
+        self.manager.relation_created()
```
Mehdi-Bendriss: similar comment on the naming from the manager standpoint and rest of occurrences

single_kernel_mongo/managers/backups.py
```diff
@@ -541,7 +541,7 @@ def _format_backup_list(self, backup_list: list[tuple[str, str, str]]) -> str:
 
         return "\n".join(backups)
 
-    @cached_property
+    @property
```
Mehdi-Bendriss, line 544: Could you explain this change? how (in)expensive was and is this call now?

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,371 @@
...
+from single_kernel_mongo.workload.mongos_workload import MongosWorkload
+
+if TYPE_CHECKING:
+    from single_kernel_mongo.managers.mongodb_operator import MongoDBOperator
+    from single_kernel_mongo.managers.mongos_operator import MongosOperator
+
+logger = getLogger(__name__)
+
+
+class ClusterProvider(Object):
```
Mehdi-Bendriss, line 42: nit, missing return types in a couple of methods

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,371 @@
...
+        config_server_db = self.state.generate_config_server_db()
+        self.dependent.mongo_manager.reconcile_mongo_users_and_dbs(relation)
+        relation_data = {
+            ClusterStateKeys.keyfile.value: self.state.get_keyfile(),
+            ClusterStateKeys.config_server_db.value: config_server_db,
+        }
+
+        int_tls_ca = self.state.tls.get_secret(label_name=SECRET_CA_LABEL, internal=True)
+
+        if int_tls_ca:
```
Mehdi-Bendriss: nit: walrus

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,371 @@
...
+        If it has departed, we run some checks and if we are a VM charm, we
+        proceed to reconcile the users and DB and cleanup mongoDB.
+        """
+        if self.state.upgrade_in_progress:
+            logger.warning(
+                "Removing integration to mongos is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
+            )
+
+        if not self.state.has_departed_run(relation.id):
+            raise DeferrableError(
```
Mehdi-Bendriss, line 130: can you explain the difference between: 
- DeferrableError
- DeferrableFailedHookChecksError

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,371 @@
...
+                raise DeferrableError
+
+            self.charm.status_manager.to_active("")
+
+        if self.charm.unit.is_leader():
+            self.state.app_peer_data.db_initialised = True
+
+        self.dependent.share_connection_info()
+
+    def relation_broken(self, relation: Relation):
```
Mehdi-Bendriss: related to a previous comment, when reading a manager code separately from the charm / lifecycles, it's hard to understand from a business logic standpoint, I believe names such as "cleanup_....."

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,371 @@
...
+                self.data_interface.delete_relation_data(
+                    relation.id, [ClusterStateKeys.int_ca_secret]
+                )
+            else:
+                self.data_interface.update_relation_data(
+                    relation.id, {ClusterStateKeys.int_ca_secret.value: new_ca}
+                )
+
+
+class ClusterRequirer(Object):
```
Mehdi-Bendriss, line 169: nit, missing return types in methods of this class

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,371 @@
...
+            user_secrets.get_content(refresh=True)
+            relation.data[self.charm.app].clear()
+
+    def is_ca_compatible(self) -> bool:
+        """Returns true if both the mongos and the config-server use the same CA.
+
+        Using the same CA is a requirement for sharded clusters.
+        """
+        config_server_relation = self.state.mongos_cluster_relation
+        if not config_server_relation:
```
Mehdi-Bendriss: nit: walrus

single_kernel_mongo/managers/mongo.py
```diff
@@ -318,24 +318,29 @@ def update_app_relation_data_for_config(self, relation: Relation, config: MongoC
...
-                )
-            if config.database != database:
-                data_interface.set_database(
-                    relation.id,
-                    config.database,
-                )
+        username = data_interface.fetch_my_relation_field(relation.id, "username")
+        password = data_interface.fetch_my_relation_field(relation.id, "password")
+        if not username or not password:
+            data_interface.set_credentials(
```
Mehdi-Bendriss, line 335: will the linter complain if we put arguments in the same line?

single_kernel_mongo/workload/mongos_workload.py
```diff
@@ -47,3 +49,16 @@ def layer(self) -> Layer:
             },
         }
         return Layer(layer_config)  # type: ignore
+
+    @property
+    def config_server_db(self) -> str | None:
+        """The config server DB on the workload."""
+        regex = re.compile(r"--configdb (\S+)")
+        env = self.get_env().get(self.env_var, None)
+        if not env:
```
Mehdi-Bendriss: nit, walrus

single_kernel_mongo/workload/mongos_workload.py
```diff
@@ -47,3 +49,16 @@ def layer(self) -> Layer:
...
+            return None
+
+        match = regex.search(env)
+        if match:
```
Mehdi-Bendriss: nit, walrus

Mehdi-Bendriss (commented):
Thanks Neha! left a few questions

single_kernel_mongo/events/cluster.py
Gu1nness: Sure, let me rename this!

single_kernel_mongo/managers/backups.py
Gu1nness: There are multiple occurences where we need to wait for the status to be updated (for example when syncing the configuration options).
The cached property sadly prevents from seeing the status updates so we need to get it every time.

The call is not extremely costly since pbm replies quickly on its status except when it cannot get some remote information but then this will be raised early

single_kernel_mongo/managers/cluster.py
Gu1nness: `DeferrableError` can be raised outside of hook checks
`DeferrableFailedHookChecksError` is only raised in hook checks.

The difference should be only in the event manager where the log line should differ according to which error is raised.

single_kernel_mongo/managers/cluster.py
Gu1nness: I have renamed all those methods to more meaningful ones

single_kernel_mongo/managers/cluster.py
Gu1nness: Even better since the variable is not use elswhere, just fully remove it

single_kernel_mongo/managers/mongo.py
Gu1nness: You mean on the `set_credentials` call ? No I don't think so but I find it easier to read this way.
No strong opinion though.

single_kernel_mongo/managers/cluster.py
```diff
@@ -115,7 +113,7 @@ def update_keyfile_and_hosts_on_mongos(self, relation: Relation) -> None:
 
         self.share_secret_to_mongos(relation)
 
-    def on_relation_broken(self, relation: Relation) -> None:
+    def cleanup_users(self, relation: Relation) -> None:
```
MiaAltieri: love this change! can you make sure this same approach is taken in the other manager PRs?

Gu1nness: Yes I'll do it !

single_kernel_mongo/managers/mongodb_operator.py
MiaAltieri: thoughts here?

single_kernel_mongo/managers/cluster.py
MiaAltieri: My thought was that most of the event handlers try except for DeferrableFailedHookChecksError and Non DeferrableFailedHookChecksError, so it might be best to be consistent in that approach.

single_kernel_mongo/events/cluster.py
```diff
@@ -0,0 +1,171 @@
...
+        """Database Created event handler.
+
+        When the database created event is received, we can proceed to store
+        credentials and share it to the client applications.
+        """
+        try:
+            self.manager.share_credentials_to_clients(event.username, event.password)
+        except (
+            DeferrableFailedHookChecksError,
+            WaitingForSecretsError,
```
MiaAltieri, line 142: I think if we start taking this approach we won't know when to raise a DeferrableFailedHookChecksError and a custom error that should also defer

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: We talked about it yesterday in the 1-1, I'll do this in a PR when all is merged to avoid some even more complex merges and rebases because of files moving around if you don't mind.
I created a ticket for that : https://warthogs.atlassian.net/browse/DPE-6330

single_kernel_mongo/managers/cluster.py
Gu1nness: No: They try/except on all expected errors (can be the ones raised by hook checks, or some that we want to bubble up to the event handler level because they should be handler in a specific way).
Failed hook checks will raise `DeferrableFailedHookChecksError` or `NonDeferrableFailedHookChecksError`.
The actual processing of the event can raise other issues.
My idea is to keep in the most accurate errors outside of that so that we keep some context and you can see in the other events handlers/managers that we handle way more errors.
What could be done in a future would be to have a bit of inheritance however to have an easier error handling at the end.

Mehdi-Bendriss (approved):

Mehdi-Bendriss (approved):

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,367 @@
...
+            )
+        if self.state.upgrade_in_progress:
+            raise DeferrableFailedHookChecksError(
+                "Processing client applications is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
+            )
+
+    def set_relation_created_status(self) -> None:
+        """Just sets a status on relation created."""
+        logger.info("Integrating to config-server")
+        self.charm.status_manager.to_waiting("Connecting to config-server")
```
MiaAltieri: this is new? Or maybe I am mistaken. If we keep it, can you modify the status to inclue `...` at the end?

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,367 @@
...
+
+        # We are a Kubernetes Mongos Charm so we are in charge of our client
+        # applications and their users and we proceed to update the users and their DBs.
+        try:
+            for relation in self.state.client_relations:
+                self.dependent.mongo_manager.reconcile_mongo_users_and_dbs(relation)
+        except PyMongoError:
+            raise DeferrableError("Failed to add users on mongos-k8s router.")
+
+    def remove_users(self, relation: Relation) -> None:
```
MiaAltieri:
```suggestion
    def remove_users_for_k8s_routers(self, relation: Relation) -> None:
```

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,367 @@
...
+        """Proceeds on relation broken."""
+        self.dependent.assert_proceed_on_broken_event(relation)
+        try:
+            self.remove_users(relation)
+        except PyMongoError:
+            raise DeferrableError("Trouble removing router users")
+
+        if not self.charm.unit.is_leader():
+            return
+
```
MiaAltieri, line 279: from what I can tell this could be moved closer to the top?

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,367 @@
...
+            self.charm.status_manager.to_active("")
+
+        if self.charm.unit.is_leader():
+            self.state.app_peer_data.db_initialised = True
+
+        self.dependent.share_connection_info()
+
+    def remove_users_and_cleanup_mongo(self, relation: Relation) -> None:
+        """Proceeds on relation broken."""
```
MiaAltieri, line 268: newb question, don't we need to [stop the mongos daemon](https://github.com/canonical/mongos-operator/blob/9f81c03c9f23faf1975125bdf77c96ea08f916df/lib/charms/mongodb/v0/config_server_interface.py#L367C1-L368C1)

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,367 @@
...
+            raise DeferrableError("Trouble removing router users")
+
+        if not self.charm.unit.is_leader():
+            return
+
+        logger.info("Cleaning database and user removed for mongos application")
+        self.state.secrets.remove(Scope.APP, AppPeerDataKeys.username.value)
+        self.state.secrets.remove(Scope.APP, AppPeerDataKeys.password.value)
+
+    def update_users(self) -> None:
```
MiaAltieri:
```suggestion
    def update_users_for_k8s_routers(self) -> None:
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -365,12 +400,14 @@ def is_mongos_running(self) -> bool:
...
-                uri = quote(f"{self.workload.paths.socket_path}", safe="")
+                host = self.state.formatted_socket_path
         else:
-            uri = self.state.unit_peer_data.host + f":{MongoPorts.MONGOS_PORT}"
+            host = self.state.unit_peer_data.host + f":{MongoPorts.MONGOS_PORT}"
+
+        uri = f"mongodb://{host}"
 
-        return self.mongo_manager.mongod_ready(uri=uri)
+        return self.mongo_manager.mongod_ready(uri=uri, direct=False)
```
MiaAltieri: direct=True

single_kernel_mongo/events/cluster.py
Gu1nness: The errors mean different things, they bear a meaning.
I strongly think that keeping meaningful errors is important and that we're able as developers to differentiate which errors should defer and which should not defer.
There's nothing really complicated here:
1) Run hook checks, and if it fails raise `DeferrableFailedHookChecksError` or `NonDeferrableFailedHookChecksError`
2) Run the business logic of the hook and if something is raised, react to it.

If what bugs you is that they are handled in the same except case, I can split and differentiate the logs, but I really think keeping multiple different errors reinforces the stability of the charm and trust developers can have in their code.

single_kernel_mongo/managers/cluster.py
Gu1nness: It's not exactly new but it comes from the last stable release : https://github.com/canonical/mongodb-operator/blob/6/edge/lib/charms/mongodb/v0/config_server_interface.py#L278-L281
I'll add the …

single_kernel_mongo/managers/cluster.py
Gu1nness: I'm not sure, it should not be a problem but I'll give it a try first I think.
All replicas of the mongos should have the same user but I want to check that it doesn't cause any issue.

single_kernel_mongo/managers/cluster.py
Gu1nness: That's a good question and I think that we should indeed, I'll add that.
This wasn't done in the former charm, however.

MiaAltieri: I think it was done in the former charm (check the link), but it looks like it was done in a separate function 

Thank you so much for getting this added!

single_kernel_mongo/events/cluster.py
MiaAltieri: no worries! I was worried about being a developer and not knowing which exceptions to raise and when. But I understand the logic now (thanks to your extremely helpful notes here!) I really appreciate your patience Neha and your careful explanations. 

Maybe at some point in the future we can add a guide on how to contribute to this repo, while maintaining the structure :)

single_kernel_mongo/managers/cluster.py
MiaAltieri: now that the stop has been added it should stay right where it is :)

MiaAltieri (approved):

PR #7, opened 2024-12-16
feat(statuses): basic integration

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

Add a (still basic but) better status handler.
This is almost a copy paste of the current status handler but spread out between state and status prioritization and setting

## 📑 Motivation and Context

One of the final steps to reach the feature parity with the former charms.

Manually tested for now :)

## ✅ Checklist


single_kernel_mongo/abstract_charm.py
```diff
@@ -97,7 +99,6 @@ def on_install(self, _):
 
     def on_leader_elected(self, _):
         """First leader elected handler."""
-        # FIXME: Check role status in databag first.
-
         # Sets the role in the databag.
-        self.operator.state.app_peer_data.role = self.parsed_config.role
+        if self.operator.state.app_peer_data.role == MongoDBRoles.UNKNOWN:
```
MiaAltieri:
```suggestion
        # For now we do not support charms migrating roles. We instead save the first role that the charm was deployed with.
        # this prevents the user from changing the config. 
        if self.operator.state.app_peer_data.role == MongoDBRoles.UNKNOWN:
```

single_kernel_mongo/core/status_provider.py
```diff
@@ -0,0 +1,17 @@
...
+
+from ops.model import StatusBase
+
+
+class StatusProvider(Protocol):
+    """Enforces all classes inheriting from this class to provide a `get_status` method."""
+
+    @abstractmethod
+    def get_status(self) -> StatusBase | None:
+        """Returns a sensitive status about the manager."""
```
MiaAltieri, line 16: i like

single_kernel_mongo/core/structured_config.py
```diff
@@ -40,13 +40,24 @@ class MongoDBRoles(str, Enum):
...
     """The possible values for the expose-external config value."""
 
+    UNKNOWN = ""
     NODEPORT = "nodeport"
     NONE = "none"
 
+    @classmethod
+    def valid_roles(cls) -> set[str]:
+        """The valid roles for expose external."""
+        return {cls.NODEPORT, cls.NONE}
```
MiaAltieri: I know this comment should prob go on the `mongos` PR , but I think it might be nice to include something for mongos vm

single_kernel_mongo/events/cluster.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env python3
```
MiaAltieri: this is a file from another PR right?

single_kernel_mongo/events/lifecycle.py
```diff
@@ -164,9 +164,12 @@ def on_relation_changed(self, event: RelationChangedEvent):
             event.defer()
             return
 
+        self.charm.status_manager.process_and_share_statuses()
+
     def on_relation_departed(self, event: RelationDepartedEvent):
         """Relation departed event."""
         self.dependent.on_relation_departed(departing_unit=event.departing_unit)
+        self.charm.status_manager.process_and_share_statuses()
```
MiaAltieri: newb question why is this in relation_departed?

single_kernel_mongo/events/sharding.py
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env python3
```
MiaAltieri: this was also introduced in a separate PR yeah?

single_kernel_mongo/managers/cluster.py
```diff
@@ -0,0 +1,349 @@
+# Copyright 2024 Canonical Ltd.
```
MiaAltieri: same here - this was introduced in a separe PR right?

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -0,0 +1,353 @@
+#!/usr/bin/python3
```
MiaAltieri: this was added as part of the mongos pr right? what here should be reviewed that is different here as a part of statuses?

single_kernel_mongo/managers/sharding.py
```diff
@@ -0,0 +1,787 @@
+# Copyright 2024 Canonical Ltd.
```
MiaAltieri: same here , what should be reviewed that is relevant to statuses?

single_kernel_mongo/core/operator.py
```diff
@@ -148,6 +152,18 @@ def is_relation_feasible(self, name: str) -> bool:
         """Checks if the relation is feasible in this context."""
         ...
 
+    def assert_proceed_on_broken_event(self, relation: Relation):
```
Mehdi-Bendriss:
```suggestion
    def assert_proceed_on_broken_event(self, relation: Relation) -> None:
```

single_kernel_mongo/events/cluster.py
```diff
@@ -0,0 +1,158 @@
...
+from single_kernel_mongo.utils.event_helpers import defer_event_with_info_log
+
+if TYPE_CHECKING:
+    from single_kernel_mongo.managers.mongodb_operator import MongoDBOperator
+    from single_kernel_mongo.managers.mongos_operator import MongosOperator
+
+logger = logging.getLogger(__name__)
+
+
+class ClusterConfigServerEventHandler(Object):
```
Mehdi-Bendriss: reviewed in https://github.com/canonical/mongo-single-kernel-library/pull/6

single_kernel_mongo/events/cluster.py
```diff
@@ -0,0 +1,158 @@
...
+        try:
+            self.manager.on_relation_broken(event.relation)
+            self.charm.status_manager.process_and_share_statuses()
+        except DeferrableFailedHookChecksError as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
+
+
+class ClusterMongosEventHandler(Object):
```
Mehdi-Bendriss: reviewed in https://github.com/canonical/mongo-single-kernel-library/pull/6

single_kernel_mongo/events/database.py
```diff
@@ -66,14 +66,13 @@ def _on_relation_event(self, event: RelationEvent):
...
         relation_changed = False
-        # TODO : Handle the Mongos VM case
-        # (https://github.com/canonical/mongos-operator/blob/6/edge/lib/charms/mongos/v0/mongos_client_interface.py)
-        # for interface unification.
         if (
             self.dependent.substrate == Substrates.VM
             and self.relation_name == RelationNames.MONGOS_PROXY
         ):
             self.dependent.share_credentials(event.relation)  # type: ignore[attr-defined]
+            return
```
Mehdi-Bendriss: Can you explain this return?

Gu1nness: The client relation is not handler the same way if we are a mongos vm than all other cases (because we are a subordinate charm).

In the very specific case of a mongos vm, we just share credentials to the client application.

single_kernel_mongo/managers/config.py
```diff
@@ -283,7 +283,10 @@ def db_path_argument(self) -> list[str]:
     @property
     def role_parameter(self) -> list[str]:
         """The role parameter."""
-        match self.state.app_peer_data.role:
+        role = self.state.app_peer_data.role
+        if role == MongoDBRoles.UNKNOWN:  # First install we don't have the role in databag yet.
+            role = self.state.config.role
```
MiaAltieri:
```suggestion
        # First install we don't have the role in databag yet.
        role = self.state.config.role if self.state.app_peer_data.role == MongoDBRoles.UNKNOWN == else self.state.app_peer_data.role
```

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -508,18 +519,10 @@ def on_update_status(self) -> None:
...
             self.perform_self_healing()
-        except ServerSelectionTimeoutError:
-            deployment = (
-                "replica set" if self.state.is_role(MongoDBRoles.REPLICATION) else "cluster"
-            )
-            self.charm.status_manager.to_waiting(
-                f"Waiting to sync internal membership across the {deployment}"
-            )
-        else:
-            self.charm.status_manager.to_active(None)
```
MiaAltieri, line 521: @Gu1nness is this done elsewhere?

single_kernel_mongo/managers/sharding.py
```diff
@@ -224,7 +240,7 @@ def get_status(self) -> StatusBase | None:
             )
 
         uri = f"mongodb://{','.join(self.state.app_hosts)}"
-        if not self.dependent.mongo_manager.mongod_ready(uri):
+        if not self.dependent.mongo_manager.mongod_ready(uri, direct=False):
```
MiaAltieri: why does this need to be False? I could be wrong, but I think it should be True when we check for readiness

single_kernel_mongo/managers/sharding.py
```diff
@@ -671,3 +675,124 @@ def drained(self, mongos_hosts: list[str], shard_name: str):
...
+        except ServerSelectionTimeoutError:
+            # Connection refused, - this occurs when internal membership is not in sync across the
+            # cluster (i.e. TLS + KeyFile).
+            return False
+
+        return True
+
+    def _is_added_to_cluster(self) -> bool:
+        """Returns true if the shard has been added to the clusted."""
+        if not self.state.config_server_name or not self.state.app_peer_data.mongos_hosts:
```
MiaAltieri, line 706:
```suggestion
        # this information is required in order to check if we have been added 
        if not self.state.config_server_name or not self.state.app_peer_data.mongos_hosts:
```

single_kernel_mongo/managers/sharding.py
```diff
@@ -671,3 +675,124 @@ def drained(self, mongos_hosts: list[str], shard_name: str):
...
+        return self.state.app_peer_data.replica_set in members
+
+    def _is_shard_aware(self) -> bool:
+        config = self.state.mongos_config_for_user(
+            OperatorUser, set(self.state.app_peer_data.mongos_hosts)
+        )
+        with MongoConnection(config) as mongo:
+            return mongo.is_shard_aware(self.state.app_peer_data.replica_set)
+
+    def skip_shard_status(self) -> bool:
```
MiaAltieri:
```suggestion
    def should_skip_shard_status(self) -> bool:
```
nit: bool functions should start with a question word

single_kernel_mongo/state/app_peer_state.py
```diff
@@ -77,8 +77,8 @@ def role(self) -> MongoDBRoles:
 
         Either from the app databag or from the default from config.
         """
-        databag_role: str | None = self.relation_data.get(AppPeerDataKeys.role.value)
-        if not databag_role:
+        databag_role: str | None = self.relation_data.get(AppPeerDataKeys.ROLE.value)
```
MiaAltieri: will this fail if there is no relation?

single_kernel_mongo/state/charm_state.py
```diff
@@ -457,6 +458,21 @@ def set_scaling_down(self, rel_id: int, departing_unit_name: str) -> bool:
         self.unit_peer_data.update({rel_departed_key: json.dumps(scaling_down)})
         return scaling_down
 
+    def share_status_with_config_server(self, status: StatusBase):
+        """Shares this shard's status with the config server."""
```
MiaAltieri:
```suggestion
        """
        Shares this shard's status with the config server.
        
        This is primarily useful for when the cluster upgrades, since the config-server will need to ensure all units are healthy.
        """
```

single_kernel_mongo/state/charm_state.py
```diff
@@ -457,6 +458,21 @@ def set_scaling_down(self, rel_id: int, departing_unit_name: str) -> bool:
...
+        if not self.is_role(MongoDBRoles.SHARD):
+            return
+        if not self.shard_relation:
```
MiaAltieri, line 470: should this be `config_relation`?

single_kernel_mongo/state/charm_state.py
```diff
@@ -457,6 +458,21 @@ def set_scaling_down(self, rel_id: int, departing_unit_name: str) -> bool:
...
+            return
+        if isinstance(status, ActiveStatus):
+            self.shard_state.status_ready_for_upgrade = True
+        if not isinstance(status, BlockedStatus):
+            self.shard_state.status_ready_for_upgrade = False
+        if status.message and "is not up-to date with config-server" in status.message:
+            self.shard_state.status_ready_for_upgrade = True
+
+        self.shard_state.status_ready_for_upgrade = False
```
MiaAltieri, line 484: do we really want to overwrite the values we just set?

single_kernel_mongo/state/charm_state.py
```diff
@@ -457,6 +458,21 @@ def set_scaling_down(self, rel_id: int, departing_unit_name: str) -> bool:
...
+    def share_status_with_config_server(self, status: StatusBase):
+        """Shares this shard's status with the config server."""
+        if not self.is_role(MongoDBRoles.SHARD):
+            return
+        if not self.shard_relation:
+            return
+        if isinstance(status, ActiveStatus):
+            self.shard_state.status_ready_for_upgrade = True
+        if not isinstance(status, BlockedStatus):
+            self.shard_state.status_ready_for_upgrade = False
```
MiaAltieri, line 478: I assume this property sets some relation data? maybe it can be renamed in a more obvious way?

single_kernel_mongo/state/charm_state.py
```diff
@@ -537,8 +553,10 @@ def operator_config(self) -> MongoConfiguration:
     @property
     def mongos_config(self) -> MongoConfiguration:
         """Mongos Configuration for the mongos user."""
-        username = self.secrets.get_for_key(Scope.APP, key=AppPeerDataKeys.username.value)
-        password = self.secrets.get_for_key(Scope.APP, key=AppPeerDataKeys.password.value)
+        if self.charm_role.name == KindEnum.MONGOD:
+            return self.mongos_config_for_user(OperatorUser, self.app_hosts)
```
MiaAltieri: if the kind above is mongod, do we want to set mongos?

single_kernel_mongo/events/cluster.py
Gu1nness: Yes it is, just some git hell

single_kernel_mongo/events/lifecycle.py
Gu1nness: My idea was that this we on most of the usual events we have a consistent status handling and not some dangling statuses but it's not extremely important and could be removed IMHO

single_kernel_mongo/events/sharding.py
Gu1nness: Yes, same some git hell

single_kernel_mongo/managers/cluster.py
Gu1nness: Yeah

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Yes and there is nothing as part of statuses :)

single_kernel_mongo/managers/sharding.py
Gu1nness: I created the `get_status` methods that collect the statuses, based on what was existing in the former charms.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: This is now in mongo.py (as a part of `get_status`)

single_kernel_mongo/managers/sharding.py
Gu1nness: Not always.
`direct` can be true when we ping on localhost, which is not the case here.
This is the exact same code as in the former charm.

single_kernel_mongo/state/app_peer_state.py
Gu1nness: It won't :)
If you have a look at `single_kernel_mongo/state/abstract_state.py` you'll see that:
```
self.relation_data = self.data_interface.as_dict(self.relation.id) if self.relation else {}
```
Which means we are safe

single_kernel_mongo/state/charm_state.py
Gu1nness: This is probably poorly named but `shard_relation` is the relation on the shard side (`sharding` is the relation name)

single_kernel_mongo/state/charm_state.py
Gu1nness: Let's just say I forgot to add the returns :sweat_smile: 

Fixing it

single_kernel_mongo/state/charm_state.py
Gu1nness: This pattern is used all over the shard state in order to have a proxy between python objects and databags.
They are all properties defined in the correct State object.
I'm not sure it's a good thing to rename it because it does exactly what it means: write `false` in the `status-shows-ready-for-upgrade` key in the databag.

IF this one is considered wrong, then all the other databag interactions are as well and I would have loved to know it earlier.

single_kernel_mongo/state/charm_state.py
Gu1nness: Because we're getting the mongoS configuration for the mongoDB charm which is useful when we run in a sharded deployment context

single_kernel_mongo/state/charm_state.py
Gu1nness: Btw this is following DA067 the spec about the new charm structures

single_kernel_mongo/events/lifecycle.py
MiaAltieri: I think it should be removed, right now `process_and_share_statuses()` doesn't contain all the feasible statuses for mongodb so it has the possibility to overwrite a status that was set earlier (this will be fixed when we implement statuses)

single_kernel_mongo/state/charm_state.py
MiaAltieri: Thanks for the explanation! The explanation helps it make more sense. :) 

Perhaps this should be renamed so that others won't have the same question?

single_kernel_mongo/state/charm_state.py
MiaAltieri: I see what you're saying. It makes sense that `self.shard_state` is the data-bag , but it isn't immediately obvious that this is written in the databag that is related to the config server. Maybe a comment in here would be useful?

single_kernel_mongo/state/charm_state.py
MiaAltieri: Ah so `if self.charm_role.name == KindEnum.MONGOD:` is a check of "is this charmed MongoDB or Charmed Mongos"? If so maybe the kind should be renamed to `charmed_mongodb` otherwise it seems like the daemon

single_kernel_mongo/state/app_peer_state.py
MiaAltieri: Thank you!

MiaAltieri (commented):
looks good! Just responded to some of the original comments from the last review :)

single_kernel_mongo/events/lifecycle.py
Gu1nness: Removing it then :)

single_kernel_mongo/state/charm_state.py
Gu1nness: If you don't mind, I'll create a ticket to do that after all is merged otherwise this will be hell to rename and will create so many conflicts: https://warthogs.atlassian.net/browse/DPE-6321

single_kernel_mongo/state/charm_state.py
Gu1nness: Yeah sure

single_kernel_mongo/state/charm_state.py
Gu1nness: Mehdi made a similar comment about the name `KindEnum` in another PR.
I have to come up with a better naming.
The `charmed_mongodb` brings a good idea and I'll could go with
```python3
class CharmKind(str, Enum):
    MONGODB = "mongodb"
    MONGOS = "mongos"
```
Does it make more sense?

MiaAltieri: omg I would love that approach! Thanks for being flexible! 🤩

single_kernel_mongo/events/cluster.py
MiaAltieri: 😭

single_kernel_mongo/core/structured_config.py
Gu1nness: WDYM? A check that we're in mongos k8s in this case? Or 
At this level we don't know who we are and have no easy way to know and it's handled below.
And in any case, Mongos VM won't expose this configuration option (or if it should I'll have some changes to bring).

single_kernel_mongo/state/charm_state.py
Gu1nness: See b31de6b

single_kernel_mongo/managers/sharding.py
MiaAltieri: Hmm I think it is a best practice that when checking readiness we have a direct connection, see docs:

> directConnection (optional): if True, forces this client to
connect directly to the specified MongoDB host as a standalone. If false, the client connects to the entire replica set of which the given MongoDB host(s) is a part. If this is True and a mongodb+srv:// URI or a URI containing multiple seeds is provided, an exception will be raised.

If we connect to a single replica and specify direct=False it can connect to the entire replica set.

single_kernel_mongo/core/structured_config.py
MiaAltieri: I totally misunderstood this function my apologies. I was referring to the fact that Mongos VM can be exposed externally (the difference between it running on the unix socket vs an ip address).

Would you mind updating the docstring though to say `"""The valid roles for expose external for k8s routers."""`

single_kernel_mongo/state/charm_state.py
MiaAltieri: maybe the comment was missed, something like:
```
# Setting the relation data on the shard state for `read_for_upgrade` is propogated to the config-server relation data
```

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -277,7 +289,7 @@ def on_start(self) -> None:
             return
 
         self._initialise_replica_set()
-        self.charm.status_manager.to_active(None)
+        self.charm.status_manager.process_and_share_statuses()
```
MiaAltieri: Just a heads up you still have several calls to `process_and_share_statuses`, which can be troublesome, see earlier comment of:

> `process_and_share_statuses()` doesn't contain all the feasible statuses for mongodb so it has the possibility to overwrite a status that was set earlier (this will be fixed when we implement statuses) 

So having this call at the end of each hook can be problematic

single_kernel_mongo/managers/sharding.py
Gu1nness: Ok but [this is not my code initially](https://github.com/canonical/mongodb-operator/blob/6/edge/lib/charms/mongodb/v1/shards_interface.py#L489), it's [yours](https://github.com/canonical/mongodb-operator/commit/437b04a81ade9cff80d73d5055ea15827fc5d3c5#diff-da2d55a82fb6a610d1da4b5ee4790dcf67f08483a55bcf15b4b79162f5322232) (and initial commit [here](https://github.com/canonical/mongodb-operator/commit/0b904b6014e3e0da489cb996c7ee8f0822d6a988)).
I think the reason is the following: the charm runs in a pod which is different from the pod which runs the mongos service.

So:
 * Either `direct = False` and we connect to all hosts
 * Or `direct = True` and we connect to our host only.

I can give a try to the second one but I think it's important to know why this decision was made in the first place.

single_kernel_mongo/managers/config.py
```diff
@@ -283,7 +283,13 @@ def db_path_argument(self) -> list[str]:
     @property
     def role_parameter(self) -> list[str]:
         """The role parameter."""
-        match self.state.app_peer_data.role:
+        # First install we don't have the role in databag yet.
+        role = (
+            self.state.config.role
+            if self.state.app_peer_data.role == MongoDBRoles.UNKNOWN
```
Mehdi-Bendriss, line 289: what about on a subsequent config change to an erroneous value? is `self.state.config.role` validated?

single_kernel_mongo/core/workload.py
```diff
@@ -270,6 +270,18 @@ def get_version(self) -> str:
...
         return version
 
+    def get_internal_revision(self) -> str:
+        """Get the internal revision.
+
+        Returns:
+            String of charm internal revision
+        """
+        try:
+            version = Path("charm_internal_version").read_text().strip()
```
Mehdi-Bendriss, line 280: are we still relying on this file? I thought we moved away from it?

single_kernel_mongo/managers/sharding.py
Mehdi-Bendriss: This might have been overlooked in the first implementation, i think we should attempt to use a direct connection here

single_kernel_mongo/managers/sharding.py
```diff
@@ -359,13 +351,13 @@ def get_unreachable_shards(self) -> list[str]:
             # use a URI that is not dependent on the operator password, as we are not guaranteed
             # that the shard has received the password yet.
             uri = f"mongodb://{','.join(hosts)}"
-            if not self.dependent.mongo_manager.mongod_ready(uri):
+            if not self.dependent.mongo_manager.mongod_ready(uri, direct=False):
```
Mehdi-Bendriss, line 355: similar comment on the `direct` arg

single_kernel_mongo/managers/sharding.py
```diff
@@ -678,3 +670,125 @@ def drained(self, mongos_hosts: list[str], shard_name: str) -> bool:
...
+                OperatorUser, set(self.state.app_peer_data.mongos_hosts)
+            )
+
+            with MongoConnection(config) as mongos:
+                mongos.get_shard_members()
+            # check our ability to use connect to mongod
+            with MongoConnection(self.state.mongo_config) as mongod:
+                mongod.get_replset_status()
+        except OperationFailure as e:
+            if e.code in [13, 18, 133]:
```
Mehdi-Bendriss, line 693: Can you add a comment explaining these codes?

single_kernel_mongo/managers/sharding.py
```diff
@@ -678,3 +670,125 @@ def drained(self, mongos_hosts: list[str], shard_name: str) -> bool:
...
+
+        try:
+            config = self.state.mongos_config_for_user(
+                OperatorUser, set(self.state.app_peer_data.mongos_hosts)
+            )
+            # check our ability to use connect to mongos
+            with MongoConnection(config) as mongos:
+                members = mongos.get_shard_members()
+        except OperationFailure as e:
+            if e.code in [13, 18, 133]:
```
Mehdi-Bendriss, line 717: similar comment on comment explaining the codes

single_kernel_mongo/managers/sharding.py
```diff
@@ -678,3 +670,125 @@ def drained(self, mongos_hosts: list[str], shard_name: str) -> bool:
...
+                return False
+            raise
+        except ServerSelectionTimeoutError:
+            # Connection refused, - this occurs when internal membership is not in sync across the
+            # cluster (i.e. TLS + KeyFile).
+            return False
+
+        return self.state.app_peer_data.replica_set in members
+
+    def _is_shard_aware(self) -> bool:
```
Mehdi-Bendriss, line 727: missing docstring

single_kernel_mongo/state/charm_state.py
```diff
@@ -457,6 +458,30 @@ def set_scaling_down(self, rel_id: int, departing_unit_name: str) -> bool:
         self.unit_peer_data.update({rel_departed_key: json.dumps(scaling_down)})
         return scaling_down
 
+    def share_status_with_config_server(self, status: StatusBase):
```
Mehdi-Bendriss:
```suggestion
    def share_status_with_config_server(self, status: StatusBase) -> None:
```

single_kernel_mongo/state/config_server_state.py
```diff
@@ -42,36 +46,49 @@ def mongos_hosts(self) -> list[str]:
...
+    def status_ready_for_upgrade(self) -> bool:
+        """Returns true if the shard is ready for upgrade."""
+        if not self.relation:
+            return True
+        return json.loads(
+            self.relation_data.get(ConfigServerKeys.STATUS_READY_FOR_UPGRADE.value, "false")
+        )
+
+    @status_ready_for_upgrade.setter
+    def status_ready_for_upgrade(self, value: bool):
```
Mehdi-Bendriss, line 93:
```suggestion
    def status_ready_for_upgrade(self, value: bool) -> None:
        """Update the current status for upgrade readiness."""
```

single_kernel_mongo/status.py
```diff
@@ -15,21 +16,49 @@
...
     """Status Manager."""
 
     def __init__(self, charm: AbstractMongoCharm):
         super().__init__(parent=charm, key="status")
         self.charm = charm
+        self.operator = charm.operator
+        self.state = charm.operator.state
 
-    def set_and_share_status(self, status: StatusBase | None):
+    def set_and_share_status(self, status: StatusBase):
```
Mehdi-Bendriss:
```suggestion
    def set_and_share_status(self, status: StatusBase) -> None:
```

single_kernel_mongo/status.py
```diff
@@ -15,21 +16,49 @@
...
         """Sets the unit status."""
-        self.charm.unit.status = status or ActiveStatus()
+        self.charm.unit.status = status
+        if self.state.is_role(MongoDBRoles.SHARD):
+            self.state.share_status_with_config_server(status)
+
+    def to_active(self, message: str | None = None):
```
Mehdi-Bendriss:
```suggestion
    def to_active(self, message: str | None = None) -> None:
```

single_kernel_mongo/status.py
```diff
@@ -15,21 +16,49 @@
...
+from single_kernel_mongo.config.literals import CharmKind
+from single_kernel_mongo.core.structured_config import MongoDBRoles
 
 if TYPE_CHECKING:
     from single_kernel_mongo.abstract_charm import AbstractMongoCharm
 
+logger = getLogger(__name__)
+
+
+StatusesDict = TypedDict(
```
Mehdi-Bendriss: why not a class with optional args to None? will avoid us having to deal with string keys in `get_statuses`

single_kernel_mongo/status.py
```diff
@@ -39,17 +68,93 @@ def to_waiting(self, message: str):
...
+        waiting_status = None
+        try:
+            statuses = self.get_statuses()
+        except OperationFailure as e:
+            if e.code in [13, 18]:
+                waiting_status = f"Waiting to sync passwords across the {deployment_mode}"
+            elif e.code == 133:
+                waiting_status = f"Waiting to sync internal membership across the {deployment_mode}"
+            else:
+                raise
```
Mehdi-Bendriss, line 134: How is this one handled? I am not sure this method should raise an exception, should it?

single_kernel_mongo/status.py
```diff
@@ -39,17 +68,93 @@ def to_waiting(self, message: str):
...
+
+        if waiting_status:
+            self.set_and_share_status(WaitingStatus(waiting_status))
+            return
+
+        main_status = self.prioritize_statuses(statuses)
+
+        logger.info(f"{' Charm Statuses ':=^40}")
+        for key, value in statuses.items():
+            if value:
```
Mehdi-Bendriss, line 146: I believe we can drop this and log `None`

Mehdi-Bendriss (commented):
Thanks Neha! I left a few comments and questions

single_kernel_mongo/events/lifecycle.py
```diff
@@ -164,6 +164,8 @@ def on_relation_changed(self, event: RelationChangedEvent):
             event.defer()
             return
 
+        self.charm.status_manager.process_and_share_statuses()
```
MiaAltieri, line 167: you sure you want this here?

single_kernel_mongo/managers/sharding.py
MiaAltieri: Okay, I read some more of the code here. I think what this check is supposed to do is:
1. check if the mongos for this unit is running

But what it looks like it does is check if there is a response for *any* mongod in the deployed application (I think we need to make the uri specify the port for mongos + use only the host of this unit) 

Once we have that we should use direct=True

My apologies for writing this incorrect in teh first place, I appreciate your patience

single_kernel_mongo/managers/sharding.py
MiaAltieri: or maybe even a constant? IIRC in another PR you made a constant for these numbers?

single_kernel_mongo/state/charm_state.py
```diff
@@ -537,8 +562,10 @@ def operator_config(self) -> MongoConfiguration:
     @property
     def mongos_config(self) -> MongoConfiguration:
         """Mongos Configuration for the mongos user."""
```
MiaAltieri:
```suggestion
        """Mongos Configuration for the admin mongos user."""
```

single_kernel_mongo/status.py
```diff
@@ -39,17 +68,93 @@ def to_waiting(self, message: str):
...
+                    "shard": self.operator.shard_manager.get_status(),
+                    "config-server": self.operator.config_server_manager.get_status(),
+                    "PBM": self.operator.backup_manager.get_status(),
+                }
+            )
+        return StatusesDict(
+            {
+                "mongodb": WaitingStatus("waiting for mongos to start")
+                if not self.operator.workload.active
+                else ActiveStatus(),
```
MiaAltieri: I believe this doesn't include cases like:
                WaitingStatus("Waiting for secrets from config-server")
            return BlockedStatus("mongos requires TLS to be enabled.")
            return BlockedStatus("mongos has TLS enabled, but config-server does not.")
            return BlockedStatus("mongos CA and Config-Server CA don't match.")

In the original verision of the mongos charm there are various states of the mongos status [seen in update_status](https://github.com/canonical/mongos-operator/blob/9f81c03c9f23faf1975125bdf77c96ea08f916df/src/charm.py#L131-L152), it might be good to include those here?

single_kernel_mongo/status.py
```diff
@@ -39,17 +68,93 @@ def to_waiting(self, message: str):
...
+        """
+        # retrieve statuses of different services running on Charmed MongoDB
+        deployment_mode = (
+            "replica set" if self.state.is_role(MongoDBRoles.REPLICATION) else "cluster"
+        )
+        waiting_status = None
+        try:
+            statuses = self.get_statuses()
+        except OperationFailure as e:
+            if e.code in [13, 18]:
```
MiaAltieri, line 129: [can we keep the original constants](https://github.com/canonical/mongodb-operator/blob/6306245b18b2627d521f50a7595c36035bcd47fb/lib/charms/mongodb/v0/set_status.py#L182-L184)

single_kernel_mongo/status.py
MiaAltieri: it isn't handled (in the original too), the exception is raised since it means the charm is in an error state in which no events should proceed.

single_kernel_mongo/managers/config.py
Gu1nness: It is checked in the config_changed event handler which prevents from changing the role (because we don't support that).

single_kernel_mongo/core/workload.py
Gu1nness: This is removed in #8 
I ported it bluntly before realising we moved away from this internal version file so no worries it will disappear.

single_kernel_mongo/managers/sharding.py
Gu1nness: I haven't made a PR yet but I created a ticket to do that in the near future.
Probably next week while the CIs will be running :)

single_kernel_mongo/managers/sharding.py
Gu1nness: Similar answer, I'll make a PR next week if that's okay for you?

single_kernel_mongo/status.py
Gu1nness: Yeah sure I can do that

single_kernel_mongo/status.py
Gu1nness: The idea was to generate only the logs that are meaningful (for example config-server won't generate a line for `shard` etc.
WDYT?

single_kernel_mongo/events/lifecycle.py
Gu1nness: I guess we discussed about it in another PR, we'll see how it goes but for now it's quite helpful and reports statuses faster (for example, no need to wait until the first `update-status` to know we're missing an integration if we're a config-server or a shard.
If something goes south, it'll be removed.

single_kernel_mongo/status.py
Gu1nness: You're right.
I'll add this tomorrow morning.

single_kernel_mongo/status.py
Gu1nness: I already answered quite a few times that I'll do it later, I have a ticket created for that [here](https://warthogs.atlassian.net/browse/DPE-6396).
I'll do it when the CI are running, next week.

single_kernel_mongo/managers/sharding.py
Gu1nness: Right :)
Let me do that change

single_kernel_mongo/managers/sharding.py
Gu1nness: Hum, the situation is different here I think.
In this case, as the config-server we want to see if the shard is reachable. I think that in that case, it's better to ping the whole replica set of the shard because a shard might not be available but the shard replicaset still valid.
WDYT?

single_kernel_mongo/managers/sharding.py
Mehdi-Bendriss: sounds good, could you create a ticket so we don't forget?

single_kernel_mongo/events/lifecycle.py
MiaAltieri: Great! thats fine by me! 

Many thanks Neha!

single_kernel_mongo/managers/sharding.py
MiaAltieri: Thanks, also my apologies for writing this incorrectly in the first place!

single_kernel_mongo/managers/sharding.py
Gu1nness: [here it is](https://warthogs.atlassian.net/browse/DPE-6396)

single_kernel_mongo/managers/sharding.py
Gu1nness: [ticket here](https://warthogs.atlassian.net/browse/DPE-6396)

single_kernel_mongo/status.py
Gu1nness: It's now done with bc7a821

single_kernel_mongo/managers/sharding.py
MiaAltieri: The migration to single kernel charm is certainly revealing a lot of areas from the original charm that are broken or not well documented. Thank you @Gu1nness for your patience and for letting us use the migration as a chance to fix these issues.

TLDR: it is good as is leave these comments so it is more clear in the future:
1. here
```
# To check if the shard is ready we check the entire replica set for readiness
```
2. in the `mongod_ready` function:
```
"""
<pre-existing content in the docstring>

When using this function pass `direct=True`, when checking if a *single replica* is ready, when checking if an entire replica set is ready pass `direct=False`
"""
```
 
Long Answer:
In this scenario we are checking whether an *entire* shard is reachable (i.e. the sharded replica set), the reason that we don't check a single replica from the shard, is because we care if the shard itself is responsive, not a single replica. The beauty of an HA deployment is that a replica can be down in a shard, but the shard itself can still be responsive and functional. 

Note that when it is critical to check *if all replicas* are responsive we do (i.e. upgrades)

Gu1nness: Thanks Mia, I'll add some comments !

Mehdi-Bendriss (approved):

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -198,37 +198,22 @@ def on_update_status(self) -> None:
...
-                "Missing integration to config-server. mongos cannot run unless connected to config-server."
-            )
-            self.charm.status_manager.to_blocked("Missing relation to config-server.")
-            return
 
-        if status := self.cluster_manager.get_tls_statuses():
-            logger.info(f"Invalid TLS integration: {status.message}")
-            self.charm.status_manager.set_and_share_status(status)
-            return
+        if self.get_status() is None:
```
MiaAltieri: Again, feel free to ignoe, but I think its better to reduce indentation:

```
if self.get_status():
    return
```

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -198,37 +198,22 @@ def on_update_status(self) -> None:
...
-            logger.info("mongos has not started yet")
-            self.charm.status_manager.to_waiting("Waiting for mongos to start.")
-            return
+            # in K8s mongos charms which are exposed externally it is possible for
+            # the node port to change. This can invalidate our current
+            # certificates. when this happens we do not receive any notifications
+            # from Juju so we must monitor it and request TLS integration to update
+            # our SANS as necessary.
+            # The connection info will be updated when we receive the new certificates.
+            if self.substrate == Substrates.K8S:
```
MiaAltieri: don't we want to do this regardless of whether or not `if self.get_status() is None:`? I imagine a case where mongos is not running since the TLS needs to be updated

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -408,6 +393,24 @@ def is_mongos_running(self) -> bool:
 
         uri = f"mongodb://{host}"
 
-        return self.mongo_manager.mongod_ready(uri=uri, direct=False)
+        return self.mongo_manager.mongod_ready(uri=uri)
+
+    def get_status(self) -> StatusBase | None:
```
MiaAltieri: I think this name can be deceptive, maybe `get_sanity_check_statuses` with a docstring that says "Retrieve statuses that directly relate to states of mongos which would prevent other more advanced mongos statuses from being checked."

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Yeah but we don't exactly return in that case, because we then run the `process_and_share_statuses` method in all cases, that's why I did it that way.
WDYT?

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: That's very true
Let me change that

single_kernel_mongo/managers/mongos_operator.py
MiaAltieri: ah yeah- duh 🤦

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Done :)

MiaAltieri (approved):

PR #8, opened 2025-01-10
feat(upgrades): Implementation of upgrades

Gu1nness:
## 🏷 Type of changes

- [x] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

Feature parity: Add upgades

## 📑 Motivation and Context

In order to reach feature parity, this is the implementation of upgrades.

I'm not satisfied with this code that I find a bit ugly and still bears the weights of some decisions made for upgrades v2.
I hope to find time to improve it in the future!

Manual testing for now

## ✅ Checklist

- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.


single_kernel_mongo/core/abstract_upgrades.py
```diff
@@ -0,0 +1,758 @@
...
+        except KeyError as exception:
+            logger.debug(f"Version missing from {previous_versions=}", exc_info=exception)
+            return False
+
+    @abstractmethod
+    def _get_unit_healthy_status(self) -> StatusBase:
+        """Status shown during upgrade if unit is healthy."""
+        raise NotImplementedError()
+
+    def get_unit_juju_status(self) -> StatusBase | None:
```
MiaAltieri:
```suggestion
    def get_unit_juju_unit_status(self) -> StatusBase | None:
```

single_kernel_mongo/core/abstract_upgrades.py
```diff
@@ -0,0 +1,758 @@
...
+        logger.debug("Running pre-refresh checks")
+
+        if self.dependent.name == KindEnum.MONGOS:
+            if not self.state.db_initialised:
+                return
+            if not self.dependent.upgrade_manager.is_mongos_able_to_read_write():
+                raise PrecheckFailedError("mongos is not able to read/write")
+            return
+
+        # TODO: if shard is getting upgraded but BOTH have same revision, then fail
```
MiaAltieri, line 239: don't forget about this ;)

single_kernel_mongo/core/abstract_upgrades.py
```diff
@@ -0,0 +1,758 @@
...
+            ClusterNotHealthyError.
+        """
+        for attempt in Retrying(stop=stop_after_attempt(10), wait=wait_fixed(1)):
+            with attempt:
+                if not self.is_cluster_healthy():
+                    raise ClusterNotHealthyError()
+
+    def is_cluster_healthy(self: GenericMongoDBUpgradeManager[MongoDBOperator]) -> bool:
+        """Returns True if all nodes in the cluster/replica set are healthy."""
+        # TODO: check mongos
```
MiaAltieri, line 513: same here

single_kernel_mongo/core/version_checker.py
```diff
@@ -0,0 +1,140 @@
...
+        goal_state = self.charm.model._backend._run("goal-state", return_output=True, use_json=True)
+        for unit_name, unit_state in goal_state["units"].items():  # type: ignore
+            if unit_name == unit_to_ignore:
+                continue
+            if unit_state["status"] == "active":
+                continue
+            if not self.is_status_related_to_mismatched_revision(unit_state["status"]):
+                return False
+
+        return True
```
MiaAltieri: I know these use version checking, but do they belong here?

single_kernel_mongo/core/version_checker.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def are_shards_status_ready_for_upgrade(self) -> bool:
```
MiaAltieri: same for this function

single_kernel_mongo/events/lifecycle.py
```diff
@@ -157,15 +157,14 @@ def on_relation_changed(self, event: RelationChangedEvent):
         """Relation changed event."""
         try:
             self.dependent.on_relation_changed()
+            self.charm.status_manager.process_and_share_statuses()
```
MiaAltieri, line 165: in our current implementation of statuses this could be dangerous. (For the same reasons I stated on the statuses PR) 

I'm okay to keep it as is, since these will be revamped. But if you're having testing issues due to statuses this would be a good first spot to check

single_kernel_mongo/events/upgrades.py
```diff
@@ -0,0 +1,121 @@
...
+        except DeferrableError as e:
+            logger.info(ROLLBACK_INSTRUCTIONS)
+            defer_event_with_info_log(logger, event, "post cluster upgrade checks", str(e))
+
+    def _run_post_cluster_upgrade_task(self, event: _PostUpgradeCheckMongoDB):
+        try:
+            self.manager.run_post_cluster_upgrade_task()
+        except DeferrableError as e:
+            logger.info(ROLLBACK_INSTRUCTIONS)
+            defer_event_with_info_log(logger, event, "post cluster upgrade checks", str(e))
```
MiaAltieri, line 125: fyi these functions are missing return types

single_kernel_mongo/events/upgrades.py
```diff
@@ -0,0 +1,121 @@
...
+            event.fail(str(e))
+
+    def _run_post_app_upgrade_task(self, event: _PostUpgradeCheckMongoDB):
+        try:
+            self.manager.run_post_app_upgrade_task()
+        except DeferrableError as e:
+            logger.info(ROLLBACK_INSTRUCTIONS)
+            defer_event_with_info_log(logger, event, "post cluster upgrade checks", str(e))
+
+    def _run_post_cluster_upgrade_task(self, event: _PostUpgradeCheckMongoDB):
```
MiaAltieri:
```suggestion
    def _run_post_cluster_upgrade_task(self, event: _PostUpgradeCheckMongoDB):
        """After a sharded cluster has been upgraded, it is necessary to check that the entire cluster is healthy."""
```

single_kernel_mongo/managers/k8s.py
```diff
@@ -274,3 +282,30 @@ def _get_partition(self, *_) -> int:
...
+
+    @cache
+    def _get_revision(self, *_) -> str:
+        stateful_set = self.client.get(res=StatefulSet, name=self.app_name)
+        if not stateful_set.status or not stateful_set.status.updateRevision:
+            raise Exception("Incomplete stateful set")
+        return stateful_set.status.updateRevision
+
+    @cache
+    def _list_revisions(self, *_) -> dict[str, str]:
```
MiaAltieri, line 298: mind adding a short docstring with context on how this is useful

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -235,21 +245,7 @@ def on_start(self) -> None:
...
-
-        # Push TLS files if necessary
-        self.tls_manager.push_tls_files_to_workload()
-
-        # Update licenses
-        self.handle_licenses()
-
-        # Sets directory permissions
-        self.set_permissions()
+        self._configure_layers()
```
MiaAltieri: maybe a different name here? Layers makes me think "k8s" but it seems that this is used for both k8s and VM

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -277,16 +273,12 @@ def on_start(self) -> None:
             raise WorkloadNotReadyError
 
         try:
-            self.mongodb_exporter_config_manager.configure_and_restart()
+            self._restart_related_services()
         except WorkloadServiceError:
-            self.charm.status_manager.to_blocked("couldn't start mongodb exporter")
             return
```
MiaAltieri, line 297: log would be nice here

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -277,16 +273,12 @@ def on_start(self) -> None:
...
 
-        try:
-            self.backup_manager.configure_and_restart()
-        except WorkloadServiceError:
-            self.charm.status_manager.to_blocked("couldn't start pbm-agent")
-            return
+        if self.substrate == Substrates.K8S:
+            self.upgrade_manager._reconcile_upgrade()
```
MiaAltieri: can you add a comment, something like: 

upgrades through k8s result in the start hook getting executed.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -361,10 +353,13 @@ def on_relation_changed(self) -> None:
 
         Adds the unit as a replica to the MongoDB replica set.
         """
+        if self.substrate == Substrates.K8S:
```
MiaAltieri, line 411: I am forgetful can you remind me why this is needed for relation changed? Perhaps a comment for it would be good too

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -412,6 +413,9 @@ def on_secret_changed(self, secret_label: str, secret_id: str) -> None:
         self.mongodb_exporter_config_manager.configure_and_restart()
         self.backup_manager.configure_and_restart()
 
+        # Always process the statuses.
+        self.charm.status_manager.process_and_share_statuses()
```
MiaAltieri, line 473: same thing applied here as my previous comments.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -496,15 +501,7 @@ def on_storage_detaching(self) -> None:
...
-        if not self.backup_manager.is_valid_s3_integration():
-            self.charm.status_manager.to_blocked(INVALID_S3_INTEGRATION_STATUS)
-            return
-        # TODO: Cluster Mismatch revision.
-        if not self.cluster_manager.is_valid_mongos_integration():
-            self.charm.status_manager.to_blocked(
-                "Relation to mongos not supported, config role must be config-server"
-            )
-        if not self.state.db_initialised:
+        if not self.pass_status_basic_checks():
```
MiaAltieri, line 560: this makes me feel soothed, thank you

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -517,10 +514,17 @@ def on_update_status(self) -> None:
             self.charm.status_manager.to_waiting("Waiting for MongoDB to start")
             return
 
-        try:
-            self.perform_self_healing()
-        except ServerSelectionTimeoutError as e:
-            logger.info(f"Failed to perform self healing: {e}")
+        if self.substrate == Substrates.K8S:
+            self.upgrade_manager._reconcile_upgrade()
```
MiaAltieri, line 574: Ah, again sorry I forgot why do we need to check this in update status

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -517,10 +514,17 @@ def on_update_status(self) -> None:
...
+
+        # It's useless to try to perform self healing if upgrade is in progress
+        # as the handlers would raise an UpgradeInProgressError anyway so
+        # better skip it when possible.
```
MiaAltieri, line 578: so informative! I love it!

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -702,36 +741,34 @@ def is_relation_feasible(self, rel_name: str) -> bool:
             return False
         return True
 
+    def _configure_layers(self) -> None:
```
MiaAltieri: I would rename this, as layer implies k8s

single_kernel_mongo/managers/sharding.py
```diff
@@ -207,11 +217,11 @@ def update_ca_secret(self, new_ca: str | None) -> None:
                 continue
             if new_ca is None:
                 self.data_interface.delete_relation_data(
-                    relation.id, [ConfigServerKeys.INT_CA_SECRET.value]
+                    relation.id, [AppShardingComponentKeys.INT_CA_SECRET.value]
                 )
                 continue
             self.data_interface.update_relation_data(
-                relation.id, {ConfigServerKeys.INT_CA_SECRET.value: new_ca}
+                relation.id, {AppShardingComponentKeys.INT_CA_SECRET.value: new_ca}
```
MiaAltieri, line 226: I like this

single_kernel_mongo/managers/upgrade.py
```diff
@@ -0,0 +1,352 @@
...
+
+        On K8S, during an upgrade event, it will set the version in all relations,
+        replan the container and process the upgrade statuses. If the upgrade
+        is compatible, it will end up emitting a post upgrade event that
+        verifies the health of the cluster.
+        On VM, during an upgrade event, it will call the reconcile upgrade
+        after setting the version across all relations.
+        """
+        if self.dependent.substrate == Substrates.VM:
+            if self.charm.unit.is_leader():
```
MiaAltieri, line 56: nit I know you and I feel differently about nesting, so feel free to ignore, but you could change this to 
```
    return

rest of what you have
```

single_kernel_mongo/managers/upgrade.py
```diff
@@ -0,0 +1,352 @@
...
+        if self._upgrade.is_compatible:
+            # Post upgrade event verifies the success of the upgrade.
+            self.dependent.upgrade_events.post_app_upgrade_event.emit()
+
+    def on_pre_upgrade_check_action(self) -> None:
+        """Pre upgrade checks."""
+        if not self.charm.unit.is_leader():
+            message = f"Must run action on leader unit. (e.g. `juju run {self.charm.app.name}/leader {UpgradeActions.PRECHECK_ACTION_NAME.value}`)"
+            raise ActionFailedError(message)
+        if not self._upgrade or self.state.upgrade_in_progress:
```
MiaAltieri, line 95: If `not self._upgrade`, that means there was a  `PeerRelationNotReadyError` in `_upgrade `, does that really mean that a  "Refresh already in progress"? I think it means "Upgrade relation not yet present"

Perhaps failing with that error would be better

single_kernel_mongo/managers/upgrade.py
```diff
@@ -0,0 +1,352 @@
...
+        event.log(f"Forcefully refreshing {self.charm.unit.name}")
+        # This happens only on mongodb VM
+        self._upgrade.upgrade_unit(dependent=self.dependent)  # type: ignore
+        logger.debug("Forced refresh")
+        return f"Forcefully refreshed {self.charm.unit.name}"
+
+    # HELPERS
+
+    def _set_upgrade_status(self):
+        if self.charm.unit.is_leader():
```
MiaAltieri: for my understanding,: why is this helper here and not in abstract_upgrades.py with the rest of the helpers?

single_kernel_mongo/managers/upgrade.py
```diff
@@ -0,0 +1,352 @@
...
+
+        # Leader of config-server must wait for all shards to be upgraded before finalising the
+        # upgrade.
+        if not self.charm.unit.is_leader() or not self.state.is_role(MongoDBRoles.CONFIG_SERVER):
+            return
+
+        self.dependent.upgrade_events.post_cluster_upgrade_event.emit()
+
+    # Unused parameter only present for typing.
+    def run_post_upgrade_checks(self, finished_whole_cluster: bool = False) -> None:
```
MiaAltieri, line 249: I feel like this could also go in abstract_upgrades.py?

single_kernel_mongo/state/charm_state.py
```diff
@@ -280,10 +365,62 @@ def db_initialised(self) -> bool:
...
+            return self.k8s_manager.get_revision()
+        return SNAP.revision
+
     @property
     def upgrade_in_progress(self) -> bool:
         """Is the charm in upgrade?"""
-        return False
+        app_version = self.app_workload_container_version
+        unit_versions = self.unit_workload_container_versions
+        logger.debug(f"{app_version} {unit_versions}")
```
MiaAltieri: I think this log can be removed

single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -626,3 +662,8 @@ def is_shard_aware(self, shard_name: str) -> bool:
                 return shard["state"] == SHARD_AWARE_STATE
 
         return False
+
+    def are_all_shards_aware(self) -> bool:
+        """Returns True if all shards are shard aware."""
+        sc_status = self.client.admin.command("listShards")
+        return all(shard["state"] == SHARD_AWARE_STATE for shard in sc_status["shards"])
```
MiaAltieri: Note: this checks that all shards in `sc_status` are shard aware, in theory it could be possible that we have a shard that is integrated to the config-server but is not yet in `sc_status`

tests/unit/mongodb_test_charm/charmcraft.yaml
```diff
@@ -30,3 +30,5 @@ parts:
     source: .
     prime:
       - workload_version
+      - charm_version
+      - charm_internal_version
```
MiaAltieri: note we will be able to remove this in the future

MiaAltieri (commented):
Amazing work here Neha! thank you!

single_kernel_mongo/core/abstract_upgrades.py
Gu1nness: If I have some time because this comment is copy pasted from the former upgrades but I'll try

single_kernel_mongo/core/abstract_upgrades.py
Gu1nness: Trying to find something better without the `unit` duplication.

single_kernel_mongo/core/abstract_upgrades.py
Gu1nness: Oh yeah right but same this is copy pasted so I'll try to find a way.

single_kernel_mongo/core/version_checker.py
Gu1nness: I think it could belong in upgrades yes.
I'd rather not having to use it at all if possible however but it would require a bit more investigations.
WDYT?

single_kernel_mongo/events/lifecycle.py
Gu1nness: I agree and I have the feeling that for now I failed to merge the upgrade status with the regular status processing, we'll see how it goes.

single_kernel_mongo/managers/k8s.py
Gu1nness: Sure.
For future reference I add it here as well, this map is used in upgrades to get a "version" for each container. This is used to see if there's an upgrade in progress.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Yeah sure, you're right.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: You're right.
(Also, k8s upgrades are really such a pain and breaking the whole cleanliness of the code :sob: )

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: It's not me, it's [the spec](https://docs.google.com/document/d/1tLjknwHudjcHs42nzPVBNkHs98XxAOT2BXGGpP7NyEU/edit?tab=t.0)
I'll add a comment here

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: We'll see how it goes :)

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: My pleasure!
It makes the code easier to read and reusable!

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Same as before: the spec says it.
(And this really makes me want at some point to have the core of this charm to be a holistic charm so all this code lives in a big and smart and lovely function)

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: This has been renamed with the former comment to `_configure_workloads` which I hope is more generic and meaningful.

single_kernel_mongo/managers/upgrade.py
Gu1nness: Nah you're right indeed.

Gu1nness: Ah nah again in fact.
It doesn't work well because the upgrade` event on k8s runs always and it runs on the upgrade of each unit, not all unit when refreshing.
Breaks my heart though.

single_kernel_mongo/managers/upgrade.py
Gu1nness: It should and it's a miss :)

single_kernel_mongo/managers/upgrade.py
Gu1nness: I'll check

Gu1nness: I think I can add the interface in the abstract upgrades but this is still specific to the charm kind

Gu1nness: There's already an interface :)

single_kernel_mongo/state/charm_state.py
Gu1nness: It might be better to rework it to something better and keep it so it's easier to understand if some upgrades are stuck.
I'll rework that.

single_kernel_mongo/utils/mongo_connection.py
Gu1nness: This is correct.
It's a bit hard however: This check is run independently by config server (which can know who are the other shards) and by the shards (who don't know).
I think we're safe however because hopefully the refresh check on the config server would catch it but I'll add a comment on the function.

tests/unit/mongodb_test_charm/charmcraft.yaml
Gu1nness: Yes and I'll start first with the internal version following your excellent work on the current charms :)

single_kernel_mongo/utils/mongo_connection.py
MiaAltieri: ack! Thank you!

single_kernel_mongo/core/abstract_upgrades.py
```diff
@@ -336,11 +337,16 @@ def run_post_upgrade_checks(self, finished_whole_cluster: bool = False) -> None:
...
     # BEGIN: Helpers
+    @mongodb_only
     def move_primary_to_last_upgrade_unit(self) -> None:
         """Moves the primary to last unit that gets upgraded (the unit with the lowest id).
 
         Raises FailedToMovePrimaryError
         """
+        # This should only ever run on mongos
+        if self.dependent.name == CharmKind.MONGOS:
+            logger.error("move_primary_to_last_upgrade_unit called on mongos charm. Please fix.")
```
MiaAltieri: isn't this a mistake on the developers side, what should the user do if they see this exception?

single_kernel_mongo/utils/helpers.py
```diff
@@ -42,3 +48,21 @@ def hostname_from_shardname(host: str) -> str:
...
+
+
+def charm_kind_only(func, charm_kind: CharmKind):
+    """Helpful decorator to ensure the charm kind."""
+
+    def wrapper(self, *args, **kwargs):
+        if self.dependent.name != charm_kind:
+            logger.error(f"{func} called on a non {charm_kind.value} charm. Please fix.")
+            raise InvalidCharmKindError(
+                f"{func} called on a non {charm_kind.value} charm. Please fix."
```
MiaAltieri: isn't this a mistake on the developers side, what should the user do if they see this exception?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -647,23 +647,30 @@ def perform_self_healing(self) -> None:
...
             for relation in self.state.client_relations:
                 self.mongo_manager.update_app_relation_data(relation)
+            return
 
         # Update the mongos host in the sharded deployment
-        self.config_server_manager.update_mongos_hosts()
-        # Update the config server DB URI on the remote mongos
-        self.cluster_manager.update_config_server_db()
+        if self.state.is_role(MongoDBRoles.SHARD):
+            self.shard_manager.update_mongos_hosts()
```
MiaAltieri: Mind explaining why a shard needs to update mongos hosts?  shards shouldn't have controll of mongos, but maybe I'm missing something here

single_kernel_mongo/managers/sharding.py
```diff
@@ -472,6 +472,10 @@ def synchronise_cluster_secrets(self, relation: Relation, leaving: bool = False)
             logger.info("Skipping relation changed event: hook checks did not pass.")
             raise
 
+        if not self.state.shard_state.has_received_credentials():
+            logger.debug("Waiting for config-server to write in the databag.")
```
MiaAltieri: nit:             logger.debug("Waiting for config-server to cluster secrets.")

MiaAltieri (commented):
Looks great! minor comments!

single_kernel_mongo/core/abstract_upgrades.py
MiaAltieri: Ah - then NW, maybe we can create a JIRA ticket and ignore it

single_kernel_mongo/core/abstract_upgrades.py
MiaAltieri: Ah - then NW, maybe we can create a JIRA ticket and ignore it

single_kernel_mongo/core/version_checker.py
MiaAltieri: thanks for moving it. And yes, when we rework statuses lets consider our options for removing this entirely.

single_kernel_mongo/managers/upgrade.py
Gu1nness: You're right

single_kernel_mongo/core/abstract_upgrades.py
Gu1nness: Actually, I have implemented some decorators for that and I just forgot to remove this check here…
Thanks, I'm cleaning this up.

single_kernel_mongo/utils/helpers.py
Gu1nness: You're right.
"Please contact the developers/support" is better?

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: It's updating it's own version of mongos hosts but I added this incorrectly, we we're poorly handling the relation changed event as I told you yesterday, let me remove this patch.

single_kernel_mongo/utils/helpers.py
MiaAltieri: sure, but I think: "Unexpected {func} called on a non {charm_kind.value} charm." is good enough :) 

otherwise we set a precedent for when users should/shouldn't contact decs/support. but also I trust you judgement!

single_kernel_mongo/managers/k8s.py
```diff
@@ -281,7 +281,7 @@ def _get_partition(self, *_) -> int:
             not partition.spec
             or not partition.spec.updateStrategy
             or not partition.spec.updateStrategy.rollingUpdate
-            or not partition.spec.updateStrategy.rollingUpdate.partition
+            or partition.spec.updateStrategy.rollingUpdate.partition is None
```
MiaAltieri:
```
or partition.spec.updateStrategy.rollingUpdate.partition is None # partition can be == 0 but we want to check for None
```

single_kernel_mongo/utils/helpers.py
Gu1nness: You're right

single_kernel_mongo/core/abstract_upgrades.py
MiaAltieri: feel free to add the JIRA ticket to the comment in the code and then just resolve this comment :)

single_kernel_mongo/config/literals.py
```diff
@@ -108,7 +108,7 @@ class VmUser(WorkloadUser[int]):
 
 MAX_PASSWORD_LENGTH = 4096
 
-PBM_RESTART_DELAY = 5
+RESTART_DELAY = 5
```
MiaAltieri: why not `PBM_RESTART_DELAY `, like it was a few commits ago? I think thats more informative

single_kernel_mongo/managers/k8s.py
MiaAltieri: thanks for making this change!

single_kernel_mongo/core/abstract_upgrades.py
Gu1nness: Sure : [ticket here](https://warthogs.atlassian.net/browse/DPE-6397)

single_kernel_mongo/config/literals.py
Gu1nness: Yes it's a miss from my side :sweat_smile: fixing this.

single_kernel_mongo/config/literals.py
```diff
@@ -98,3 +109,11 @@ class VmUser(WorkloadUser[int]):
 MAX_PASSWORD_LENGTH = 4096
 
 PBM_RESTART_DELAY = 5
+
+FEATURE_VERSION_6 = "6.0"
+
+UNHEALTHY_UPGRADE = BlockedStatus("Unhealthy after refresh.")
```
Mehdi-Bendriss, line 115: I cannot remember exactly where, but was there not a module where all statuses were listed?

single_kernel_mongo/managers/mongodb_operator.py
Mehdi-Bendriss: A function that's big and smart is rarely lovely 😄

single_kernel_mongo/core/abstract_upgrades.py
```diff
@@ -0,0 +1,910 @@
...
+
+    @unit_state.setter
+    def unit_state(self, value: UnitState) -> None:
+        self.state.unit_upgrade_peer_data.unit_state = value
+
+    @property
+    def is_compatible(self) -> bool:
+        """Whether upgrade is supported from previous versions."""
+        previous_version_strs = self.state.app_upgrade_peer_data.versions
+        if not previous_version_strs:
```
Mehdi-Bendriss: nit: walrus

single_kernel_mongo/core/abstract_upgrades.py
```diff
@@ -0,0 +1,910 @@
...
+        previous_versions: dict[str, poetry_version.Version] = {
+            key: poetry_version.Version.parse(value) for key, value in previous_version_strs.items()
+        }
+        current_version_strs = copy.copy(self._current_versions)
+        current_version_strs["charm"] = current_version_strs["charm"].split("+")[0]
+        current_versions = {
+            key: poetry_version.Version.parse(value) for key, value in current_version_strs.items()
+        }
+        try:
+            # TODO Future PR: change this > sign to support downgrades
```
Mehdi-Bendriss, line 135: downgrades or rollbacks?

single_kernel_mongo/core/abstract_upgrades.py
```diff
@@ -0,0 +1,910 @@
...
+
+        if self.state.is_role(MongoDBRoles.CONFIG_SERVER):
+            if not self.dependent.upgrade_manager.are_pre_upgrade_operations_config_server_successful():
+                raise PrecheckFailedError("Pre-refresh operations on config-server failed.")
+
+
+# END: Useful classes
+
+
+class GenericMongoDBUpgradeManager(Generic[T], Object, ABC):
```
Mehdi-Bendriss, line 270: missing docstrings and return types in methods of this class

single_kernel_mongo/core/abstract_upgrades.py
Mehdi-Bendriss: Could we create a ticket and track it as well, we have a few of these already - and we should make sure they don't get forgotten on the backlog during this refactor

single_kernel_mongo/core/machine_upgrades.py
```diff
@@ -0,0 +1,171 @@
...
+
+    @unit_state.setter
+    def unit_state(self, value: UnitState) -> None:
+        # Super call
+        AbstractUpgrade.unit_state.fset(self, value)  # type: ignore[attr-defined]
+
+    def _get_unit_healthy_status(self) -> StatusBase:
+        if self.state.unit_workload_container_version == self.state.app_workload_container_version:
+            return ActiveStatus(
+                f'MongoDB {self._unit_workload_version} running; Snap revision {self.state.unit_workload_container_version}; Charm revision {self._current_versions["charm"]}'
```
Mehdi-Bendriss: can you break down this text and the next one on multilines, they're too long and not so straightforward on an IDE

single_kernel_mongo/managers/config.py
```diff
@@ -213,6 +213,7 @@ def binding_ips(self) -> list[str]:
         """The binding IP parameters."""
         if (
```
MiaAltieri, line 218: ```# for VM mongos charms we bind to the socket (when non-external), this gives us one less network hop when communicating with the client ```

Mehdi-Bendriss (commented):
Thanks Neha, it looks good overall, left minor comments.
I haven't reviewed too profoundly, as a lot of it is the core logic of the upgrades which has been exhaustively reviewed previously by Mia and Carl. 
I trust Mia's review on this PR.

MiaAltieri (approved):

single_kernel_mongo/config/literals.py
Gu1nness: No there is not yet

single_kernel_mongo/core/abstract_upgrades.py
Gu1nness: This comment comes from the current charms, but given the code, I think it means downgrade (because we compare the version with the "overall version" which is stored in a DB

single_kernel_mongo/core/abstract_upgrades.py
Gu1nness: Sure

Gu1nness: https://warthogs.atlassian.net/browse/DPE-6413

Mehdi-Bendriss (approved):

PR #9, opened 2025-01-20
feat: Observability manager

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

Integration of observability stacks.

## 📑 Motivation and Context

Full feature parity	

Needs to be tested manually.

## ✅ Checklist


single_kernel_mongo/managers/observability.py
```diff
@@ -0,0 +1,107 @@
...
+class ObservabilityManager(Object):
+    """Include the right observability stack."""
+
+    def __init__(
+        self,
+        dependent: MongoDBOperator,
+        state: CharmState,
+        substrate: Substrates,
+    ) -> None:
+        self.dependent = dependent
```
Mehdi-Bendriss, line 44: missing `super` call

single_kernel_mongo/managers/observability.py
```diff
@@ -0,0 +1,107 @@
...
+                self.charm,
+                log_files=[
+                    f"{self.dependent.workload.paths.log_file}",
+                    f"{self.dependent.workload.paths.audit_file}",
+                ],
+                relation_name=ExternalProviderRelations.LOGGING.value,
+                container_name=self.dependent.role.name,
+            )
+
+    def vm_mongo_scrape_config(self):
```
Mehdi-Bendriss: 1. missing return type 
2. should it be a `@property`?

Mehdi-Bendriss (commented):
Thanks Neha, LGTM - minor comments

single_kernel_mongo/managers/observability.py
Gu1nness: Oh yeah it can and then it will make the whole code a bit smaller (I had to dig in the COSAgentProvider code to ensure that)

Mehdi-Bendriss (approved):

single_kernel_mongo/config/relations.py
```diff
@@ -42,3 +42,4 @@ class ExternalProviderRelations(str, Enum):
     """The relations we provide to non mongo related charms."""
 
     COS_AGENT = "cos-agent"
+    LOGGING = "logging"
```
MiaAltieri: note: 
`cos-agent` is only used by VM
`logging`(loki_push_api)  relation is only used by K8s
`grafana_dashboard` is a cos relation that is only used by K8s. I think its missing here 
`prometheus_scrape` is a cos relation that is only used by K8s. I think its missing here

single_kernel_mongo/managers/observability.py
```diff
@@ -0,0 +1,95 @@
...
+                        "targets": [
+                            f"{self.state.unit_peer_data.internal_address}:{OBSERVABILITY_CONFIG.mongodb_exporter_port}"
+                        ],
+                        "labels": {
+                            "cluster": self.state.config_server_name or self.charm.app.name,
+                        },
+                    }
+                ],
+            }
+        ]
```
MiaAltieri, line 94: question is the following already done (by different PR):

- rotate monitor user password
- start mongodb_exporter service

I feel like I remember it being done already, but I wanted to double check

single_kernel_mongo/config/relations.py
Gu1nness: I don't think we need to specify the other relations that's why I did not add it there, and actually the `COS_AGENT` relation should go away if it's unused.
WDYT, we should rather write all relations even if unused, or only the ones used?

single_kernel_mongo/managers/observability.py
Gu1nness: Rotate monitor password can be done with the `set-password` action yes.
Start mongodb_exported service is done as well.

single_kernel_mongo/config/relations.py
MiaAltieri: if the constant `COS_AGENT ` is not used it can be removed
if the constant `LOGGING` is used, it might be nice to add a comment # note this relation is only used for K8s charms

Gu1nness: This is now done

MiaAltieri (approved):

PR #10, opened 2025-01-20
feat(DPE-6396): mongo error codes

Gu1nness:
* Collect all mongo error codes in an enum.


Mehdi-Bendriss (approved):
Thank you! No strong opinion, should we perhaps put this into the `literals` module?

Gu1nness:
> Thank you! No strong opinion, should we perhaps put this into the `literals` module?

That was my first guess but then I thought it would be better to keep it with the mongo code (which will also make it a bit easier if the list starts to grow)

MiaAltieri (approved):

PR #17, opened 2025-02-21
tests: avoid patching private ops testing

james-garner-canonical:
## 🏷 Type of changes

- [ ] Documentation Update
- [x] Tooling and CI changes

## 📝 Description
- The purpose of this PR is to allow the library's tests to run with the latest versions of `ops`. Currently, the unit tests patch a private component of `ops.testing`, which has changed in `ops`, breaking the tests. This PR updates the tests to avoid patching this component by removing `patch_network_get` and using the default network values that harness sets. See also https://github.com/canonical/mongodb-operator/issues/484
- This PR impacts this library's unit tests only.

## 📑 Motivation and Context
We would like to run the unit tests of charms that depend on this library against the latest version of `ops` in our CI, to avoid breaking changes in future `ops` versions. Currently we are unable to do this, because this library specifies a maximum `ops` version of `<2.16`, and also this library's tests do not pass on the latest `ops` version (due to changes to private components of `ops.testing` which the unit tests patch). This PR updates the tests.

I believe that the tests in [mongodb-operator](https://github.com/canonical/mongodb-operator/blob/6/edge/tests/unit/helpers.py#L31) and [mongodb-k8s-operator](https://github.com/canonical/mongodb-k8s-operator/blob/6/edge/tests/unit/helpers.py#L29) need to be updated in the same way as this PR. Would someone from the team be able to handle that?

When the tests have been updated, I think that this library will be able to specify a wider range of supported `ops` versions -- perhaps `~=2.15` instead of the current `~=2.15.0` (which limits the library to `<2.16`), after which we will be able to run the mongodb charms' unit tests in our CI.

I've made a [dummy PR](https://github.com/james-garner-canonical/mongo-single-kernel-library/pull/2
) in my fork which changes the tests and also uses the latest version of `ops`. The tests pass. Without these changes to the unit tests, they fail when run with the latest `ops` release (see [here](https://github.com/james-garner-canonical/mongo-single-kernel-library/actions/runs/13446847736/job/37573944057) for an example test run with `ops==2.18.1` and no changes to the tests).

## ✅ Checklist


Gu1nness:
Thanks a lot for the PR!
Everything looks good to me, and I'll take time to submit similar PRs on the 4 charms unit testing and then we can proceed to update ops!

Gu1nness (approved):

MiaAltieri (approved):
This is such a great improvement @james-garner-canonical ! Thank you so much! Quick question, if we need to mock a specific type of ip getting returned is that possible? If so how? :)

Thanks again!

james-garner-canonical:
- @Gu1nness Thanks for offering to take care of the charm unit tests -- I somehow hadn't noticed that there are actually 4 charms (not just 2) using this library. I wonder if a monorepo for the charms + library would be a good idea someday? Just mentioning it in case the idea hasn't been suggested (it seems to work well for the [sunbeam charms](https://opendev.org/openstack/sunbeam-charms)).

- @MiaAltieri I think that the way to mock a specific ip address would be to set up a network using [Harness.add_network](https://ops.readthedocs.io/en/latest/reference/ops-testing-harness.html#ops.testing.Harness.add_network). I'm not very experienced with `Harness` though, I'd say @tonyandrewmeyer is our expert there (this PR is based on his suggestions in the mobgodb-operator issue https://github.com/canonical/mongodb-operator/issues/484).

Gu1nness:
> * @Gu1nness Thanks for offering to take care of the charm unit tests -- I somehow hadn't noticed that there are actually 4 charms (not just 2) using this library. I wonder if a monorepo for the charms + library would be a good idea someday? Just mentioning it in case the idea hasn't been suggested (it seems to work well for the [sunbeam charms](https://opendev.org/openstack/sunbeam-charms)).

This is something I'd love to do (a monorepo) but from what I understood with @Mehdi-Bendriss we're not there yet (not on the library side, but it seems there are some challenges that we need to handle first.

Mehdi-Bendriss (approved):
Thank you James! 

On the mono repo, this is definitely the direction we're following, we however still have divergence in the integration tests depending on the charm and substrate, once we resolve that - we will effectively be ready for a full "single mono repo" approach for mongodb

PR #22, opened 2025-03-04
feat: Add config options for LDAP

Gu1nness:
## 🏷 Type of changes

- [x] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

* Adds the two new config options for LDAP integration.
* Adds the validators for those two options and their combinations.
* Adds some unit tests for the config changed events.

## 📑 Motivation and Context

As a part of [MongoDB LDAP integration](https://docs.google.com/document/d/1J483J987gaY1znCQE79VEucwgdWO1RkWa5gRdkGpOT0) two new config options are required, along with their validation and tests.

This is unit tested.
It will be integration tested in the charms when the rest of the code is ready.

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/core/structured_config.py
+++ b/single_kernel_mongo/core/structured_config.py
@@ -76,12 +76,12 @@ class SingleLdapUserToDnMapping(BaseModel):
 
     match: str
     substitution: str | None = None
-    ldapQuery: str | None = None  # noqa: N815
+    ldap_query: str | None = Field(default=None, alias="ldapQuery")  # noqa: N815
 
     @model_validator(mode="after")
     def verify_subst_or_ldapquery(self) -> Self:
         """Ensures that only one of substitution or ldapQuery is present."""
-        if self.substitution and self.ldapQuery or (not self.substitution and not self.ldapQuery):
+        if self.substitution and self.ldap_query or (not self.substitution and not self.ldap_query):
             raise ValueError("substitution and ldapQuery are mutually exclusive")
         return self
 
--- a/single_kernel_mongo/core/structured_config.py
+++ b/single_kernel_mongo/core/structured_config.py
@@ -10,7 +10,16 @@ Modifiable configurations should be defined in `config.yaml` in each charm.
 from enum import Enum
 from typing import Annotated, TypeVar
 
-from pydantic import BaseModel, ConfigDict, Field, PlainSerializer, field_validator
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    Field,
+    PlainSerializer,
+    TypeAdapter,
+    field_validator,
+    model_validator,
+)
+from typing_extensions import Self
 
 # Generic TypeVar for serializers
 T = TypeVar("T")
@@ -62,6 +71,24 @@ class ExposeExternal(str, Enum):
         return {cls.NODEPORT, cls.NONE}
 
 
+class SingleLdapUserToDnMapping(BaseModel):
+    """Validating the Mapping structure."""
+
+    match: str
+    substitution: str | None = None
+    ldapQuery: str | None = None  # noqa: N815
+
+    @model_validator(mode="after")
+    def verify_subst_or_ldapquery(self) -> Self:
+        """Ensures that only one of substitution or ldapQuery is present."""
+        if self.substitution and self.ldapQuery or (not self.substitution and not self.ldapQuery):
+            raise ValueError("substitution and ldapQuery are mutually exclusive")
+        return self
+
+
+LdapUserToDnMapping = TypeAdapter(list[SingleLdapUserToDnMapping])
+
+
 # NewType for typing (ghost type)
 class MongoConfigModel(BaseConfigModel):
     """Default class for typing."""
@@ -71,6 +98,8 @@ class MongoConfigModel(BaseConfigModel):
     )
     role: SerializeLiteralAsStr[MongoDBRoles]
     auto_delete: bool = Field(default=False, alias="auto-delete")
+    ldap_user_to_dn_mapping: str | None = Field(default=None, alias="ldap-user-to-dn-mapping")
+    ldap_query_template: str | None = Field(default=None, alias="ldap-query-template")
 
     @field_validator("expose_external", mode="before")
     @classmethod
--- a/single_kernel_mongo/events/lifecycle.py
+++ b/single_kernel_mongo/events/lifecycle.py
@@ -45,6 +45,8 @@ from single_kernel_mongo.config.relations import PeerRelationNames
 from single_kernel_mongo.core.operator import OperatorProtocol
 from single_kernel_mongo.exceptions import (
     ContainerNotReadyError,
+    InvalidLdapQueryTemplateError,
+    InvalidLdapUserToDnMappingError,
     UpgradeInProgressError,
     WorkloadServiceError,
 )
@@ -132,6 +134,9 @@ class LifecycleEventsHandler(Object):
             self.dependent.on_config_changed()
         except UpgradeInProgressError:
             event.defer()
+            return
+        except (InvalidLdapUserToDnMappingError, InvalidLdapQueryTemplateError) as err:
+            self.charm.status_manager.to_blocked(f"{err}")
             return
 
     def on_update_status(self, event: UpdateStatusEvent):
--- a/single_kernel_mongo/events/lifecycle.py
+++ b/single_kernel_mongo/events/lifecycle.py
@@ -48,6 +48,7 @@ from single_kernel_mongo.exceptions import (
     InvalidLdapQueryTemplateError,
     InvalidLdapUserToDnMappingError,
     UpgradeInProgressError,
+    WaitingForLeaderError,
     WorkloadServiceError,
 )
 from single_kernel_mongo.utils.mongo_connection import NotReadyError
@@ -132,7 +133,7 @@ class LifecycleEventsHandler(Object):
         """Config Changed Event."""
         try:
             self.dependent.on_config_changed()
-        except UpgradeInProgressError:
+        except (UpgradeInProgressError, WaitingForLeaderError):
             event.defer()
             return
         except (InvalidLdapUserToDnMappingError, InvalidLdapQueryTemplateError) as err:
--- a/single_kernel_mongo/managers/mongodb_operator.py
+++ b/single_kernel_mongo/managers/mongodb_operator.py
@@ -48,6 +48,7 @@ from single_kernel_mongo.exceptions import (
     ContainerNotReadyError,
     EarlyRemovalOfConfigServerError,
     FailedToElectNewPrimaryError,
+    InvalidLdapQueryTemplateError,
     InvalidLdapUserToDnMappingError,
     NonDeferrableFailedHookChecksError,
     SetPasswordError,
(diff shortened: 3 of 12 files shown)


single_kernel_mongo/core/structured_config.py
```diff
@@ -62,6 +71,24 @@ def valid_roles(cls) -> set[str]:
...
 
+class SingleLdapUserToDnMapping(BaseModel):
+    """Validating the Mapping structure."""
+
+    match: str
+    substitution: str | None = None
+    ldapQuery: str | None = None  # noqa: N815
+
+    @model_validator(mode="after")
+    def verify_subst_or_ldapquery(self) -> Self:
```
MiaAltieri, line 82: nit: should this return self? 

The main directive of the this function says it verifies, so not sure if we also want to use it as a getter. If we do then perhaps changing it to `get_and_verify_[...]`

single_kernel_mongo/events/lifecycle.py
```diff
@@ -133,6 +135,9 @@ def on_config_changed(self, event: ConfigChangedEvent):
         except UpgradeInProgressError:
             event.defer()
             return
+        except (InvalidLdapUserToDnMappingError, InvalidLdapQueryTemplateError) as err:
+            self.charm.status_manager.to_blocked(f"{err}")
```
MiaAltieri, line 140: I assume elsewhere in the code (i.e. the function that raises these exceptions) logs informative messages

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -362,7 +368,29 @@ def on_config_changed(self) -> None:
...
+            raise InvalidLdapQueryTemplateError(
+                "Invalid LDAP Query template, please update your config."
+            )
+
+        if self.charm.unit.is_leader():
+            # Store in the databag so we never miss it.
+            self.state.app_peer_data.ldap_user_to_dn_mapping = self.config.ldap_user_to_dn_mapping
+            # TODO: Send this to mongos as well.
+
+            self.state.app_peer_data.ldap_query_template = self.config.ldap_query_template
```
MiaAltieri: do we want to proceed in the case of upgrades? Or does it not matter? I think we don't but also I am curious what you think

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -362,7 +368,29 @@ def on_config_changed(self) -> None:
...
+            raise InvalidLdapUserToDnMappingError(
+                "Invalid LdapUserToDnMapping, please update your config."
+            )
+
+        if self.config.ldap_query_template and not validate_ldap_options(
+            self.config.ldap_user_to_dn_mapping, self.config.ldap_query_template
+        ):
+            raise InvalidLdapQueryTemplateError(
+                "Invalid LDAP Query template, please update your config."
+            )
```
MiaAltieri: probably wise to include this (and the previous checks) in their own separate function (sainity hooks check)

single_kernel_mongo/utils/helpers.py
```diff
@@ -57,6 +60,33 @@ def unit_number(unit: UnitUpgradePeerData) -> int:
     return int(unit.component.name.split("/")[-1])
 
 
+def validate_ldapusertodnmapping(ldap_user_to_dn_mapping: str) -> bool:
```
MiaAltieri: nit style/readability: for these types of functions its nicer to have it titled as `is_valid_[...]`

single_kernel_mongo/utils/helpers.py
```diff
@@ -57,6 +60,33 @@ def unit_number(unit: UnitUpgradePeerData) -> int:
...
+    """Validates the mapping, returning a boolean."""
+    try:
+        LdapUserToDnMapping.validate_json(ldap_user_to_dn_mapping)
+        return True
+    except ValidationError:
+        return False
+
+
+def validate_ldap_options(ldap_user_to_dn_mapping: str, ldap_query_template: str) -> bool:
```
MiaAltieri: same as above

single_kernel_mongo/utils/helpers.py
```diff
@@ -57,6 +60,33 @@ def unit_number(unit: UnitUpgradePeerData) -> int:
...
+    """Validates the combination of the two LDAP options.
+
+    Rules are the following:
+     * if ldap_user_to_dn_mapping is empty then ldap_query_template MUST
+     contain the placeholder `{PROVIDED_USER}`.
+     * if ldap_user_to_dn_mapping is not empty then ldap_query_template MUST
+     contain the placeholder `{USER}`.
```
MiaAltieri, line 82: noob question. What would a valid empty config look like (i.e. user is not using ldap) and if its not covered in the cases below it might be a good idea to have a check empty config case

single_kernel_mongo/core/structured_config.py
Gu1nness: This is the interface of a model validator in `after` mode as stated by the [documentation](https://docs.pydantic.dev/latest/concepts/validators/#model-validators)
There are multiple types of model validators that run at different moments of the validation.
The `after` validators take an already parsed object and then add some additional checks. They can be seen as `post initialisation hooks` to make a parallel with dataclasses.

The documentation specifically says that the validated instance should be returned if validation succeeds :)

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: That's a very good question.
I think in case of upgrades indeed we should not proceed because we don't want to break anything.
Let me change that control flow :)

single_kernel_mongo/utils/helpers.py
Gu1nness: We never reach that point if the two are empty :)
This is because there are some checks for non-emptyness in the config-changed event handler.
It can make sense however to bring those checks here if you think it's a better flow.

single_kernel_mongo/events/lifecycle.py
Gu1nness: I will add some informative logs indeed, I forgot to do so.
They should refer to the configuration options of the charm which will provide some examples.

single_kernel_mongo/core/structured_config.py
MiaAltieri: TIL! Thank you for teaching me something new every day!

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -362,22 +368,45 @@ def on_config_changed(self) -> None:
...
         if self.state.is_role(MongoDBRoles.UNKNOWN):  # We haven't run the leader elected event yet.
             self.state.app_peer_data.role = self.config.role
-            return
 
-        if self.state.is_role(self.config.role):
-            return
+        if self.config.ldap_user_to_dn_mapping and not is_valid_ldapusertodnmapping(
+            self.config.ldap_user_to_dn_mapping
+        ):
+            logger.info("Please refer to the config option description.")
```
MiaAltieri: nit
```suggestion
            logger.error("Invalid ldap Config - Please refer to the config option description.")
```

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -362,22 +368,45 @@ def on_config_changed(self) -> None:
...
+            raise InvalidLdapUserToDnMappingError(
+                "Invalid LdapUserToDnMapping, please update your config."
+            )
+
+        if self.config.ldap_query_template and not is_valid_ldap_options(
+            self.config.ldap_user_to_dn_mapping, self.config.ldap_query_template
+        ):
+            logger.info("Please refer to the config option description.")
```
MiaAltieri: nit
```suggestion
            logger.error("Invalid LDAP query template - Please refer to the config option description.")
```

single_kernel_mongo/utils/helpers.py
MiaAltieri: Ah okay nice! 

I *think* we might want to copy those checks here. 

My reason is that: if we want to use this function elsewhere (idk - say update status to compute a LDAP status or something) we could have the check baked in here

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: See 347569b

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: See 347569b

single_kernel_mongo/utils/helpers.py
Gu1nness: Done in 347569b

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: If needed, I'll do this later :)

single_kernel_mongo/core/structured_config.py
```diff
@@ -62,6 +71,24 @@ def valid_roles(cls) -> set[str]:
         return {cls.NODEPORT, cls.NONE}
 
 
+class SingleLdapUserToDnMapping(BaseModel):
+    """Validating the Mapping structure."""
+
+    match: str
+    substitution: str | None = None
+    ldapQuery: str | None = None  # noqa: N815
```
Mehdi-Bendriss:
```suggestion
    ldap_query: str | None = None  # noqa: N815
```

single_kernel_mongo/core/structured_config.py
```diff
@@ -62,6 +71,24 @@ def valid_roles(cls) -> set[str]:
...
+
+    @model_validator(mode="after")
+    def verify_subst_or_ldapquery(self) -> Self:
+        """Ensures that only one of substitution or ldapQuery is present."""
+        if self.substitution and self.ldapQuery or (not self.substitution and not self.ldapQuery):
+            raise ValueError("substitution and ldapQuery are mutually exclusive")
```
Mehdi-Bendriss, line 85: we should perhaps stay aligned with the error handling of pydantic and raise instead a ValidationError

tests/unit/mongodb_k8s_test_charm/config.yaml
```diff
@@ -14,3 +14,21 @@ options:
...
+        match : "([^@]+)@([^@\\.]+)\\.example\\.com",
+        substitution: "CN={0},CN=Users,DC={1},DC=example,DC=com"
+      }]'
+  ldap-query-template:
+    default: ''
+    type: string
+    description: |
+      A RFC4516 formatted LDAP query URL template, which is used for authorization.
+      It must contain either `{USER}` representing the authenticated user, or `{PROVIDED_USER}` representing the supplied username (before authentication or LDAP transformation).
+      `{PROVIDED_USER}` should be used only if no ldapUserToDNMapping is provided.
```
Mehdi-Bendriss: Perhaps?
```suggestion
      `{PROVIDED_USER}` should only be used if no value in the `ldap-user-to-dn-mapping` config option is provided.
```

Mehdi-Bendriss (commented):
Thanks Neha! looks good, just minor suggestions

single_kernel_mongo/core/structured_config.py
Gu1nness: This requires a bit more changes because we don't have control on the structure but let me do something that works.

single_kernel_mongo/core/structured_config.py
Gu1nness: This will be converted to a ValidationError by pydantic!
See [documentation](https://docs.pydantic.dev/latest/concepts/validators/#raising-validation-errors) for the errors that will be converted into `ValidationError`.

Mehdi-Bendriss (dismissed):

MiaAltieri (dismissed):

Gu1nness:
/test/mongodb/vm

Gu1nness:
/test/mongodb/vm

Mehdi-Bendriss (dismissed):

MiaAltieri (dismissed):

Gu1nness:
Needs some updates, [CI run](https://github.com/canonical/mongodb-operator/actions/runs/13794115413/job/38581650934?pr=549) has shown that it leaves the config hook failing sometimes leaving the hook in error.

Gu1nness:
/test/mongodb/vm

MiaAltieri (approved):

Mehdi-Bendriss (approved):

PR #23, opened 2025-03-06
feat: Use config file instead of cli parameters

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

* Replaces passing configuration with CLI arguments by writing in files.
* Updates snap to use to revision 128
* Updates rock to use to sha256:7ddb80a3b5ddffa95704a8980fc11037ba1a23273a9805214bc42be9f507107f which supports config file by default

## 📑 Motivation and Context

As part of MongoDB LDAP integration, we will have to provide credentials to the service, to secure the binding with the LDAP server.
In order to secure those credentials, we allow to write them only in the config file, and in order to avoid having a mix of config options in the config file and in the cli, we move everything to the files.

This is unit tested.
I will test it manually with the test charms but I plan also to bring some mechanisms to make it easier to build the test charms and run integration testing directly in this repository.

## ✅ Checklist


single_kernel_mongo/core/workload.py
```diff
@@ -22,6 +22,7 @@ class MongoPaths:
     """Object to store the common paths for a mongodb instance."""
 
     def __init__(self, role: CharmSpec):
+        self._role = role
```
Mehdi-Bendriss: This doesn't seem used anywhere?

single_kernel_mongo/managers/config.py
```diff
@@ -59,6 +65,30 @@ def build_parameters(self) -> list[list[str]]:  # pragma: nocover
...
+    @abstractmethod
+    def build_config(self) -> dict[str, Any]:
+        """Builds the config dict."""
+
+    def set_environment(self):
+        """Write update parameters in the file."""
+        data = "\n".join(self.workload.read(self.file))
+        current_content = safe_load(data)
+
+        new_content = always_merger.merge(deepcopy(current_content), self.build_config())
```
Mehdi-Bendriss: I have 2 points here: 
1. Why do we need to perform a deepcopy here? I don't see an argument in its favor 
2. This method seems to only be safe on the first call where the base is the "default" config file, correct?

single_kernel_mongo/managers/config.py
```diff
@@ -190,30 +220,37 @@ def configure_and_restart(self):
...
-            self.tls_parameters,
-            self.log_options,
-            self.audit_options,
-        ]
+        return [[]]
+
+    @override
+    def build_config(self) -> dict[str, Any]:
+        return reduce(
+            always_merger.merge,
```
Mehdi-Bendriss, line 241: why is this needed here? why not a simple dict concatenation?

single_kernel_mongo/managers/config.py
```diff
@@ -224,59 +261,83 @@ def binding_ips(self) -> list[str]:
...
-                "--clusterAuthMode=x509",
-                "--tlsAllowInvalidCertificates",
-                f"--tlsClusterCAFile={self.workload.paths.int_ca_file}",
-                f"--tlsClusterFile={self.workload.paths.int_pem_file}",
-            ]
-        return cmd + [
-            "--clusterAuthMode=keyFile",
-            f"--keyFile={self.workload.paths.keyfile}",
-        ]
+            return always_merger.merge(
```
Mehdi-Bendriss, line 309: I think we can use python standard ways here, either:
1. return {**cmd, {"security": {} .....}}
2. return cmd | {"security": {} .....}

Mehdi-Bendriss (commented):
Thanks Neha! I left some comments

single_kernel_mongo/core/workload.py
Gu1nness: You're right, it's something I added at some point but forgot to remove afterwards

single_kernel_mongo/managers/config.py
Gu1nness: 1. The merger overwrites the data in the first item, which would break the comparison below to know if we need to update the config file.
2. This is completely true and I missed that. This only adds but never removes the configurations. Maybe it would be best to do a bit more work here for the final bits that are missed and just dump without merging.

single_kernel_mongo/managers/config.py
Gu1nness: Because the dicts are nested and different method bring some same structure at the first level and it would be overwritten.
Eg:
```pythoN3
self.binding_ips == {"net": {bindIpAll": True}}
self.port_parameter == {"net": {"port": 27017}}

self.binding_ips | self.port_parameter == {"net": {"port": 27017}}
always_merger.merge(self.binding_ips, self.port_parameter) == {"net": {"bindIpAll": True, "port": 27017}}
```

Mehdi-Bendriss: ah yes! you're right

Mehdi-Bendriss (dismissed):

Gu1nness:
My plan for the future will be to have a pydantic model for the mongodb config so it can be easier to manage :)

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (dismissed):

single_kernel_mongo/core/operator.py
```diff
@@ -248,3 +248,9 @@ def set_permissions(self) -> None:
                     f"{path}",
                 ]
             )
+
+        for path in (
+            self.workload.paths.config_file,
+            self.workload.paths.mongos_config_file,
+        ):
+            self.workload.exec(["chmod", "600", f"{path}"])
```
MiaAltieri, line 256: newb questions:

- should the snap do this, given that the snap sets up these files for us? 
- will people using the snap have to do this in order to get `mongod`/`mongos` working from the start? If so then this should be moved to the snap

single_kernel_mongo/managers/config.py
```diff
@@ -190,30 +219,37 @@ def configure_and_restart(self):
...
 
-class MongoConfigManager(CommonConfigManager, ABC):
+class MongoConfigManager(FileBasedConfigManager, ABC):
     """The common configuration manager for both MongoDB and Mongos."""
 
     auth: bool
 
     @override
     def build_parameters(self) -> list[list[str]]:
-        return [
```
MiaAltieri, line 200: nit, add context about design decision:
"""<about this function>

We aim to pass most config options inside the config file instead of as parameters. In earlier charm versions we pass parameters, however for security reasons in LDAP params we have now decided to set these in the config file.
"""

single_kernel_mongo/managers/config.py
```diff
@@ -7,10 +7,15 @@
 import logging
 import time
 from abc import ABC, abstractmethod
+from functools import reduce
 from itertools import chain
+from pathlib import Path
+from typing import Any
 
+from deepmerge import always_merger
```
MiaAltieri, line 15: TIL

single_kernel_mongo/core/operator.py
Gu1nness: They won't have to do it no.
This is purely a question of security in this case because it will contains some secrets that we want to allow only the owner to be able to read.
In the other cases it's not needed.

single_kernel_mongo/managers/config.py
Gu1nness: Indeed, very good idea!

single_kernel_mongo/managers/config.py
Gu1nness: Me too tbh, I looked a while to find a good merging library :)

single_kernel_mongo/core/operator.py
MiaAltieri: Good Explanation - thank you!

MiaAltieri (dismissed):

Mehdi-Bendriss (dismissed):

MiaAltieri (approved):

Mehdi-Bendriss (approved):
Thanks Neha! It seems like 4 checks are failing to be ran on pull_request - can you check those before merging?

MiaAltieri (approved):

Gu1nness:
> Thanks Neha! It seems like 4 checks are failing to be ran on pull_request - can you check those before merging?

This is due to yesterday's GH Action outage.

PR #26, opened 2025-03-12
feat(DPE-6715) : LDAP Manager

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

This PR introduces the manager for LDAP integration.
It also introduces a new peer relation for the LDAP data. This gives a better separation of concerns of the databags.
This PR does not listen to any events, it just introduces the managers.

## 📑 Motivation and Context

This is part of the LDAP integration project.
There are two other PRs opened, #22 that adds config options for this LDAP integration and #23 that updates the configuration handling to write in config files instead of passing arguments as CLI parameters.

This has been tested using unit tests.

## ✅ Checklist


single_kernel_mongo/managers/backups.py
```diff
@@ -150,7 +150,8 @@ def on_relation_broken(self, relation: Relation) -> None:
         if self.state.is_scaling_down(relation.id):
             logger.info("Relation broken event occurring due to scale down.")
             return
-        self.remove_ca_cert_from_trust_store()
+        self.dependent.remove_ca_cert_from_trust_store(TrustStoreFiles.PBM.value)
```
MiaAltieri: nice

single_kernel_mongo/core/secrets.py
```diff
@@ -21,14 +21,16 @@
...
+def generate_secret_label(app_name: str, scope: Scope, relation: str | None = None) -> str:
     """Generate unique group_mappings for secrets within a relation context.
 
     Defined as a standalone function, as the choice on secret labels definition belongs to the
     Application Logic. To be kept separate from classes below, which are simply to provide a
     (smart) abstraction layer above Juju Secrets.
     """
     members = [app_name, scope.value]
+    if relation:
+        members = [relation, app_name, scope.value]
```
MiaAltieri, line 33: nit
```suggestion
    members = [app_name, scope.value]
    if relation:
        members.append(relation)
```

reason: easier to see the diff

single_kernel_mongo/exceptions.py
```diff
@@ -243,3 +243,11 @@ class ActionFailedError(Exception):
...
 class UnhealthyUpgradeError(Exception):
     """Raised when the upgrade is unhealthy during an post upgrade check."""
+
+
+class WaitingForLdapDataError(DeferrableError):
+    """Raised when the charm hasn't received data from ldap yet."""
+
+
+class LDAPSNotEnabledError(Exception):
+    """Raised when the charm hasn't received ldaps urls.."""
```
MiaAltieri:
```suggestion
    """Raised when the charm hasn't received ldaps urls."""
```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+        self.cert_relation_name = cert_relation_name
+        self.ldap_requirer = LdapRequirer(self.charm, self.relation_name)
+        self.certificate_transfer = CertificateTransferRequires(self.charm, self.cert_relation_name)
+
+    def assert_pass_hook_checks(self) -> None:
+        """Runs some hook checks before allowing the hook to run."""
+        if not self.state.db_initialised:
+            raise DeferrableFailedHookChecksError("DB is not initialised")
+        if self.state.is_role(MongoDBRoles.SHARD):
+            self.charm.status_manager.to_blocked("Cannot integrate LDAP with shard.")
```
MiaAltieri: nit: mind doing `set_and_share_status`, makes it easier in my status work when I create constants :)

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def is_valid_ldap_integration(self) -> bool:
+        """Returns True if the integration to ldap is valid."""
+        return (self.state.ldap_relation is not None) and not self.state.is_role(MongoDBRoles.SHARD)
+
+    def on_ldap_ready(self, relation: Relation) -> None:
+        """Runs when LDAP is ready."""
+        self.assert_pass_hook_checks()
+
+        self.charm.status_manager.to_maintenance("Configuring LDAP")
```
MiaAltieri: same comment about statuses as above

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+
+        ldap_data = self.ldap_requirer.consume_ldap_relation_data(relation=relation)
+        if not ldap_data:
+            logger.info("Waiting for LDAP data.")
+            raise WaitingForLdapDataError("Waiting for LDAP data.")
+
+        if not ldap_data.ldaps_urls:
+            logger.info("You must configure Glauth with LDAPS support")
+            raise LDAPSNotEnabledError("LDAPS not enabled.")
```
MiaAltieri: should a status be set here in either of these cases?

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def on_ldap_ready(self, relation: Relation) -> None:
+        """Runs when LDAP is ready."""
+        self.assert_pass_hook_checks()
+
+        self.charm.status_manager.to_maintenance("Configuring LDAP")
+
+        ldap_data = self.ldap_requirer.consume_ldap_relation_data(relation=relation)
+        if not ldap_data:
+            logger.info("Waiting for LDAP data.")
```
MiaAltieri, line 100: should we defer?

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+            raise WaitingForLdapDataError("Waiting for LDAP data.")
+
+        if not ldap_data.ldaps_urls:
+            logger.info("You must configure Glauth with LDAPS support")
```
MiaAltieri:
```suggestion
            logger.error("You must configure Glauth with LDAPS support")
```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+from single_kernel_mongo.state.charm_state import CharmState
+
+if TYPE_CHECKING:
+    from single_kernel_mongo.core.operator import OperatorProtocol
+
+logger = getLogger(__name__)
+
+
+class LDAPManager(Object, StatusProvider):
+    """Manage the relation between glauth-k8s and mongos router."""
```
MiaAltieri:
```suggestion
    """Manage the relation between glauth-k8s and replica set, config-server, or mongos-router."""
```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+        super().__init__(parent=dependent, key=relation_name)
+        self.dependent = dependent
+        self.charm = dependent.charm
+        self.workload = self.dependent.workload
+        self.state = state
+        self.substrate = substrate
+        self.relation_name = relation_name
+        self.cert_relation_name = cert_relation_name
+        self.ldap_requirer = LdapRequirer(self.charm, self.relation_name)
+        self.certificate_transfer = CertificateTransferRequires(self.charm, self.cert_relation_name)
```
MiaAltieri, line 73: I believe you are missing the event file to listen to:
```
        self.framework.observe(
            self.ldap_requirer.on.ldap_ready,
            self._on_ldap_ready,
        )
        self.framework.observe(
            self.ldap_requirer.on.ldap_unavailable,
            self._on_ldap_unavailable,
        )

```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+
+            self.charm.status_manager.to_active()
+
+    def on_ldap_unavailable(self) -> None:
+        """Runs when the LDAP integration is broken."""
+        self.state.ldap.clean_databag()
+
+        self.dependent.restart_charm_services()
+
+    def on_certificate_available(self, certificate: str, ca: str, chain: list[str]):
```
MiaAltieri: where are you listening to this event?

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+        if self.state.ldap.is_ready():
+            logger.info("Restarting mongodb server for LDAP integration")
+            self.dependent.restart_charm_services()
+
+    def on_certificate_removed(self):
+        """Runs when the certificate is removed."""
+        self.state.ldap.set_certificates(None, None, None)
+        self.dependent.remove_ca_cert_from_trust_store(TrustStoreFiles.LDAP)
+
+        self.dependent.restart_charm_services()
```
MiaAltieri: If we are in one of the cases of not having the cert and having the ldap interface (and vice versa) is the charm in the broken state? i.e. can't proceed with other events? Or is it only blocking clients?

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def on_ldap_unavailable(self) -> None:
+        """Runs when the LDAP integration is broken."""
+        self.state.ldap.clean_databag()
+
+        self.dependent.restart_charm_services()
+
+    def on_certificate_available(self, certificate: str, ca: str, chain: list[str]):
+        """Runs when we receive the LDAP certificates."""
+        self.assert_pass_hook_checks()
```
MiaAltieri, line 145: don't shards need these certs? Since clients eventually communicate with shards?

MiaAltieri (commented):
Great work! Is it possible at this stage to test manually within the charm doing a few sanity check ops? If so that would be great! If not NW

single_kernel_mongo/core/secrets.py
Gu1nness: This is not the same behavior as it would put the relation at the end, which differs from what is done everywhere else: `<relation>.<app>.<scope>`.
I agree that this is not the best diff but I'm trying to stick with the standards :)

single_kernel_mongo/managers/ldap.py
Gu1nness: This will be set on the event handler I think.
However I'm open to set it here if that makes your work later easier.

single_kernel_mongo/managers/ldap.py
Gu1nness: Defer will be done in the event handler as usual.
The manager has no information of events (and I want to keep it that way) so the event handler will catch this error and defer.
If you have a look at the exception definition, you can already see that it does inherits from `DeferrableError` :)

single_kernel_mongo/managers/ldap.py
Gu1nness: Cf PR description:
>  This PR does not listen to any events, it just introduces the managers.

single_kernel_mongo/managers/ldap.py
Gu1nness: Cf PR description:
>  This PR does not listen to any events, it just introduces the managers.

single_kernel_mongo/managers/ldap.py
Gu1nness: It will not be in a broken state, it can process other events.
However identity users will not be able to log in as ldap won't be configured on the services.

single_kernel_mongo/managers/ldap.py
Gu1nness: This will be done in another PR if needed.

Keep in mind that this is for identity users, not client applications.
It is not recommended to log in directly on a shard, but use mongos instead.
The decision was taken not to integrate with shards, and sharing the necessary information will be done if necessary by the config-server.

As I said, this PR introduces the manager, it does not even update the configuration of the services.
I'm trying to keep small PR sizes and minimal changes to reduce the risk of breaking things.
On top of that, since this PR does not listen to events, it is for now completely dead code which make continuous development easier, not breaking the charms.

Gu1nness:
> Great work! Is it possible at this stage to test manually within the charm doing a few sanity check ops? If so that would be great! If not NW

The best we can do for now is unit testing as I did + ensuring that we don't break anything with the remote integration workflow. However remote integration will still need to be updated because we need to update the metadata files to add the new relations.

single_kernel_mongo/managers/ldap.py
MiaAltieri: oopsies duh 🤦‍♀️ - thanks for your patience!

single_kernel_mongo/managers/ldap.py
MiaAltieri: oopsies duh 🤦‍♀️ - thanks for your patience!

single_kernel_mongo/managers/ldap.py
MiaAltieri: Thanks for the explanation! Gotcha!

single_kernel_mongo/core/secrets.py
MiaAltieri: ah - thanks for the explaination. No worries then!

single_kernel_mongo/managers/ldap.py
MiaAltieri: setting it here will make my work later easier!

single_kernel_mongo/managers/ldap.py
MiaAltieri: Ah - yes! thank you Neha. I appreciate your patience!

MiaAltieri (dismissed):

single_kernel_mongo/config/literals.py
```diff
@@ -122,4 +122,11 @@ class VmUser(WorkloadUser[int]):
     "vm.max_map_count": "262144",
 }
 
-TRUST_STORE_CERTIFICATE_PATH = Path("/usr/local/share/ca-certificates/pbm.crt")
+TRUST_STORE_PATH = Path("/usr/local/share/ca-certificates")
```
Mehdi-Bendriss, line 125: Can we define a custom `${SNAP_DATA}` relative localized trust store that is not system-wide? (with restricted permissions)

single_kernel_mongo/managers/backups.py
```diff
@@ -377,7 +378,10 @@ def set_config_options(self, credentials: dict[str, str]) -> None:
             credentials: A dictionary provided by backup event handler.
         """
         # Add certificate to trust store
-        self.save_ca_cert_to_trust_store(credentials)
+        cert_chain_list = credentials.get("tls-ca-chain", None)
+        if cert_chain_list:
```
Mehdi-Bendriss:
```suggestion
        if cert_chain_list := credentials.get("tls-ca-chain", None):
```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+        return (self.state.ldap_relation is not None) and not self.state.is_role(MongoDBRoles.SHARD)
+
+    def on_ldap_ready(self, relation: Relation) -> None:
+        """Runs when LDAP is ready."""
+        self.assert_pass_hook_checks()
+
+        self.charm.status_manager.set_and_share_status(MaintenanceStatus("Configuring LDAP"))
+
+        ldap_data = self.ldap_requirer.consume_ldap_relation_data(relation=relation)
+        if not ldap_data:
```
Mehdi-Bendriss, line 99: nit: walrus operator

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+        if not ldap_data.ldaps_urls:
+            logger.error("You must configure Glauth with LDAPS support")
+            raise LDAPSNotEnabledError("LDAPS not enabled.")
+
+        if self.charm.unit.is_leader():
+            self.state.ldap.set_from(ldap_data)
+
+        self.restart_when_ready()
+
+    def restart_when_ready(self):
```
Mehdi-Bendriss:
```suggestion
    def restart_when_ready(self) -> None:
```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+            self.dependent.restart_charm_services()
+            self.charm.status_manager.set_and_share_status(ActiveStatus())
+
+    def on_ldap_unavailable(self) -> None:
+        """Runs when the LDAP integration is broken."""
+        self.state.ldap.clean_databag()
+
+        self.dependent.restart_charm_services()
+
+    def on_certificate_available(self, certificate: str, ca: str, chain: list[str]):
```
Mehdi-Bendriss:
```suggestion
    def on_certificate_available(self, certificate: str, ca: str, chain: list[str]) -> None:
```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+        """Runs when we receive the LDAP certificates."""
+        self.assert_pass_hook_checks()
+        self.state.ldap.set_certificates(certificate, ca, chain)
+
+        full_chain = "\n".join(chain)
+        self.dependent.save_ca_cert_to_trust_store(TrustStoreFiles.LDAP, full_chain)
+
+        self.restart_when_ready()
+
+    def on_certificate_removed(self):
```
Mehdi-Bendriss:
```suggestion
    def on_certificate_removed(self) -> None:
```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+            return
+        if not self.state.ldap_cert_relation:
+            logger.info(
+                f"Not restarting, missing the `{ExternalRequirerRelations.LDAP_CERT}` relation"
+            )
+            return
+
+        if self.state.ldap.is_ready():
+            logger.info("Restarting mongodb server for LDAP integration")
+            self.dependent.restart_charm_services()
```
Mehdi-Bendriss: I like that this handled by the `dependent`, for when we introduce rolling restarts

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+            self.charm.status_manager.set_and_share_status(ActiveStatus())
```
Mehdi-Bendriss: Setting an Active Status from the ldap module seems too radical for me, can we be more targeting?

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def get_status(self) -> StatusBase | None:
+        """Generates the status of a unit based on its status reported by mongod."""
+        if self.state.ldap_relation is None and self.state.ldap_cert_relation is None:
+            return None
+
+        if self.state.is_role(MongoDBRoles.SHARD):
+            return BlockedStatus("Cannot integrate LDAP with shard.")
+
+        if self.state.ldap_relation is not None and self.state.ldap_cert_relation is None:
```
Mehdi-Bendriss: The first part of the condition is implied
```suggestion
        if self.state.ldap_cert_relation is None:
```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+            logger.info(
+                "Integrate the certificate interface between glauth and charm using"
+                f"`juju integrate {self.state.ldap_relation.app.name}:send-ca-cert {self.charm.app.name}:{ExternalRequirerRelations.LDAP_CERT}`"
+            )
+            return BlockedStatus("TLS is mandatory for LDAP transport.")
+        if self.state.ldap_relation is None and self.state.ldap_cert_relation is not None:
```
Mehdi-Bendriss: Similar comment:
```suggestion
        if self.state.ldap_relation is None:
```

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+        ldap_certificate_integration_status = self.state.ldap.ldap_certs_ready()
+
+        match (ldap_relation_status, ldap_certificate_integration_status):
+            case False, False:
+                return BlockedStatus("Waiting for both LDAP data and Glauth certificates.")
+            case True, False:
+                return BlockedStatus("Waiting for Glauth certificates.")
+            case False, True:
+                logger.info("Waiting for LDAP data.")
+                return BlockedStatus("Missing LDAP data from Glauth.")
```
Mehdi-Bendriss: These should be WaitingStatus(es) I think - as it is just a matter of time until those are resolved, and will not require user intervention

single_kernel_mongo/state/ldap_state.py
```diff
@@ -0,0 +1,196 @@
...
+    def chain(self) -> str | None:
+        """The chain."""
+        return self.secrets.get_for_key(Scope.UNIT, LdapStateKeys.CHAIN.value)
+
+    @chain.setter
+    def chain(self, value: str | None) -> None:
+        if not value:
+            self.secrets.remove(Scope.UNIT, LdapStateKeys.CHAIN.value)
+            return
+        self.secrets.set(LdapStateKeys.CHAIN.value, value, Scope.UNIT)
```
Mehdi-Bendriss: In another PR - We'll need to readjust the secrets API, some methods have the Scope as a first arg, where others have it as the last arg

single_kernel_mongo/core/operator.py
```diff
@@ -248,3 +253,27 @@ def set_permissions(self) -> None:
...
+
+        # Update ca certificates.
+        self.workload.exec("update-ca-certificates")
+
+    def remove_ca_cert_from_trust_store(self, file: str):
+        """Removes the certificate from the trust store."""
+        # Remove the file
+        self.workload.delete(TRUST_STORE_PATH / file)
+        # Update CA certificates to remove the certificate from the trust store
+        self.workload.exec("update-ca-certificates")
```
Mehdi-Bendriss, line 289: related to the above comment on localized trust store

Mehdi-Bendriss (commented):
Thanks Neha! Great work! I have some comments to clarify

single_kernel_mongo/managers/ldap.py
```diff
@@ -0,0 +1,175 @@
...
+            raise DeferrableFailedHookChecksError(
+                "Adding LDAP is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
+            )
+        return
+
+    def is_valid_ldap_integration(self) -> bool:
+        """Returns True if the integration to ldap is valid."""
+        return (self.state.ldap_relation is not None) and not self.state.is_role(MongoDBRoles.SHARD)
+
+    def on_ldap_ready(self, relation: Relation) -> None:
```
Mehdi-Bendriss: Could you rename all of these methods to a more "manager" friendly name, that indicates the actual business logic it is handling. The `on_*` naming should be reserved to the events

single_kernel_mongo/config/literals.py
Gu1nness: I'll investigate.
I have never done such things so I'll read a bit about that and see if that is doable or not.

single_kernel_mongo/managers/ldap.py
Gu1nness: It's a bit tricky, because we are just restarting the mongod service so it might not be available yet.
I can add a more generic status but we know that it tends to be capricious.
TBH, I used an ActiveStatus mainly not to make any decisions here as statuses would be reworked right after by advanced statuses.

single_kernel_mongo/state/ldap_state.py
Gu1nness: Definitely, I'm always having the exact same remark in the back of my head.

single_kernel_mongo/managers/ldap.py
Gu1nness: Sure !

single_kernel_mongo/config/literals.py
Gu1nness: Done in #29

single_kernel_mongo/managers/ldap.py
```diff
@@ -113,9 +128,7 @@ def on_certificate_available(self, certificate: str, ca: str, chain: list[str]):
         full_chain = "\n".join(chain)
         self.dependent.save_ca_cert_to_trust_store(TrustStoreFiles.LDAP, full_chain)
 
-        if self.state.ldap.is_ready():
-            logger.info("Restarting mongodb server for LDAP integration")
-            self.dependent.restart_charm_services()
+        self.restart_when_ready()
```
MiaAltieri: Are there any dangers of not deferring when not ready?

single_kernel_mongo/managers/ldap.py
```diff
@@ -79,25 +82,37 @@ def on_ldap_ready(self, relation: Relation) -> None:
...
 
         if not ldap_data.ldaps_urls:
-            logger.info("You must configure Glauth with LDAPS support")
+            logger.error("You must configure Glauth with LDAPS support")
             raise LDAPSNotEnabledError("LDAPS not enabled.")
 
         if self.charm.unit.is_leader():
             self.state.ldap.set_from(ldap_data)
 
+        self.restart_when_ready()
```
MiaAltieri: Are there any dangers of not deferring when not ready?  (I assume no since the relation data would update and trigger the event calling this function)

MiaAltieri:
Looks great! Small comments only <3

single_kernel_mongo/managers/ldap.py
Gu1nness: We specifically don't want to defer here:
1) we are waiting on two relations
2) restart when ready first gets the component status:
 * if there is no status, there is no integration so we don't restart
 * if there is an active status, then everything is integrated and we have checked that the connection works
 * Else, there is something blocking and we go in blocked state because this will not auto resolve.

single_kernel_mongo/managers/ldap.py
Gu1nness: We specifically don't want to defer here:
1) we are waiting on two relations
2) restart when ready first gets the component status:
 * if there is no status, there is no integration so we don't restart
 * if there is an active status, then everything is integrated and we have checked that the connection works
 * Else, there is something blocking and we go in blocked state because this will not auto resolve.

Mehdi-Bendriss (dismissed):
Thanks Neha!

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):
Thank you Neha!

PR #27, opened 2025-03-12
fix: Init replica set with one host only.

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

The replicaset should always be initialised only with only one node.
The other nodes should be added by the relation changed events.

## 📑 Motivation and Context

This issue was reported in [this issue](https://github.com/canonical/mongodb-k8s-operator/issues/397) and seen as well in [CI runs](https://github.com/canonical/mongodb-k8s-operator/actions/runs/13784633384/job/38567069294)

This has been tested manually for now.
I'll run the CI tests with this PR open.

## ✅ Checklist


Gu1nness:
/test/mongodb/k8s

-> https://github.com/canonical/mongodb-k8s-operator/pull/399

MiaAltieri (commented):
I need to look at the bug linked in the description more carefully. But I don't think this change is needed. For the following reasons:

1. Init replica set is performed before we add the other replicas (i.e. `is_db_initialised` is a blocking condition.) so when we init there is always one member and so it will by default be the primary 
2. [the calling node for rs.intitae by default becomes the primary](https://stackoverflow.com/questions/25585452/how-to-initialize-mongodb-replication-set-without-calling-rs-initiate) 


When I looked at the issue that Gruyaume brought up, it was on 6/stable. It would be worth asking if the issue occurs on 6/edge since I know some **major** changes have occurred since then! I can ask in that thread

MiaAltieri:
Also a quick glance showed a different error in CI than from the one Gruyaume reported?

Gu1nness:
> Also a quick glance showed a different error in CI than from the one Gruyaume reported?

Indeed and my bad.
However the bug I'm fixing is a real bug that happens so I'd rather ensure that it doesn't happen.
I have the feeling that when we pass the nodes in the config, they are not ordered as expected and that's what is causing it and it doesn't happen anymore with this fix.

single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -120,15 +120,22 @@ def is_ready(self) -> bool:
...
+    def init_replset(self, leader_host: str) -> None:
         """Create replica set config the first time.
 
         Raises:
             ConfigurationError, ConfigurationError, OperationFailure
         """
         config = {
             "_id": self.config.replset,
-            "members": [{"_id": i, "host": h} for i, h in enumerate(self.config.hosts)],
+            "members": [
```
MiaAltieri: Context from earlier Charmed MongoDB design decision:

We want to `init_replica_set` with only the juju leader unit, and then slowly add each unit with `database_relation_changed` events. 

To ensure this (while solving your issue) I believe this is the correct approach is to update the callers `mongod_config` so that it only has one host and not all internal hosts. You can update this [where we try to init the replica set](https://github.com/canonical/mongo-single-kernel-library/blob/0bdb15f119a2cee2f7fa462087d9fae5922c9ecf/single_kernel_mongo/managers/mongo.py#L108). 

What I believe is happening is that there is a race condition where:
- start defers
- we get peer relation events
- start gets retriggered with now a list of internal hosts
- your error occurs. 

I hope that helps

MiaAltieri (changes requested):

single_kernel_mongo/utils/mongo_connection.py
Gu1nness: That's probably it.
However I'm wondering, if we got peer relation events because start deferred, after initialising the replica set, when will we add the other hosts? Because the relation events have already run right?

Gu1nness: I still think that the solution I provide is quite more stable and does not care of how the start event was deferred or not and the priority will get set to 1 very soon after that because we reset it to 1 on status updates.

MiaAltieri: Neha! I am so sorry I did not see these comments! I am getting to them now and adding as much insight as I can so that they can guide you in your decision making when you return.

>  if we got peer relation events because start deferred, after initialising the replica set, when will we add the other hosts? Because the relation events have already run right?

I might not be following - but I think the scenario you proposed isn't our current situation because if we defer **after initialising the replica set** that means that the initialisation was successful and that we won't have to worry fixing the replica-set initialisation? I believe this issue only occurs when deferring **before initialising the replica set** as the primary is decided at init. 

Also , relation-events shouldn't run to completion until `db_initialised` so we don't have to worry about those events running

But maybe I followed that line of thought incorrectly ? If I am please describe further! 

=====


> I still think that the solution I provide is quite more stable and does not care of how the start event was deferred or not and the priority will get set to 1 very soon after that because we reset it to 1 on status updates.

My concern is that if there is some instability after deployment and update-status is set to run at a longer interval i.e. 10minutes, and the current primary goes down, or has lag, etc, that the entire replica set will experience poor performance since it is waiting to self heal.

Do you have an example of how my solution might be equally unstable? 

If my solution is comparably unstable and you decide to proceed with your solution I would heavily suggest adding a comment + a log explaining the situation since this does seem like a workaround to the root cause.

MiaAltieri: Please don't let that comment discourage you - you are doing great and I appreciate you working hard to clear up these bugs - I wish everyone had your attitude!

Gu1nness: > Also , relation-events shouldn't run to completion until db_initialised so we don't have to worry about those events running

Initial relation events cannot be blocked on db_initialised since it's stored in the relation databag.
Moreover, we get the list of ips on the members of the peer relation, so we'll get this list anyway.
I'll provide a better fix :)

Gu1nness:
/test/mongodb/k8s

Gu1nness:
/test/mongodb/k8s

single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -120,16 +120,13 @@ def is_ready(self) -> bool:
...
         """Create replica set config the first time.
 
         Raises:
             ConfigurationError, ConfigurationError, OperationFailure
         """
-        config = {
-            "_id": self.config.replset,
-            "members": [{"_id": i, "host": h} for i, h in enumerate(self.config.hosts)],
-        }
+        config = {"_id": self.config.replset, "members": [{"_id": 0, "host": host}]}
```
Mehdi-Bendriss, line 129: Shouldn't the `id` be bumped for subsequent added replicas?

Mehdi-Bendriss (commented):
Thanks Neha, I have a question

single_kernel_mongo/utils/mongo_connection.py
Gu1nness: This is called only once so when we add the other nodes, it will be done using `replSetReconfig`, and the `_id` are computed to always be `max(current_ids) + 1`
So this shouldn't be an issue

Mehdi-Bendriss (approved):

PR #29, opened 2025-03-13
feat(DPE-6716): LDAP Events

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

LDAP events and refactor a bit the management of restarts
This PR introduces the `LDAPEventHandler` which handles the events of the three following relations:
 * ldap
 * ldap-certificate-transfer
 * ldap-peers

It also introduces a custom ops event that is used to provide a single source of truth for restarts in the LDAP integration.
There is a bit of rework of the restarts, and an improvement of the get_status that now also checks for the ldap connection by doing a bind operation.

## 📑 Motivation and Context

This is the implementation of [DPE-6716](https://warthogs.atlassian.net/browse/DPE-6716) for the ldap integration.
It adds the support for the events.
The next steps will be to update the config so that the services are configured to integrate with LDAP.

This has been manually tested for now.
I'm currently adding unit tests.

## ✅ Checklist

[DPE-6716]: https://warthogs.atlassian.net/browse/DPE-6716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


single_kernel_mongo/events/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+            self.charm.on["ldap-peers"].relation_changed, self._on_restart_if_ready
+        )
+
+    def _on_ldap_ready(self, event: LdapReadyEvent):
+        """Handles the ops event that indicates that ldap relation is ready."""
+        action = "ldap-ready"
+        try:
+            self.manager.on_ldap_ready(event.relation)
+        except WaitingForLdapDataError as err:
+            self.charm.status_manager.set_and_share_status(BlockedStatus("Waiting for LDAP data."))
```
MiaAltieri:
```suggestion
            self.charm.status_manager.set_and_share_status(WaitingStatus("Waiting for LDAP data."))
```

single_kernel_mongo/events/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+            self.charm.status_manager.set_and_share_status(
+                BlockedStatus("LDAPS not enabled on LDAP application.")
+            )
+        except NonDeferrableFailedHookChecksError as err:
+            logger.error(f"{err}")
+            self.charm.status_manager.set_and_share_status(BlockedStatus(err.args[0]))
+
+    def _on_ldap_unavailable(self, event: LdapUnavailableEvent):
+        """Handles the ops event that indicates that ldap relation is now unavailable."""
+        self.manager.on_ldap_unavailable()
```
MiaAltieri: Double checking, is there nothing to handle?

single_kernel_mongo/events/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _on_certificate_available(self, event: CertificateAvailableEvent):
+        """Handles the ops event that indicates that ldap-certificates relation is ready."""
+        try:
+            self.manager.on_certificate_available(event.certificate, event.ca, event.chain)
+        except DeferrableFailedHookChecksError as err:
+            defer_event_with_info_log(logger, event, "ldap-cert-ready", f"{err}")
+        except NonDeferrableFailedHookChecksError as err:
+            logger.error(f"{err}")
+            self.charm.status_manager.set_and_share_status(BlockedStatus(err.args[0]))
```
MiaAltieri: What scenarios does this error occur?

I think in a scenario like this where we cannot retry the event we should: fail noisily and raise

single_kernel_mongo/events/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+    def _on_restart_if_ready(self, event: RestartIfReadyEvent):
+        """Custom ops revent to trigger restart of leader with a single source of truth."""
+        action = "restart-ldap-if-ready"
+        try:
+            self.manager.restart_when_ready()
+        except DeferrableFailedHookChecksError as err:
+            defer_event_with_info_log(logger, event, action, f"{err}")
+        except NonDeferrableFailedHookChecksError as err:
+            logger.error(f"{err}")
+            self.charm.status_manager.set_and_share_status(BlockedStatus(err.args[0]))
```
MiaAltieri, line 118: same here

single_kernel_mongo/managers/ldap.py
```diff
@@ -90,53 +91,61 @@ def on_ldap_ready(self, relation: Relation) -> None:
...
+                self.charm.status_manager.set_and_share_status(ActiveStatus())
+            case status:
+                self.charm.status_manager.set_and_share_status(status)
 
     def on_ldap_unavailable(self) -> None:
         """Runs when the LDAP integration is broken."""
         self.state.ldap.clean_databag()
 
         self.dependent.restart_charm_services()
+        self.charm.status_manager.set_and_share_status(self.get_status() or ActiveStatus())
```
MiaAltieri, line 125: Is the relation already unavailable when the ldap_unavailable event comes through? Because I wonder how `get_status` would behave in this context

single_kernel_mongo/utils/helpers.py
```diff
@@ -73,3 +80,29 @@ def wrapper(self, *args, **kwargs):
 
 mongodb_only = partial(charm_kind_only, charm_kind=CharmKind.MONGOD)
 mongos_only = partial(charm_kind_only, charm_kind=CharmKind.MONGOS)
+
+
+def get_ldap_connection_status(state: LdapState) -> StatusBase:
```
MiaAltieri: should this not be a part of the `get_status` function?

single_kernel_mongo/events/ldap.py
Gu1nness: We clean the databag and restart, so for now it seems that no?
Maybe we can go through this tomorrow during 1-1 if you want.

single_kernel_mongo/events/ldap.py
Gu1nness: This is the case where the role is invalid.
I think it's sufficient to go to blocked with the correct error message.

I don't like raising to the top because it breaks the status display and it's painful to recover.
I'd rather set blocked status with correct information so that in the end it will be easier for the operator to recover: they have integrated the wrong relation, let's remove the relation and everything will be good again, which would be way harder to fix with a raise that would put the charm in error state and make the deletion of the integration painful.

single_kernel_mongo/events/ldap.py
Gu1nness: This is the case where the role is invalid.
I think it's sufficient to go to blocked with the correct error message.

I don't like raising to the top because it breaks the status display and it's painful to recover.
I'd rather set blocked status with correct information so that in the end it will be easier for the operator to recover: they have integrated the wrong relation, let's remove the relation and everything will be good again, which would be way harder to fix with a raise that would put the charm in error state and make the deletion of the integration painful.

If we refer to the advanced status spec, this would be a perfect candidate for a `Critical Status`.

single_kernel_mongo/managers/ldap.py
Gu1nness: It's emitted in the `relation_broken` event.
`get_status` handles both events so it should be fine and if I'm not wrong, it should update the status so that we know there is only one of the relations integrated. I'll give it a try !

single_kernel_mongo/utils/helpers.py
Gu1nness: This is called in the last case of the `get_status`, but exported in the helpers to reduce file size and function complexity, on top of providing an easily reusable function if needed.

single_kernel_mongo/managers/ldap.py
Gu1nness: it behaves as expected

single_kernel_mongo/events/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+            self.manager.certificate_transfer.on.certificate_available,
+            self._on_certificate_available,
+        )
+        self.framework.observe(
+            self.manager.certificate_transfer.on.certificate_removed, self._on_certificate_removed
+        )
+        self.framework.observe(self.restart_if_ready_event, self._on_restart_if_ready)
+
+        self.framework.observe(
+            self.charm.on["ldap-peers"].relation_changed, self._on_restart_if_ready
```
Mehdi-Bendriss: Could you define `ldap-peers` in a constant?

single_kernel_mongo/events/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    def _on_ldap_ready(self, event: LdapReadyEvent):
```
Mehdi-Bendriss: And in all subsequent methods
```suggestion
    def _on_ldap_ready(self, event: LdapReadyEvent) -> None:
```

single_kernel_mongo/events/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+        """Handles the ops event that indicates that ldap relation is ready."""
+        action = "ldap-ready"
+        try:
+            self.manager.on_ldap_ready(event.relation)
+        except WaitingForLdapDataError as err:
+            self.charm.status_manager.set_and_share_status(WaitingStatus("Waiting for LDAP data."))
+            defer_event_with_info_log(logger, event, action, f"{err}")
+
```
Mehdi-Bendriss:
```suggestion
```

single_kernel_mongo/events/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+            self.manager.on_certificate_available(event.certificate, event.ca, event.chain)
+        except DeferrableFailedHookChecksError as err:
+            defer_event_with_info_log(logger, event, "ldap-cert-ready", f"{err}")
+        except NonDeferrableFailedHookChecksError as err:
+            logger.error(f"{err}")
+            self.charm.status_manager.set_and_share_status(BlockedStatus(err.args[0]))
+
+    def _on_certificate_removed(self, event: CertificateRemovedEvent):
+        """Handles the ops event that indicates that ldap-certificates relation is unavailable."""
+        self.manager.on_certificate_removed()
```
Mehdi-Bendriss: Similar comment on manager methods naming, it is unclear from here what exactly does the manager do in this hook

single_kernel_mongo/events/ldap.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _on_restart_if_ready(self, event: RestartIfReadyEvent):
+        """Custom ops revent to trigger restart of leader with a single source of truth."""
+        action = "restart-ldap-if-ready"
+        try:
+            self.manager.restart_when_ready()
```
Mehdi-Bendriss, line 113: There seems to be a bit of a discrepancy between the manager restarts and this one, I see 3 points of entry for a restart:
- restart_charm_services 
- restart_when_ready
- restart_if_ready_event.emit()

single_kernel_mongo/utils/helpers.py
```diff
@@ -73,3 +80,29 @@ def wrapper(self, *args, **kwargs):
...
+    if not state.ldaps_urls:
+        return BlockedStatus("Missing LDAPS URLs.")
+
+    try:
+        for ldap_uri in state.ldaps_urls:
+            tls = LDAPTls(validate=ssl.CERT_REQUIRED, version=ssl.PROTOCOL_TLSv1_2)
+            server = LDAPServer(host=ldap_uri, use_ssl=True, tls=tls)
+            conn = LDAPConnection(server, user=bind_dn, password=bind_password)
+            conn.bind()  # We consider sufficient to be able to bind.
+            conn.unbind()
```
Mehdi-Bendriss: Can you explain this?

single_kernel_mongo/utils/helpers.py
Mehdi-Bendriss: I think this is only used in the ldap module from within the `get_status` method, it would make sense to keep this method as part of the ldap manager instead of the helper

Mehdi-Bendriss (commented):
Thanks Neha! great work! I have some comments

single_kernel_mongo/events/ldap.py
MiaAltieri: Thanks for the explaination. For readability, can we set this to .set_and_share_status(BlockedStatus("Invalid role <...>"))

single_kernel_mongo/events/ldap.py
MiaAltieri: same comment as above for readability. Thanks for the explaination! 

As for the `Critical Status` - these statuses are defined as:
```
The very first action a user should take
The user might need to take this action immediately 
```

`The user might need to take this action immediately ` should be updated to be more clear, in that it should mean that if the user doesn't take this action immediately they can break their cluster or go into an unrecoverable state. Which I don't believe qualifies this status. 

Also I believe that in the case that other actions are needed i.e. `enable TLS on shard` - that it is **not** necessary for the ldap action to be the first action for the user to take. Or am I wrong there? Would the cluster go into a broken state?

single_kernel_mongo/events/ldap.py
Gu1nness: Oh yeah sure, completely forgot to update that

single_kernel_mongo/events/ldap.py
Gu1nness: All restarts go trough the `restart_when_ready` which  first runs some checks and then calls the `restart_charm_services`.

`restart_when_ready` is called by the relation changed events (for the non leader units) or by the custom ops event (for the leader unit).

single_kernel_mongo/utils/helpers.py
Gu1nness: We ensure that we have successful connectivity with LDAP server using TLS.
It's not mandatory because mongod will refuse to start if configured with ldap but cannot connect to ldap but this can prevent unsuccessful restarts.

single_kernel_mongo/utils/helpers.py
Gu1nness: I'll move it there then.

single_kernel_mongo/events/ldap.py
Gu1nness: Sure but this will not be sturdy if we have more cases that can raise this.
Let me rephrase this case (and following case) in a better manner.

single_kernel_mongo/events/ldap.py
Gu1nness: No it does not go in an error state however this is a case that we don't support and the user must know that and should remove this integration

MiaAltieri: > No it does not go in an error state however this is a case that we don't support and the user must know that and should remove this integration

Thanks for the explanation! Based on this I believe it is sufficient to be kept as a normal status - since the user will see that there is a action to perform in the status message (in case another status is shown over ldap)

single_kernel_mongo/managers/ldap.py
```diff
@@ -191,4 +194,36 @@ def get_status(self) -> StatusBase | None:
...
+                tls = LDAPTls(
+                    validate=ssl.CERT_REQUIRED,
+                    version=ssl.PROTOCOL_TLSv1_2,
+                    ca_certs_file=f"{self.state.paths.ldap_certificates_file}",
+                )
+                server = LDAPServer(host=ldap_uri, use_ssl=True, tls=tls)
+                conn = LDAPConnection(server, user=bind_dn, password=bind_password)
+                conn.bind()  # We consider sufficient to be able to bind.
+                conn.unbind()
+        except Exception as err:
```
MiaAltieri: perhaps we can check for a specific error?

single_kernel_mongo/managers/ldap.py
```diff
@@ -191,4 +194,36 @@ def get_status(self) -> StatusBase | None:
...
+        try:
+            for ldap_uri in self.state.ldap.ldaps_urls:
+                tls = LDAPTls(
+                    validate=ssl.CERT_REQUIRED,
+                    version=ssl.PROTOCOL_TLSv1_2,
+                    ca_certs_file=f"{self.state.paths.ldap_certificates_file}",
+                )
+                server = LDAPServer(host=ldap_uri, use_ssl=True, tls=tls)
+                conn = LDAPConnection(server, user=bind_dn, password=bind_password)
+                conn.bind()  # We consider sufficient to be able to bind.
```
MiaAltieri: `# to verify each ldap URI can connect we checks its ability to bind`

I would also maybe add in your domain specific knowledge (i.e. what is binding)

single_kernel_mongo/managers/ldap.py
```diff
@@ -191,4 +194,36 @@ def get_status(self) -> StatusBase | None:
                 logger.info("Waiting for LDAP data.")
                 return BlockedStatus("Missing LDAP data from Glauth.")
             case _:
-                return get_ldap_connection_status(self.state.ldap, self.state.paths)
+                return self.get_ldap_connection_status()
+
+    def get_ldap_connection_status(self) -> StatusBase:
+        """Checks if the LDAP connection is working or not.
```
MiaAltieri, line 202: """Returns a status for LDAP connections

This involves checking that each LDAP uri can connect to the server

single_kernel_mongo/events/ldap.py
MiaAltieri: Ah - okay! I see - it is probably a nicer developer interface to have it this way (especially if there are many cases raising)

What if each error had an error code associated with it, and each code was associated with a status (in the future status config file)? Or is that over complex? If so maybe you can add a comment here along the lines of:

```
# status is set as a predefined value in the raised exception and not simply an error raised from an underlying API
```

single_kernel_mongo/managers/ldap.py
Gu1nness: Sure, I'll improve that.

single_kernel_mongo/managers/ldap.py
Gu1nness: Sure, let me add that.

Mehdi-Bendriss (approved):

MiaAltieri (approved):
Great work!

PR #33, opened 2025-03-20
feat(DPE-6792): Share config with mongos

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

Synchronization of MongoDB and Mongos for the config option.

## 📑 Motivation and Context

Instead of requiring the operator to config the two services, the config-server shares the ldapUserToDNMapping with mongos through the cluster relation.

Manually tested + unit testing.

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/core/operator.py
+++ b/single_kernel_mongo/core/operator.py
@@ -280,9 +280,10 @@ class OperatorProtocol(ABC, Object):
 
     def remove_ca_cert_from_trust_store(self, file: TrustStoreFiles):
         """Removes the certificate from the trust store."""
-        # Remove the file
-        self.workload.delete(TRUST_STORE_PATH / file.value)
-        # Update CA certificates to remove the certificate from the trust store
-        self.workload.exec("update-ca-certificates")
-        # Restart the service
-        self.restart_charm_services(force=True)
+        if self.workload.exists(TRUST_STORE_PATH / file.value):
+            # Remove the file
+            self.workload.delete(TRUST_STORE_PATH / file.value)
+            # Update CA certificates to remove the certificate from the trust store
+            self.workload.exec("update-ca-certificates")
+            # Restart the service
+            self.restart_charm_services(force=True)
--- a/single_kernel_mongo/core/operator.py
+++ b/single_kernel_mongo/core/operator.py
@@ -280,10 +280,12 @@ class OperatorProtocol(ABC, Object):
 
     def remove_ca_cert_from_trust_store(self, file: TrustStoreFiles):
         """Removes the certificate from the trust store."""
-        if self.workload.exists(TRUST_STORE_PATH / file.value):
-            # Remove the file
-            self.workload.delete(TRUST_STORE_PATH / file.value)
-            # Update CA certificates to remove the certificate from the trust store
-            self.workload.exec("update-ca-certificates")
-            # Restart the service
-            self.restart_charm_services(force=True)
+        if not self.workload.exists(TRUST_STORE_PATH / file.value):
+            return
+
+        # Remove the file
+        self.workload.delete(TRUST_STORE_PATH / file.value)
+        # Update CA certificates to remove the certificate from the trust store
+        self.workload.exec("update-ca-certificates")
+        # Restart the service
+        self.restart_charm_services(force=True)
--- a/single_kernel_mongo/managers/cluster.py
+++ b/single_kernel_mongo/managers/cluster.py
@@ -161,7 +161,11 @@ class ClusterProvider(Object):
 
     def update_ldap_user_to_dn_mapping(self) -> None:
         """Updates the ldap user to dn mapping value in the databag."""
-        self.assert_pass_hook_checks()
+        try:
+            self.assert_pass_hook_checks()
+        except (DeferrableFailedHookChecksError, NonDeferrableFailedHookChecksError):
+            logger.info("Not updating ldap user to dn mapping now, not ready.")
+            return
 
         if not self.charm.unit.is_leader():
             return
--- a/single_kernel_mongo/managers/cluster.py
+++ b/single_kernel_mongo/managers/cluster.py
@@ -102,6 +102,9 @@ class ClusterProvider(Object):
         if int_tls_ca := self.state.tls.get_secret(label_name=SECRET_CA_LABEL, internal=True):
             relation_data[ClusterStateKeys.INT_CA_SECRET.value] = int_tls_ca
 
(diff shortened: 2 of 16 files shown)


single_kernel_mongo/managers/cluster.py
```diff
@@ -102,6 +102,9 @@ def share_secret_to_mongos(self, relation: Relation) -> None:
         if int_tls_ca := self.state.tls.get_secret(label_name=SECRET_CA_LABEL, internal=True):
             relation_data[ClusterStateKeys.INT_CA_SECRET.value] = int_tls_ca
```
MiaAltieri, line 104: please add a small comment with your expert ldap+mongo knowledge on why config-server must share this data to mongos

single_kernel_mongo/managers/cluster.py
```diff
@@ -154,6 +157,17 @@ def update_config_server_db(self) -> None:
                 },
             )
 
+    def update_ldap_user_to_dn_mapping(self) -> None:
+        """Updates the ldap user to dn mapping value in the databag."""
+        self.assert_pass_hook_checks()
+        for relation in self.state.cluster_relations:
+            self.data_interface.update_relation_data(
```
MiaAltieri, line 211: assert_pass_hook_checks doesn't check for leader and I think you will want a check for leader here  to avoid multiple events setting the relation databag

single_kernel_mongo/managers/cluster.py
```diff
@@ -229,6 +243,10 @@ def update_mongos_and_restart(self) -> None:
         key_file_contents = self.state.cluster.keyfile
         config_server_db_uri = self.state.cluster.config_server_uri
 
+        if ldap_user_to_dn_mapping := self.state.cluster.ldap_user_to_dn_mapping:
```
MiaAltieri: will mongos be able to start without this data? 

another question - is this the best place in this file to save `ldap_user_to_dn_mapping ` to the peer databag? Where are `self.state.cluster.keyfile` and `self.state.cluster.config_server_uri` saved?

If this is the correct place to check - you will need a check for leadership since you are writing to the app data bag

single_kernel_mongo/managers/ldap.py
```diff
@@ -205,10 +205,14 @@ def get_ldap_connection_status(self) -> StatusBase:
         bind_dn = self.state.ldap.bind_user
         bind_password = self.state.ldap.bind_password
         base_dn = self.state.ldap.base_dn
+        cert_chain = self.state.ldap.chain
 
         if not base_dn:
             return BlockedStatus("Missing base DN.")
```
MiaAltieri:
```suggestion
            return BlockedStatus("Missing base DN for ldap.")
```
What does the user need to do in this case?

Perhaps the needed user action can be added to the status

Also can you log something here

single_kernel_mongo/managers/ldap.py
```diff
@@ -205,10 +205,14 @@ def get_ldap_connection_status(self) -> StatusBase:
...
 
+        if not cert_chain:
+            return BlockedStatus("Missing chain.")
```
MiaAltieri:
```suggestion
            return BlockedStatus("Missing chain for ldap.")
```

Perhaps the needed user action can be added to the status


Also can you log something here

single_kernel_mongo/managers/ldap.py
```diff
@@ -217,7 +221,7 @@ def get_ldap_connection_status(self) -> StatusBase:
                 tls = LDAPTls(
                     validate=ssl.CERT_REQUIRED,
                     version=ssl.PROTOCOL_TLSv1_2,
-                    ca_certs_file=f"{self.state.paths.ldap_certificates_file}",
+                    ca_certs_data="\n".join(cert_chain),
```
MiaAltieri, line 257: I saw in a sepearate PR you had a `chain ` property that concatenates them together, I assume this will be merged her e

MiaAltieri (changes requested):

single_kernel_mongo/managers/cluster.py
Gu1nness: Yes sure !

single_kernel_mongo/managers/cluster.py
Gu1nness: We can add one more check but it's already checked in the caller.
Let me add it however indeed.

single_kernel_mongo/managers/cluster.py
Gu1nness: > will mongos be able to start without this data?

Yes, definitely it can :)

> another question - is this the best place in this file to save `ldap_user_to_dn_mapping ` to the peer databag? Where are `self.state.cluster.keyfile` and `self.state.cluster.config_server_uri` saved?

I'm storing it in the same location where it's stored on MongoDB.
I'm weighing however moving it to the ldap-peers relation databag to keep a better single responsibility principle of the databags.

> If this is the correct place to check - you will need a check for leadership since you are writing to the app data bag

That's very true, thanks a lot for the clear eyes ;)

single_kernel_mongo/managers/ldap.py
Gu1nness: I'll add some logs with some informations.
All thoses cases should never happen because we already checked that in the caller but I'll log some informations.

single_kernel_mongo/managers/ldap.py
Gu1nness: I have updated the chain property to be consistent: it takes a list, and returns a list so that we always handle the chain as a list of certificates instead of a concatenated string of certificates.
There were multiple cases where it was painful to keep a big string while keeping the list is more reliable and more consistent with what it represents.

single_kernel_mongo/core/operator.py
```diff
@@ -280,9 +280,10 @@ def save_ca_cert_to_trust_store(self, file: TrustStoreFiles, chain: str) -> None
 
     def remove_ca_cert_from_trust_store(self, file: TrustStoreFiles):
         """Removes the certificate from the trust store."""
-        # Remove the file
-        self.workload.delete(TRUST_STORE_PATH / file.value)
-        # Update CA certificates to remove the certificate from the trust store
-        self.workload.exec("update-ca-certificates")
-        # Restart the service
-        self.restart_charm_services(force=True)
+        if self.workload.exists(TRUST_STORE_PATH / file.value):
```
Mehdi-Bendriss: nit: I prefer a negation / return to reduce nestedness

Mehdi-Bendriss (approved):
Wow, really clean and clear! Thanks Neha !!

PR #34, opened 2025-03-20
feat(DPE-6681): Remove logic based on status 

MiaAltieri:
## 🏷 Type of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation Update
- [ ] Tooling and CI changes
- [ ] Dependencies upgrade or change

## ✅ Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the [**CONTRIBUTING**](../blob/6/edge/CONTRIBUTING.md) document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #35, opened 2025-04-03
fix: Enable services and restart on upgrades.

Gu1nness:
## 🏷 Type of changes
- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

With snapd 2.67.1, the behavior of refreshing a snap has changed when it comes to services.

The services are not restarted if they were disabled and started, they will be left disabled and stopped.

In order to fix that, two steps:
 * On start of services, enable them
 * On stop of services, disable them (so that they don't restart unexpectedly upon machine restart)
 * During refreshes, after installing the new snap, start the services (so that the upgrade flow doesn't break)

## 📑 Motivation and Context

This is a bug fix uncovered while working on #34 .

Tested manually on upgrades / rollbacks.

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/core/vm_workload.py
+++ b/single_kernel_mongo/core/vm_workload.py
@@ -48,7 +48,7 @@ class VMWorkload(WorkloadBase):
     @override
     def start(self) -> None:
         try:
-            self.mongod_snap.start(services=[self.service])
+            self.mongod_snap.start(services=[self.service], enable=True)
         except snap.SnapError as e:
             logger.exception(str(e))
             raise WorkloadServiceError(str(e)) from e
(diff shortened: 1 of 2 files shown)


single_kernel_mongo/core/vm_workload.py
```diff
@@ -70,7 +70,7 @@ def update_env(self, parameters: chain[str]):
     @override
     def stop(self) -> None:
         try:
-            self.mongod_snap.stop(services=[self.service])
+            self.mongod_snap.stop(services=[self.service], disable=True)
```
Mehdi-Bendriss, line 73: Do we have to disable on stop?

Mehdi-Bendriss (commented):
Thanks Neha! I have a question

single_kernel_mongo/core/vm_workload.py
Gu1nness: This is something we discussed with Carl [on MM](https://chat.canonical.com/canonical/pl/d6xauhhkptrcig1e79pizybq6r).
Disabling on stop allows to be sure that the restart of the machine will not start a service that should not be running.
It's not mandatory but I think it's a good idea to keep it: when we stop the service, in no way we want it to be able to restart because we are doing some maintenance work that we want to manage completely.

Mehdi-Bendriss (approved):

PR #36, opened 2025-04-16
feat(DPE-6724): LDAP Integration testing

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [x] Tooling and CI changes

## 📝 Description

LDAP integration testing using the new integration testing framework.

## 📑 Motivation and Context
This is the implementation of [DPE-6724](https://warthogs.atlassian.net/browse/DPE-6724) for the ldap integration.
It provides integration testing in the single kernel repository.

This add integration tests and improves the helpers.

## ✅ Checklist

[DPE-6724]: https://warthogs.atlassian.net/browse/DPE-6724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/integration/helpers.py
```diff
@@ -22,15 +27,115 @@
...
+    charm: Path | str,
+    substrate: str,
+    mongod_resource: str,
+    app_name: str,
+    num_units: int = 3,
+    channel: str | None = None,
+    config: dict | None = None,
+    subordinate: bool = False,
+):
+    if substrate == "lxd":
```
Mehdi-Bendriss: This is not entirely accurate (for other machine substrates), the condition should be inverted to test against k8s instead

tests/integration/helpers.py
```diff
@@ -22,15 +27,115 @@
...
+    username = username
+    port = MONGOS_PORT if mongos else MONGOD_PORT
+    hosts = [f"{host}:{port}" for host in hosts]
+    hosts = ",".join(hosts)
+    auth_source = ""
+    database = "admin"
+
+    complement = f"{auth_source}"
+    if not mongos:
+        complement = complement + f"replicaSet={app_name}"
```
Mehdi-Bendriss: I'm not sure I understand this, `auth_source` is always set to `""`

tests/integration/helpers.py
```diff
@@ -296,7 +401,9 @@ async def check_or_scale_app(ops_test: OpsTest, user_app_name: str, required_uni
     await ops_test.model.wait_for_idle()
 
 
-async def get_app_name(ops_test: OpsTest, test_deployments: list[str] = []) -> str:
+async def get_app_name(
+    ops_test: OpsTest, ch_name: str = "mongodb", test_deployments: list[str] = []
```
Mehdi-Bendriss:
```suggestion
    ops_test: OpsTest, charm_name: str = "mongodb", test_deployments: list[str] = []
```

tests/integration/helpers.py
```diff
@@ -438,34 +547,87 @@ async def get_application_units(ops_test: OpsTest, app: str) -> list[Unit]:
...
+        .decode("utf-8")
+        .split("\n")
+    )
+
+    for status_item in juju_status:
+        if app_name not in status_item:
+            continue
+        # no need to check that status of the application since the application can have a
+        # different status than the units.
+        is_app = "/" not in status_item
```
Mehdi-Bendriss: what does a status_item represent here? checking on the `/` seems not very robust

tests/integration/helpers.py
```diff
@@ -22,15 +27,115 @@
...
+    await action.wait()
+    return action.results
+
+
+async def get_address_of_unit(
+    ops_test: OpsTest, substrate: str, unit_id: int, app_name: str
+) -> str:
+    """Retrieves the address of the unit based on provided id."""
+    status: FullStatus = await ops_test.model.get_status()
+    if substrate == "microk8s":
```
Mehdi-Bendriss, line 101: similarly, this is too restrictive, our integration tests will eventually need to be run in other k8s substrates

tests/integration/helpers.py
```diff
@@ -483,3 +645,26 @@ def is_relation_joined(ops_test: OpsTest, endpoint_one: str, endpoint_two: str)
         if endpoint_one in endpoints and endpoint_two in endpoints:
             return True
     return False
+
+
+async def execute_on_mongod(
```
Mehdi-Bendriss, line 649: was this ported from the mongodb-k8s integ tests ?

tests/integration/data/add.ldif
Mehdi-Bendriss, line 1: ldap_entries.ldif

tests/integration/helpers.py
```diff
@@ -22,15 +27,115 @@
...
+DATA_INTEGRATOR_APP_NAME = "data-integrator"
 logger = logging.getLogger(__name__)
 
 
+def mongosh(substrate: str) -> str:
+    match substrate:
+        case "lxd":
+            return "charmed-mongodb.mongosh"
+        case "microk8s":
+            return "mongosh"
```
Mehdi-Bendriss, line 46: can we account for any machine or k8s substrate? the logic should not differ between say ec2 and lxd and microk8s and canonical k8s

tests/integration/ldap_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+
+        second_offer_command = (
+            f"offer {kubernetes_model.info.name}.{LDAP_APP_NAME}:send-ca-cert {LDAP_CERT_OFFER}"
+        )
+
+        await ops_test.juju(*second_offer_command.split())
+
+
+async def consume_offers(ops_test: OpsTest, kubernetes_model: Model):
+    # On k8s consuming an offer on a remote model fails somehow, so we fallback to juju command.
```
Mehdi-Bendriss, line 107: Can you explain this? i'm not sure what you mean by remote model

tests/integration/ldap_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+        await ops_test.juju(*first_offer_command.split())
+
+        second_offer_command = (
+            f"offer {kubernetes_model.info.name}.{LDAP_APP_NAME}:send-ca-cert {LDAP_CERT_OFFER}"
+        )
+
+        await ops_test.juju(*second_offer_command.split())
+
+
+async def consume_offers(ops_test: OpsTest, kubernetes_model: Model):
```
Mehdi-Bendriss:
```suggestion
async def consume_glauth_offers(ops_test: OpsTest, kubernetes_model: Model):
```

Given those are helper classes it would make sense to make it clear to avoid collisions

tests/integration/ldap_helpers.py
Mehdi-Bendriss, line 1: Missing docstrings and return type hints in this module

tests/integration/ldap_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+    await ops_test.juju(*first_remove_offer_command.split())
+
+    second_remove_offer_command = (
+        f"remove-offer admin/{kubernetes_model.name}.{LDAP_CERT_OFFER} --force --yes"
+    )
+    logger.info("Removing ldap cert offer")
+    await ops_test.juju(*second_remove_offer_command.split())
+
+
+async def create_groups(
```
Mehdi-Bendriss: perhaps?
```suggestion
async def create_mongodb_user_roles(
```

tests/integration/ldap_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+async def generate_mongodb_ldap_client(
+    ops_test: OpsTest,
+    substrate: str,
+    app_name: str,
+    database: str,
+    username: str,
+    password: str,
+    mongos: bool = False,
+) -> str:
+    if mongos and substrate == "lxd":
```
Mehdi-Bendriss, line 162: similar comment on substrate check

tests/integration/mongodb/ldap/test_ldap.py
Mehdi-Bendriss, line 1: Missing docstrings in some tests

tests/integration/mongodb/ldap/test_ldap.py
```diff
@@ -0,0 +1,230 @@
...
+    await wait_for_mongodb_units_blocked(
+        ops_test,
+        db_app_name,
+        status="TLS is mandatory for LDAP transport.",
+        timeout=300,
+    )
+
+
+@pytest.mark.abort_on_fail
+async def test_integrate_also_ldap_cert(ops_test: OpsTest):
```
Mehdi-Bendriss:
```suggestion
async def test_integrate_ldap_cert(ops_test: OpsTest):
```

tests/integration/mongodb/ldap/test_ldap.py
```diff
@@ -0,0 +1,230 @@
...
+    uri = await generate_mongodb_ldap_client(
+        ops_test,
+        substrate,
+        db_app_name,
+        database="superdb",
+        username="cn=johndoe,ou=superheroes,ou=users,dc=glauth,dc=com",
+        password="dogood",
+    )
+
+    await execute_on_mongod(ops_test, db_app_name, substrate, uri, "db.test.insertOne({number: 1})")
```
Mehdi-Bendriss, line 134: We should also add a check that the record has indeed been written

tests/integration/mongodb/ldap/test_ldap.py
```diff
@@ -0,0 +1,230 @@
...
+            "ldap-user-to-dn-mapping": '[{"match": "([^@]+)@([^@]+)", "substitution": "cn={0},ou={1},ou=users,dc=glauth,dc=com"}]',
+            "ldap-query-template": "dc=glauth,dc=com??sub?(&(objectClass=posixGroup)(uniqueMember={USER}))",
+        }
+    )
+
+    await ops_test.model.wait_for_idle(apps=[db_app_name], status="active", timeout=TIMEOUT)
+
+    path = (
+        "/var/snap/charmed-mongodb/current/etc/mongod/mongod.conf"
+        if substrate == "lxd"
```
Mehdi-Bendriss, line 160: perhaps checking if not a k8s model

tests/integration/mongodb/ldap/test_ldap.py
```diff
@@ -0,0 +1,230 @@
...
+    for unit in ops_test.model.applications[db_app_name].units:
+        cat_cmd = cat_cmd.format(unit.name, path)
+        return_code, output, err = await ops_test.juju(*cat_cmd.split())
+
+        if return_code != 0:
+            raise ProcessError(f"Could not cat configuration. {output=} {err=}")
+
+        configuration = safe_load(output)
+        assert configuration["security"]["ldap"].get("userToDNMapping", "") != ""
+        assert configuration["security"]["ldap"]["authz"].get("queryTemplate", "") != ""
```
Mehdi-Bendriss: We can improve this with an exact equality check on the expected value, and add assert failures messages

tests/integration/mongodb/ldap/test_ldap.py
```diff
@@ -0,0 +1,230 @@
...
+
+    uri = await generate_mongodb_ldap_client(
+        ops_test,
+        substrate,
+        db_app_name,
+        database="superdb",
+        username="johndoe@superheroes",
+        password="dogood",
+    )
+    await execute_on_mongod(ops_test, db_app_name, substrate, uri, "db.test.insertOne({number: 2})")
```
Mehdi-Bendriss, line 195: Could you add a check that the record was indeed persisted?

tests/integration/mongodb/ldap/test_ldap.py
```diff
@@ -0,0 +1,230 @@
...
+    # John should not be able to log in now.
+    uri = await generate_mongodb_ldap_client(
+        ops_test,
+        substrate,
+        db_app_name,
+        database="superdb",
+        username="johndoe@superheroes",
+        password="dogood",
+    )
+
```
Mehdi-Bendriss, line 232: for posterity, could you add a comment stating that we should expect the write to fail when the the ldap relation is missing?

tests/integration/mongodb/ldap/test_ldap.py
```diff
@@ -0,0 +1,230 @@
...
+        db_app_name,
+        database="superdb",
+        username="johndoe@superheroes",
+        password="dogood",
+    )
+    await execute_on_mongod(ops_test, db_app_name, substrate, uri, "db.test.insertOne({number: 2})")
+
+
+@pytest.mark.abort_on_fail
+async def test_remove_ldap_goes_to_blocked(ops_test: OpsTest, substrate: str):
```
Mehdi-Bendriss, line 201: Can we also add a test when we remove the relation with `send-ca-cert`?

tests/integration/sharding_helpers.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+
+    await ops_test.model.wait_for_idle(
+        apps=[config_server_name, shard_one_name, shard_two_name],
+        idle_period=20,
+        timeout=DEPLOYMENT_TIMEOUT,
+    )
+
+
+async def integrate_cluster(
```
Mehdi-Bendriss:
```suggestion
async def integrate_sharding_components(
```

tests/integration/mongos/ldap/test_ldap.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # deploy the glauth-k8s charm
+    await deploy_glauth(ops_test, kubernetes_model)
+    await consume_offers(ops_test, kubernetes_model)
+
+    db_app_name = CONFIG_SERVER_APP_NAME
+
+    await create_groups(
+        ops_test, substrate, db_app_name, "ou=superheroes,ou=users,dc=glauth,dc=com"
+    )
```
Mehdi-Bendriss: nit:
```suggestion
    await create_groups(
        ops_test, substrate, CONFIG_SERVER_APP_NAME, "ou=superheroes,ou=users,dc=glauth,dc=com"
    )
```

tests/integration/mongos/ldap/test_ldap.py
Mehdi-Bendriss, line 1: missing docstrings in multiple tests in this module

tests/integration/mongos/ldap/test_ldap.py
```diff
@@ -0,0 +1,215 @@
...
+            substrate=substrate,
+            mongod_resource=mongod_resource,
+            app_name=base_app_name,
+            num_units=3,
+            subordinate=(substrate == "lxd"),
+        )
+        app_name = base_app_name
+
+    await ops_test.model.deploy(
+        DATA_INTEGRATOR_APP_NAME,
```
Mehdi-Bendriss, line 112: This is excusively needed for sharding on VM, no?

tests/integration/mongos/ldap/test_ldap.py
```diff
@@ -0,0 +1,215 @@
...
+        ops_test,
+        app_name,
+        status="mongos and config-server not integrated with the same ldap server.",
+        timeout=300,
+        subordinate=(substrate == "lxd"),
+    )
+
+
+@pytest.mark.abort_on_fail
+async def test_all_integrated(ops_test: OpsTest):
```
Mehdi-Bendriss:
```suggestion
async def test_glauth_fully_integrated(ops_test: OpsTest):
```

tests/integration/mongos/ldap/test_ldap.py
```diff
@@ -0,0 +1,215 @@
...
+    )
+    await ops_test.model.wait_for_idle(
+        apps=[CONFIG_SERVER_APP_NAME, SHARD_ONE_APP_NAME, SHARD_TWO_APP_NAME, app_name],
+        idle_period=20,
+        status="active",
+    )
+
+
+@pytest.mark.abort_on_fail
+async def test_only_mongos_integrated(ops_test: OpsTest, substrate: str):
```
Mehdi-Bendriss:
```suggestion
async def test_glauth_only_integrated_with_mongos(ops_test: OpsTest, substrate: str):
```

tests/integration/mongos/ldap/test_ldap.py
```diff
@@ -0,0 +1,215 @@
...
+    )
+
+    await execute_on_mongod(
+        ops_test,
+        app_name,
+        substrate,
+        uri,
+        "db.test.insertOne({number: 1})",
+        container_name="mongos",
+    )
```
Mehdi-Bendriss, line 205: similar comment on adding a check on read

Mehdi-Bendriss (commented):
Thank you Neha, great work! I left a few comments and questions

tests/integration/helpers.py
Gu1nness: OK

tests/integration/helpers.py
Gu1nness: It's a miss, I first wanted to add the auth source and then I forgot.
Let me remove it for now.

tests/integration/helpers.py
Gu1nness: This is copy pasted from mongos-operator tbh.
A status item is a line of the status, and this should be made more robust in the future but I was in a bit of a hurry.
It the best world, we would parse the json output and knowing which app we are subordinate to, we can get the unit status.

I can rename `status_item` to `status_line` maybe?

tests/integration/helpers.py
Gu1nness: Maybe when the time comes we can update it to a better mapping ?

For now there is not much difference and we can also "just" rename the substrates to `vm` vs `k8s`, it was merely a naming choice made during the integration test PR.
Furthermore, if you look at the substrate fixture, you'll see that for now it accepts only two values so some updates will be necessary anyway.

tests/integration/helpers.py
Gu1nness: Yes, with an update to handle the lxd case.
The reason being that we cannot connect from the host to a full replicaSet on kubernetes so we have to execute the command on the pod instead.

tests/integration/helpers.py
Gu1nness: Same answer as before, it can be renamed, it was a choice made during the integration test PR.
Can be renamed without a lot of issues I think but this brings many small changes in many files in the whole testing framework.

tests/integration/ldap_helpers.py
Gu1nness: We have two models, one for glauth, one for the mongoDB deployment.
When I tried using the `ops_test.model.consume` it was failing due to a missing file somewhere (I don't remember which one).
I need to open a ticket for this one (I guess in ops_test or in libjuju) but I fell back to using juju commands instead as they were working.

tests/integration/helpers.py
Gu1nness: As discussed, lets keep it that way and open another PR later.

tests/integration/mongodb/ldap/test_ldap.py
Gu1nness: Ok !

tests/integration/mongodb/ldap/test_ldap.py
Gu1nness: So add back the relation and remove the other one and ensure that it still works ?

tests/integration/mongos/ldap/test_ldap.py
Gu1nness: Yes, but it is easier for those tests to follow the exact same workflow.
I'm still working on how to make those tests extremely clean, and how we will have different workflows for different substrates.
In that case, it doesn't change what we test so I'd rather keep it (and even more if we create roles with data integrator in the future, this will already be there).

Gu1nness: TODO: Add comment for posterity.

Mehdi-Bendriss (approved):

PR #39, opened 2025-05-02
patch: new release workflow

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [x] Tooling and CI changes

## 📝 Description

New release workflow with dynamic versioning of the package.

## 📑 Motivation and Context

Improve the current release workflow to go towards continuous deployment of the library.

This is the integration of the new workflow from DP Workflows, replacing the current one.

## ✅ Checklist


pyproject.toml
```diff
@@ -54,14 +54,23 @@ packages = [
...
 [tool.poetry.requires-plugins]
 poetry-plugin-export = ">=1.8"
+poetry-dynamic-versioning = { version = ">=1.0.0,<2.0.0", extras = ["plugin"] }
+
+[tool.poetry-dynamic-versioning]
+enable = true
+vcs = "git"
+dirty = true
+strict = true
+fix-shallow-repository = true
```
carlcsaposs-canonical, line 68:
```suggestion
```
would recommend fixing in the checkout instead of relying on poetry-dynamic-versioning to attempt to fix it

e.g. https://github.com/canonical/data-platform-workflows/blob/b2f382151057749f3a3674e502519ba7d67a26a6/.github/workflows/release_python_package_part1.yaml#L26-L29

scripts/build_lib_for_integration.sh
```diff
@@ -29,6 +29,17 @@ for directory in "${TEST_CHARMS[@]}"; do
...
+    # Backup files
+    cp pyproject.toml pyproject.toml.backup
+    cp poetry.lock poetry.lock.backup
+
+    # Disable strict mode for build test lib.
+    pushd "${LIB_PATH}"
+    git init
+    sed 's/strict = true/strict = false/' -i "pyproject.toml"
+    cat pyproject.toml
+    popd
```
carlcsaposs-canonical, line 41: you may run into issues disabling strict mode

this is not supported (i.e. if you encounter issues because of this, don't expect support in dpw)

carlcsaposs-canonical (commented):
(reviewing per request of @Mehdi-Bendriss)

The release python package reusable workflow is not intended for this use case

It's intended to be used for python packages that expose a public API & are semantically versioned (e.g. shared code used by several charms such as refresh v3, relation interfaces, etc.)

For this use case, I don't think a public API is desired (since changes to the shared code & consumers of the shared code are done in parallel and the consumers of the shared code always use the latest version [i.e. there effectively isn't any versioning in the consumers])

For your use case, it could be worth considering copying a subset of the workflow & using that directly (e.g. something like https://github.com/canonical/charmcraftcache/blob/12d388b788deb55634d0fa976ee132afbd5b8046/.github/workflows/publish.yaml) since I think the DPW workflow may be more complicated than what would be desired for your use case

With that said, if this works as a temporarily solution until we have the approval/infrastructure figured out for a monorepo, feel free to use it

Just have these expectations:
- this is not supported by DPW
- we will not make changes to the DPW workflow to accommodate this use case
- we may make breaking changes or bug fixes to the DPW workflow to better accommodate the semantic version case which may break this usage (that might happen in a patch version bump of dpw, since this usage is not considered a usage of the public API, but those changes would likely happen in a major version bump)

Mehdi-Bendriss (approved):

pyproject.toml
Gu1nness: This checkout comes from the pre-commit tool, and this is the recommended solution from the developer of `poetry-dynamic-versioning` [here](https://github.com/mtkennerly/poetry-dynamic-versioning/discussions/132).
I agree that it is not the best solution, I do however enjoy having the hooks run during the CI as a first step to block poorly formatted code.

scripts/build_lib_for_integration.sh
Gu1nness: This comes from the same issue, due to dynamic versioning.

Given your other comment, but considering that I want to have dynamic versioning enabled, I'm wondering what would be the best in our case.
Still, this is for test charms only and if it starts to fail there, maybe it will be time to fix it.

Otherwise I go for a homebrew solution but I'd rather follow the standards of DPW until we have a proper solution for monorepo?

carlcsaposs-canonical: > This comes from the same issue, due to dynamic versioning.

to check if I'm understanding—this issue comes from pre-commit?

strict mode means that the build will fail if the dynamic version cannot be determined—in other words, if it only works without strict mode, that means the versioning is not working (the version will always be 0.0.0)

Gu1nness: No it does not come from the pre-commit but it comes when we build the test charms. The reasons for that seems to arise from the lack of git history that would allow to get tags.
When I build it without strict mode, it will fail with:
```
PEP517 build of a dependency failed

Backend subprocess exited when trying to invoke build_wheel

    | Command '['/tmp/tmpt8lso221/.venv/bin/python', '/home/neha/.local/share/mise/installs/python/3.10.15/lib/python3.10/site-packages/pyproject_hooks/_in_process/_in_process.py', 'build_wheel', '/tmp/tmpt0atn567']' returned non-zero exit status 1.
    |
    | Traceback (most recent call last):
    |   File "/home/neha/.local/share/mise/installs/python/3.10.15/lib/python3.10/site-packages/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
    |     main()
    |   File "/home/neha/.local/share/mise/installs/python/3.10.15/lib/python3.10/site-packages/pyproject_hooks/_in_process/_in_process.py", line 373, in main
    |     json_out["return_val"] = hook(**hook_input["kwargs"])
    |   File "/home/neha/.local/share/mise/installs/python/3.10.15/lib/python3.10/site-packages/pyproject_hooks/_in_process/_in_process.py", line 280, in build_wheel
    |     return _build_backend().build_wheel(
    |   File "/tmp/tmpt8lso221/.venv/lib/python3.10/site-packages/poetry/core/masonry/api.py", line 55, in build_wheel
    |     poetry = Factory().create_poetry(Path().resolve(), with_groups=False)
    |   File "/tmp/tmpt8lso221/.venv/lib/python3.10/site-packages/poetry_dynamic_versioning/patch.py", line 26, in alt_poetry_create
    |     name = _get_and_apply_version(
    |   File "/tmp/tmpt8lso221/.venv/lib/python3.10/site-packages/poetry_dynamic_versioning/__init__.py", line 735, in _get_and_apply_version
    |     version, instance = _get_version(config, name)
    |   File "/tmp/tmpt8lso221/.venv/lib/python3.10/site-packages/poetry_dynamic_versioning/__init__.py", line 542, in _get_version
    |     version = _get_version_from_dunamai(vcs, pattern, config)
    |   File "/tmp/tmpt8lso221/.venv/lib/python3.10/site-packages/poetry_dynamic_versioning/__init__.py", line 504, in _get_version_from_dunamai
    |     return Version.from_vcs(
    |   File "/tmp/tmpt8lso221/.venv/lib/python3.10/site-packages/dunamai/__init__.py", line 2125, in from_vcs
    |     return cls._do_vcs_callback(
    |   File "/tmp/tmpt8lso221/.venv/lib/python3.10/site-packages/dunamai/__init__.py", line 2180, in _do_vcs_callback
    |     return callback(**kwargs)
    |   File "/tmp/tmpt8lso221/.venv/lib/python3.10/site-packages/dunamai/__init__.py", line 1167, in from_git
    |     return cls._fallback(strict, distance=0, dirty=True, branch=branch, concerns=concerns, vcs=vcs)
    |   File "/tmp/tmpt8lso221/.venv/lib/python3.10/site-packages/dunamai/__init__.py", line 1001, in _fallback
    |     raise RuntimeError("No tags available and fallbacks disabled by strict mode")
    | RuntimeError: No tags available and fallbacks disabled by strict mode
```

I'd rather not create a git history with a single commit and a single empty tag as I don't think it is very clean to do so, especially on test charms.
What do you think?

carlcsaposs-canonical: I think I'm missing something

is the error when building the wheel for the single kernel python package? if so, is there a reason the package isn't being cloned with git history (e.g. with https://github.com/canonical/data-platform-workflows/blob/b2f382151057749f3a3674e502519ba7d67a26a6/.github/workflows/release_python_package_part1.yaml#L26-L29)?

> I'd rather not create a git history with a single commit and a single empty tag [...] especially on test charms

I don't think it should be necessary to create a git history

Gu1nness: The error happens when building the wheel for the test charms.
The main reason not to clone is during local development, to have the exact code on the machine (it might not be committed or pushed yet).
My usual development would be:
```
* change some code in the library
* bash scripts/build_lib_for_integration.sh
* tox -e integration -- ...
* if that works, commit.
```
For that to work, we copy the library code + pyproject.toml file in each of the 4 test charms and build.
This will fail without `strict = false` because the copied version is lacking the git history.

carlcsaposs-canonical: > when building the wheel for the test charms

As far as I understand, charmcraft does not build a wheel for charms (i.e. it does not install the charm as a python package)

> The main reason not to clone is during local development

This is exactly why I wouldn't recommend using the DPW workflow—I think the ideal case is to have a local python package that is not versioned separately from the charms—I don't think the dynamic versioning/semantic version bumping in the DPW workflow is behavior that you want (I would suggest just copy-pasting a subset of the dpw workflow that does what you want until we have the monorepo infrastructure figured out)

Gu1nness: One of the requirements is still to have some dynamic versioning to have automated bumps of the version on every push to main so we might need dynamic versioning but maybe not the semantic versioning that you added.
I'll go back to this tomorrow!

carlcsaposs-canonical (approved):
Approving with the understanding that
- this is a temporarily solution until we have the approval/infrastructure figured out for a monorepo
- this is not supported by DPW & not how this workflow is intended to be used
- we do not plan to make changes to the DPW workflow to accommodate this use case
- we may make breaking changes or bug fixes to the DPW workflow to better accommodate the semantic version case which may break this usage (that might happen in a patch version bump of dpw, since this usage is not considered a usage of the public API)

would you be willing to add a comment to each of the uses of dpw indicating that this usage is not supported & not how the workflows are intended to be used?

Gu1nness:
This is indeed a temporary solution and this is not fully supported by Data Platform Workflows.
If anything breaks, I'll fix it here and nothing will get changed on DPW.

When we have the monorepo helpers, all this will be reworked with the new helpers and we'll proceed with that.

PR #46, opened 2025-06-26
patch(DPE-7780): enable MongoDB Exporter and update-status order

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

Three small bug fixes in this commit:
 * Enable mongodb-exporter systemd service so that it does not stop on refreshes
 * Don't block update-status if the mongodb-exporter is not running. Only blocking checks should prevent that
 * Correct order for update-status handler: operator and then status-handler to have up to date statuses

## 📑 Motivation and Context

Those are a few bugfixes discovered when playing around with snap upgrades.

Manually tested (I'll run the integration tests)

## ✅ Checklist


single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -655,7 +655,7 @@ def update_status(self) -> None:
         """Status update Handler."""
         # TODO update the usage of this once the spec is approved and we have a consistent way of
         # handling statuses
-        if self.get_statuses(scope="unit", recompute=True):
+        if self.basic_statuses():
```
Mehdi-Bendriss, line 658: I thought we should be recomputing all statuses on update statuses?

Mehdi-Bendriss (changes requested):
I only have a comment on the recomputing part of the statuses that's not clear to me.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: The statuses are recomputed in the update-status event handled by the library.
Those are the safeguard checks and if you look at `basic_statuses` it does some computations instead of just looking in the databag.

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):
Thanks Neha

PR #48, opened 2025-07-08
patch(DPE-7671):  Add logrotate user to charm and adjust the logrotate routine

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

- Adds a new logrotate user to the charm with extremely limited permissions to allow rotating logs only for audit or server.
- Split the logrotate config in two parts to rotate based on which file needs rotation.
- Extend the size of the file and number of files to keep.
- Ensure the user is created upon upgrade.

## 📑 Motivation and Context

When testing the latest MongoDB 6 snap, it appeared that there was an issue with the way we were handling log rotation.
The two log files `audit.log` and `mongodb.log` were rotated all the time even if they were not big enough

Integration tests + manual tests.

## ✅ Checklist


single_kernel_mongo/managers/config.py
```diff
@@ -177,20 +177,31 @@ def __init__(
...
-            self.workload.setup_cron(
-                [
-                    f"* 1-23 * * * root logrotate {LogRotateConfig.rendered_template}\n",
-                    f"1-59 0 * * * root logrotate {LogRotateConfig.rendered_template}\n",
-                ]
-            )
-        else:
-            self.workload.start()
+        if not self.state.db_initialised:
+            return
```
Mehdi-Bendriss, line 186: worth adding a log line here - just like we do in the `BackupConfigManager`

single_kernel_mongo/managers/config.py
```diff
@@ -177,20 +177,31 @@ def __init__(
...
+
+        if not self.state.get_user_password(LogRotateUser):
+            return
```
Mehdi-Bendriss, line 190: similar comment here

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -707,6 +708,8 @@ def set_password(self, username: str, password: str | None = None) -> tuple[str,
         if user == MonitorUser:
             # Update and restart mongodb exporter.
             self.mongodb_exporter_config_manager.configure_and_restart()
+        if user == LogRotateUser:
```
Mehdi-Bendriss, line 706: nit: could you add a comment for consistency like we do on the 2 previous conditions

single_kernel_mongo/managers/config.py
```diff
@@ -177,20 +177,31 @@ def __init__(
...
+            self.workload.build_template()
+            if self.substrate == Substrates.VM:
+                self.workload.setup_cron(
+                    [
+                        f"* 1-23 * * * root logrotate {LogRotateConfig.rendered_template}\n",
+                        f"1-59 0 * * * root logrotate {LogRotateConfig.rendered_template}\n",
+                    ]
+                )
+            else:
+                self.workload.restart()
```
Mehdi-Bendriss, line 203: Shouldn't we catch a `WorkloadServiceError` here? perhaps return on the previous condition and handle it without the else

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -933,9 +938,6 @@ def _configure_workloads(self) -> None:
         self.config_manager.set_environment()
         self.mongos_config_manager.set_environment()
 
-        # Start logrotate
-        self.logrotate_config_manager.configure_and_restart()
```
Mehdi-Bendriss, line 932: are we removing this because it now requires the user creation?

single_kernel_mongo/managers/upgrade.py
```diff
@@ -75,6 +76,8 @@ def _kubernetes_upgrade(self) -> None:
                 self.state.unit_upgrade_peer_data.current_revision = (
                     self.dependent.cross_app_version_checker.version  # type: ignore
                 )
+                # If the user was not existing yet, create it
```
Mehdi-Bendriss: could you add, "we're introducing a new user that revisions predating XYZ didn't have, so we need to create it if ......"

tests/integration/mongodb/test_charm.py
```diff
@@ -378,11 +379,20 @@ async def test_log_rotate(ops_test: OpsTest, substrate: Substrate, application_p
     log_not_rotated = "audit.log.1" not in log_files
     assert log_not_rotated, f"Found rotated log in {log_files}"
 
-    await start_continous_writes(ops_test, client_app_name=application_name)
-    time.sleep(time_to_write_50m_of_data)
-    await stop_continous_writes(ops_test, client_app_name=application_name)
+    for i in range(5):
```
Mehdi-Bendriss, line 383: could you add a small comment here - to explain why we are iterating at all (future readers will not understand unless we tell them it's to speed up writes)

Mehdi-Bendriss (commented):
Thanks Neha, pretty clean work. I left small comments, but mainly waiting on CI to pass.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Yes !

single_kernel_mongo/managers/upgrade.py
Gu1nness: Yeah but I'm realising that it's a bit harder than what I thought, I'll move those things around: Only the leader can do it and it should check if it needs to (and if it does, create also the password).
So I'll rework that section.

single_kernel_mongo/managers/config.py
```diff
@@ -201,7 +200,11 @@ def configure_and_restart(self) -> None:
                     ]
                 )
             else:
+                self.set_environment()
```
Mehdi-Bendriss: This is already done earlier

single_kernel_mongo/managers/upgrade.py
```diff
@@ -63,6 +63,15 @@ def _kubernetes_upgrade(self) -> None:
         assert self._upgrade
         if self.charm.unit.is_leader() and self.dependent.name == CharmKind.MONGOD:
             self.dependent.cross_app_version_checker.set_version_across_all_relations()  # type: ignore
+
+            # If the user was not existing yet, create it.
+            # This user was added after upgrades where developed so we have to
```
Mehdi-Bendriss:
```suggestion
            # This user was added after the first stable release so we have to
```

single_kernel_mongo/managers/upgrade.py
```diff
@@ -100,17 +107,21 @@ def _vm_upgrade(self):
...
-                logger.info("Charm refreshed. MongoDB version unchanged")
-            if self.dependent.name == CharmKind.MONGOD:
-                self.state.app_upgrade_peer_data.upgrade_resumed = False
-                self.dependent.cross_app_version_checker.set_version_across_all_relations()  # type: ignore
+        if self.charm.unit.is_leader() and not self.state.upgrade_in_progress:
+            logger.info("Charm refreshed. MongoDB version unchanged")
 
         if self.dependent.name == CharmKind.MONGOD and self.charm.unit.is_leader():
+            # If the user was not existing yet, create it.
+            # This user was added after upgrades where developed so we have to
```
Mehdi-Bendriss:
```suggestion
            # This user was added after the first stable release so we have to
```

Mehdi-Bendriss (commented):
Thanks Neha! minor comments

Mehdi-Bendriss (approved):

PR #52, opened 2025-07-18
patch(backup): TLS on backups.

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

This commit adds the support for CA propagation using the shard relation
This commit fixes the serialisation bug.
This commit ensures removal of the files if necessary.

## 📑 Motivation and Context

The TLS support on sharded deployment was bugged:
 * The certificates were not propagated to shards
 * The list of certificate was not serialised befor being written
This has been found by Mateo during a test deployment.

This has been manually tested only for now but I'll add some integration tests in the PR very soon.

## ✅ Checklist


single_kernel_mongo/events/sharding.py
```diff
@@ -167,6 +168,7 @@ def _on_relation_broken(self, event: RelationBrokenEvent):
         """On relation broken, we drain the shard before allowing it to disconnect."""
         try:
             self.manager.drain_shard_from_cluster(event.relation)
+            self.dependent.remove_ca_cert_from_trust_store(TrustStoreFiles.PBM)
```
Mehdi-Bendriss, line 171: why is this particularly needed on relation broken?

single_kernel_mongo/managers/backups.py
```diff
@@ -787,6 +790,24 @@ def _remap_replicaset(self, backup_id: str) -> str | None:
         )
         return f"{current_cluster_name}={old_cluster_name}"
 
+    def share_certificate_to_shards(self, ca_chain_list: list[str]):
```
Mehdi-Bendriss:
```suggestion
    def share_certificate_with_shards(self, ca_chain_list: list[str]):
```

single_kernel_mongo/managers/sharding.py
```diff
@@ -112,6 +116,13 @@ def prepare_sharding_config(self, relation: Relation) -> None:
             AppShardingComponentKeys.HOST.value: json.dumps(sorted(self.state.internal_hosts)),
         }
 
+        if self.state.s3_relation:
+            credentials = self.dependent.backup_events.s3_client.get_s3_connection_info()
+            if cert_chain_list := credentials.get("tls-ca-chain", None):
```
Mehdi-Bendriss, line 122: very nice catch!

single_kernel_mongo/managers/sharding.py
```diff
@@ -598,6 +609,16 @@ def synchronise_cluster_secrets(self, relation: Relation, leaving: bool = False)
...
 
+        # Add the certificate if it is present
+        if backup_tls_chain := self.state.shard_state.backup_ca_secret:
+            self.dependent.save_ca_cert_to_trust_store(TrustStoreFiles.PBM, backup_tls_chain)
+        else:
+            # If it is not in the databag, always remove it, it won't change a
+            # thing if the file is not present, remove_ca_cert_from_trust_store will early return.
+            self.dependent.remove_ca_cert_from_trust_store(TrustStoreFiles.PBM)
+        # We updated the configuration, so we restart PBM.
+        self.dependent.backup_manager.configure_and_restart(force=True)
```
Mehdi-Bendriss: I don't think we should _always_ restart even when no ca_chain is passed / removed. I would rather we only do it when there is an addition / removal of the CA chain

single_kernel_mongo/managers/sharding.py
```diff
@@ -645,6 +666,16 @@ def handle_secret_changed(self, secret_label: str | None) -> None:
...
 
+        # If the relation was already established, only the secret will be changed
+        if backup_tls_chain := self.state.shard_state.backup_ca_secret:
+            self.dependent.save_ca_cert_to_trust_store(TrustStoreFiles.PBM, backup_tls_chain)
+        else:
+            # If it is not in the databag, always remove it, it won't change a
+            # thing if the file is not present, remove_ca_cert_from_trust_store will early return.
+            self.dependent.remove_ca_cert_from_trust_store(TrustStoreFiles.PBM)
+        # We updated the configuration, so we restart PBM.
+        self.dependent.backup_manager.configure_and_restart(force=True)
```
Mehdi-Bendriss: similar comment

Mehdi-Bendriss (commented):
Thanks Neha! Really good catch :) I have a few small comments, but looking very good

single_kernel_mongo/managers/sharding.py
Gu1nness: Ok, I'll rewrite a bit that part so that we detect when it's been removed and when it's been added

single_kernel_mongo/events/sharding.py
Gu1nness: I wanted to ensure that shards would be properly cleaned especially given that we decide if the certs where added by comparing the file on the fs and the file on the databag.
Better ensure that the shard cannot access PBM anymore if it was configured when removing the shard.

Mehdi-Bendriss (commented):
Thanks Neha! Looks good now. 
I do have one comment left regarding exception handling, it seems that in a few places, `WorkloadServiceError` is not caught (e.g: `set_password_action` - where the only exception caught is `NonDeferrableError`.

Could you please inspect carefully the bubbling of exceptions?

Mehdi-Bendriss (approved):
Thanks Neha!

PR #53, opened 2025-07-24
patch(DPE-7576): New MongoDB 6.0.24 snap

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

This PR bumps the version of the snap and rock to use the latest MongoDB 6 revision: 6.0.24

## 📑 Motivation and Context

Prepare the ground for a new stable release and for the major upgrade from Mongodb 6 to 8
CF [this JIRA ticket](https://docs.google.com/document/d/1GU_05gjTC4MvVERqm8Oai7buGLUtrw4ooxMQXV0BwGE/edit?usp=sharing)

Integration tests here, snap tested and rock tested before.

## ✅ Checklist


single_kernel_mongo/config/statuses.py
```diff
@@ -116,13 +116,29 @@ class BackupStatuses(Enum):
...
-    PBM_INCORRECT_CREDS = StatusObject(status="blocked", message="s3 credentials are incorrect.")
+    PBM_INCORRECT_CREDS = StatusObject(
+        status="blocked",
+        message="s3 credentials are incorrect.",
+        action="Check S3 credentials on s3-integrator",
+    )
     PBM_INCOMPATIBLE_CONF = StatusObject(
-        status="blocked", message="s3 configurations are incompatible."
+        status="blocked",
+        message="s3 configurations are incompatible.",
```
Mehdi-Bendriss:
```suggestion
        message="s3 config options are incompatible.",
```

single_kernel_mongo/config/statuses.py
```diff
@@ -116,13 +116,29 @@ class BackupStatuses(Enum):
...
+        message="Unknown PBM error, check logs.",
+        action="Check logs for more information",
     )
-    UNKNOWN_PBM_ERROR = StatusObject(status="blocked", message="Unknown PBM error, check logs.")
     CANT_CONFIGURE = StatusObject(status="blocked", message="Couldn't configure s3 backup options.")
     ACTIVE_IDLE = StatusObject(status="active", message="")
+    FAILED_TO_CREATE_BUCKET = StatusObject(
+        status="blocked",
+        message="Failed to create buckets, check logs.",
+        action="Check S3 configuration",
```
Mehdi-Bendriss:
```suggestion
        action="Check S3 configuration on s3-integrator.",
```

single_kernel_mongo/events/backups.py
```diff
@@ -119,12 +121,41 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent) -> None:  #
             )
             return
 
+        if not self.manager.validate_s3_config():
+            logger.info(
```
Mehdi-Bendriss: Probably `.warning` ?

single_kernel_mongo/managers/backups.py
```diff
@@ -141,6 +153,92 @@ def backup_id(self) -> str:
...
+            else:
+                bucket.create()
+            bucket.wait_until_exists()
+        except ClientError as e:
+            if (
+                # AWS returns these if the bucket was already created
+                "BucketAlreadyOwnedByYou" in e.args[0]
+                or "BucketAlreadyExists" in e.args[0]
+                # GCP returns this if the bucket was already created
+                or "BucketNameUnavailable" in e.args[0]
```
Mehdi-Bendriss: when would this ever happen? given you're already checking on the existence of the bucket earlier and return if it exists

single_kernel_mongo/events/backups.py
```diff
@@ -119,14 +121,42 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent) -> None:  #
...
         credentials = self.s3_client.get_s3_connection_info()
 
         try:
+            # First create the bucket if it does not exist.
+            self.manager.create_bucket(credentials=credentials)
+            # Then set the config options on PBM.
             self.manager.set_config_options(credentials=credentials)
+            # Finally, resync the configuration.
             self.manager.resync_config_options()
+        except InvalidS3CredentialsError:
```
Mehdi-Bendriss, line 146: We are missing the handling of a couple of exceptions here `SSLError, ConnectionTimeoutError`

single_kernel_mongo/managers/backups.py
```diff
@@ -141,6 +153,92 @@ def backup_id(self) -> str:
...
+                # boto3 client will need the certificate on the node that runs the command
+                fd.write("\n".join(tls_ca_chain))
+
+        bucket = self._get_bucket_resource(credentials)
+
+        try:
+            bucket.meta.client.head_bucket(Bucket=bucket_name)
+            logger.info(f"Using existing bucket {bucket_name}")
+            exists = True
+        except ConnectTimeoutError as e:
```
Mehdi-Bendriss: This is worth putting in a Tenacity retry context manager with a few second wait, to potentially resolve the `ConnectionTimeoutError`

single_kernel_mongo/managers/backups.py
```diff
@@ -308,6 +412,14 @@ def backup_state(self) -> BackupState:
         if not self.workload.active():
             return BackupState.WAITING_PBM_START
 
+        try:
+            credentials = self.dependent.backup_events.s3_client.get_s3_connection_info()
+            self.create_bucket(credentials)
+        except InvalidS3CredentialsError:
+            return BackupState.INCORRECT_CREDS
+        except FailedToCreateS3BucketError:
```
Mehdi-Bendriss: We are missing the handling of a couple of exceptions here `SSLError, ConnectionTimeoutError`

single_kernel_mongo/managers/backups.py
```diff
@@ -426,7 +540,8 @@ def validate_s3_config(self) -> bool:
             logger.info("Missing region - this is required for AWS and GCP")
 
         if not provided_configs.get("storage.s3.endpointUrl"):
-            logger.info("Missing endpointUrl - this is required for MinIO and GCP")
+            logger.info("Missing s3 endpoint.")
+            return False
```
Mehdi-Bendriss, line 536: If I understand correctly, this should be False only in the case of MinIO and GCP, not aws - in which case the config will never be valid even it technically is.

single_kernel_mongo/managers/backups.py
```diff
@@ -676,9 +809,13 @@ def _wait_pbm_status(self) -> None:
                     component_name=self.name,
                 )
                 raise ResyncError
+
+            # We're done with the sync, let's clear the statuses.
+            self.state.statuses.clear(scope="unit", component=self.name)
         except WorkloadExecError as e:
             if status := self.process_pbm_error_as_status(e.stdout):
                 self.state.statuses.add(status, scope="unit", component=self.name)
+                return
```
Mehdi-Bendriss, line 810: Should we not raise if no status?

tests/integration/mongodb/backups/test_backups.py
```diff
@@ -102,31 +107,15 @@ async def test_blocked_incorrect_creds(ops_test: OpsTest, substrate: Substrate)
     )
 
 
-@pytest.mark.abort_on_fail
-async def test_blocked_incorrect_conf(
```
Mehdi-Bendriss, line 106: why is this test no longer relevant?

single_kernel_mongo/managers/backups.py
```diff
@@ -141,6 +153,92 @@ def backup_id(self) -> str:
...
+            region_name=credentials.get("region"),
+            endpoint_url=credentials["endpoint"],
+            aws_access_key_id=credentials["access-key"],
+            aws_secret_access_key=credentials["secret-key"],
+            config=BotoConfig(
+                # https://github.com/boto/boto3/issues/4400#issuecomment-2600742103
+                request_checksum_calculation="when_required",
+                response_checksum_validation="when_required",
+            ),
+            verify=(TRUST_STORE_PATH / TrustStoreFiles.PBM.value)
```
Mehdi-Bendriss, line 173: Could we replace the value of the `TRUST_STORE_PATH` with a localized path? (withing `SNAP_DATA`)

Mehdi-Bendriss (commented):
Thanks Neha! Great work! I left a few comments

single_kernel_mongo/events/backups.py
Gu1nness: So more statuses ? Or we let the charm error, let the user fix and then resolve ?

single_kernel_mongo/managers/backups.py
Gu1nness: On VM that path is already used to store the SSL certificate, this is specifically important for k8s where we have to land the certificate on the charm container.
I can move it to another location but I'm not sure where and if that would make sense or not?

single_kernel_mongo/managers/backups.py
Gu1nness: I forgot to remove this one, in the first code I wrote it was not checking for the bucket existence before trying to create it.
Let me remove those checks.

single_kernel_mongo/managers/backups.py
Gu1nness: We rely on the endpoint in the bucket creation even for AWS so I'd rather make it mandatory (and have the same level of checks as etcd charm or postgres charm (for example) than having an exception for AWS. I don't now why it was made optional for AWS while PBM does not make it optional for AWS.

tests/integration/mongodb/backups/test_backups.py
Gu1nness: This test was designed to check if we succesfully ended up in the "s3 configurations are incompatible", which was raised if the bucket was not existing (because we were under the assumption that the operator had to provide a working bucket).
Since we now create the bucket before anything and will not proceed further if the bucket is not created, it's not needed anymore.

single_kernel_mongo/events/backups.py
Mehdi-Bendriss: I think we should log those and treat them as `FAILED_TO_CREATE_BUCKET` exceptions

single_kernel_mongo/managers/backups.py
Mehdi-Bendriss: for k8s it should be fine, however the path on VM is still `/usr/local/share/ca-certificates` and not relative to `SNAP_DATA`

Gu1nness: Yes but we have not decided on an implementation for those paths and this addition would imply ours of development and new snap + rock releases to add the correct environment variables.
When I say "On VM that path is already used" it means that this is the path that we're currently using when storing that exact certificate for PBM anyway.
I had asked if there would be a need for such a spec/implementation for certificates in a Data Office Hour and the answer seemed to be that there was no need for that from mykola.

single_kernel_mongo/events/backups.py
Gu1nness: Ok  !

single_kernel_mongo/managers/backups.py
Mehdi-Bendriss: The issue with `/usr/local/share/ca-certificates` is that the new certificate or CA is trusted system wide, not only by mongodb. This should be our last recourse only IF pbm doesn't accept custom trust stores.

I am fine moving forward as-is for the time being, but we need to investigate more profoundly how to make PBM trust certificates located in a non-system wide trust store.

Let's track it [here](https://github.com/canonical/mongo-single-kernel-library/issues/54)

Gu1nness: PBM use the system certificates.
So to use a "local trust store" for the snap would mean:
 * Update snap to create `/var/snap/charmed-mongodb/current/etc/ssl/certs` + prepopulate that directory.
 * In the snap, add an environment variable to PBM and PBM agent that specify to look for certificates in `/var/snap/charmed-mongodb/current/etc/ssl/certs`.
 * Update the calls to `/update-ca-certificates` to add `--etccertsdir /var/snap/charmed-mongodb/current/etc/ssl/certs`
 * Optionally also create `/var/snap/charmed-mongodb/current/usr/local/share/ca-certificates` and same add the right argument to the `update-ca-certificates` call

Mehdi-Bendriss: This would be the way, we have introduced a [spec](https://docs.google.com/document/d/1dvy2sEkCliSFLAXKBkGiyMuJCE6wiO794N3HkaDNGfE/edit?tab=t.0) for jvm based databases on the same topic

Mehdi-Bendriss (dismissed):
Thanks Neha!

single_kernel_mongo/config/statuses.py
```diff
@@ -116,13 +116,29 @@ class BackupStatuses(Enum):
...
+        status="blocked",
+        message="Unknown PBM error, check logs.",
+        action="Check logs for more information",
     )
-    UNKNOWN_PBM_ERROR = StatusObject(status="blocked", message="Unknown PBM error, check logs.")
     CANT_CONFIGURE = StatusObject(status="blocked", message="Couldn't configure s3 backup options.")
     ACTIVE_IDLE = StatusObject(status="active", message="")
+    FAILED_TO_CREATE_BUCKET = StatusObject(
+        status="blocked",
+        message="Failed to create buckets, check logs.",
```
Mehdi-Bendriss:
```suggestion
        message="Failed to create S3 bucket, check logs.",
```

single_kernel_mongo/events/backups.py
```diff
@@ -119,14 +122,43 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent) -> None:  #
...
+            logger.error("Failed to create bucket: %s", err)
+            self.manager.state.statuses.add(
+                BackupStatuses.FAILED_TO_CREATE_BUCKET.value,
+                scope="unit",
+                component=self.manager.name,
+            )
+            event.defer()
+            return
         except SetPBMConfigError:
-            logger.error("Failed to configure s3 backup options")
```
Mehdi-Bendriss, line 129: why not keep?

single_kernel_mongo/managers/backups.py
```diff
@@ -141,6 +153,89 @@ def backup_id(self) -> str:
...
+            with open(TRUST_STORE_PATH / TrustStoreFiles.PBM.value, mode="w") as fd:
+                # boto3 client will need the certificate on the node that runs the command
+                fd.write("\n".join(tls_ca_chain))
+
+        bucket = self._get_bucket_resource(credentials)
+
+        try:
+            bucket.meta.client.head_bucket(Bucket=bucket_name)
+            logger.info(f"Using existing bucket {bucket_name}")
+            exists = True
```
Mehdi-Bendriss: this is not needed, we can return here, and omit the need for this variable altogether

single_kernel_mongo/managers/backups.py
```diff
@@ -141,6 +153,89 @@ def backup_id(self) -> str:
...
+            raise e
+        except ClientError:
+            logger.warning("Bucket %s doesn't exist or you don't have access to it.", bucket_name)
+            exists = False
+        except SSLError as e:
+            logger.error(f"error: {e!s} - Is TLS enabled and CA chain set on S3?")
+            raise e
+
+        if exists:
+            return
```
Mehdi-Bendriss: related to the above comment

single_kernel_mongo/managers/backups.py
```diff
@@ -388,7 +499,7 @@ def resync_config_options(self):  # pragma: nocover
                     case BackupState.BACKUP_RUNNING | BackupState.RESTORE_RUNNING:
                         raise PBMBusyError
                     case BackupState.WAITING_TO_SYNC:
-                        self.workload.restart()
+                        # self.workload.restart()
```
Mehdi-Bendriss: Can we delete?

single_kernel_mongo/managers/mongo.py
```diff
@@ -274,18 +274,22 @@ def add_user(self, relation: Relation):
...
             self.model,
             relation.name,
         )
 
         username = f"relation-{relation.id}"
         password = data_interface.fetch_my_relation_field(relation.id, "password")
+
         with MongoConnection(self.state.mongo_config) as mongo:
             has_user = mongo.user_exists(username)
-        if has_user:
```
Mehdi-Bendriss, line 287: This was a funny bug, glad you found it

Mehdi-Bendriss (commented):
Thank you Neha! Really good job here! 
I only have a few minor comments and we should be good

single_kernel_mongo/events/backups.py
Gu1nness: It's already logged when the error is raised with more information. Prevent duplication of logs.

single_kernel_mongo/managers/mongo.py
Gu1nness: This has been raised by Mateo !
He realised he was lacking the access to his user but it had the accesses to the default database.
But this is good that it's fixed :)

Mehdi-Bendriss (approved):

PR #62, opened 2025-08-22
patch: extend status messages for MongoDBStatuses

patriciareinoso:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

- Extend the status in `MongoDBStatuses`.
  - Make the `short_message` shorter than 40 characters.
  - Add `check` and `action` when judged necessary
  - Make the variable names, messages and vocabulary consistent
- Remove duplicated check for sharding relation on replica sets

## 📑 Motivation and Context

This PR improves the quality of the message displayed to the operator about the state of the charms

Add some UTs and tested it locally

## ✅ Checklist

image

### DIFF
--- a/single_kernel_mongo/config/statuses.py
+++ b/single_kernel_mongo/config/statuses.py
@@ -31,31 +31,31 @@ class MongoDBStatuses(Enum):
         message="Waiting for mongodb-exporter to start...",
         check="MongoDB Exporter status check.",
     )
-    INCOMPATIBLE_SHARDING_REL = StatusObject(
+    INVALID_SHARDING_REL = StatusObject(
         status="blocked",
         message="The sharding interface cannot be used with replica sets.",
-        short_message="Incompatible sharding relation.",
+        short_message="Invalid sharding relation.",
         check="Relation validation.",
         action="Remove the relation on the shards interface (config-server or sharding relation) from this application.",
     )
-    INCOMPATIBLE_MONGOS_REL = StatusObject(
+    INVALID_MONGOS_REL = StatusObject(
         status="blocked",
         message="The cluster relation is can only be used with config servers.",
-        short_message="Incompatible cluster relation.",
+        short_message="Invalid cluster relation.",
         check="Relation validation.",
         action="Remove the cluster relation (config-server interface) from this application.",
     )
-    INCOMPATIBLE_S3_REL = StatusObject(
+    INVALID_S3_REL = StatusObject(
         status="blocked",
         message="The s3-credentials relation cannot be used with config servers and replica sets.",
-        short_message="Incompatible s3-credentials relation.",
+        short_message="Invalid s3-credentials relation.",
         check="Relation validation.",
         action="Remove the s3-credentials relation (s3 interface) from this application.",
     )
-    INCOMPATIBLE_DB_REL = StatusObject(
+    INVALID_DB_REL = StatusObject(
         status="blocked",
         message="The database relation cannot be used with sharding components (shards or config servers).",
-        short_message="Incompatible database relation.",
+        short_message="Invalid database relation.",
         check="Relation validation.",
         action="Remove the database relation (mongodb_client interface) from this application.",
     )
--- a/single_kernel_mongo/config/statuses.py
+++ b/single_kernel_mongo/config/statuses.py
@@ -21,24 +21,43 @@ class MongoDBStatuses(Enum):
     """MongoDB related statuses."""
 
     # STATE statuses:
-    WAITING_FOR_MONGODB_START = StatusObject(status="waiting", message="Waiting to start mongod...")
+    WAITING_FOR_MONGODB_START = StatusObject(
+        status="waiting",
+        message="Waiting for mongod to start...",
+        check="MongoDB process status check.",
+    )
     WAITING_FOR_EXPORTER_START = StatusObject(
-        status="waiting", message="Waiting to start mongodb-exporter..."
+        status="waiting",
+        message="Waiting for mongodb-exporter to start...",
+        check="MongoDB Exporter status check.",
     )
-    SHARDING_ON_REPLICA = StatusObject(
-        status="blocked", message="Sharding interface cannot be used by replicas."
-    )
-    UNSUPPORTED_MONGOS_REL = StatusObject(
+    INCOMPATIBLE_SHARDING_REL = StatusObject(
         status="blocked",
-        message="Relation to mongos not supported, config role must be config-server.",
+        message="The sharding interface cannot be used with replica sets.",
+        short_message="Incompatible sharding relation.",
+        check="Relation validation.",
+        action="Remove the relation on the shards interface (config-server or sharding relation) from this application.",
     )
-    INVALID_S3_INTEGRATION_STATUS = StatusObject(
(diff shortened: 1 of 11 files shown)


single_kernel_mongo/core/structured_config.py
```diff
@@ -121,7 +121,7 @@ class MongoDBCharmConfig(MongoConfigModel):
     @field_validator("role", mode="before")
     @classmethod
     def invalid_role_to_unknown(cls, v: str) -> MongoDBRoles:
-        """If the value is neither none or nodeport, returns unknown."""
+        """If the value is neither replication, config-server, or shard, returns unknown."""
```
Gu1nness, line 124: :purple_heart:

single_kernel_mongo/config/statuses.py
```diff
@@ -21,24 +21,43 @@ class MongoDBStatuses(Enum):
     """MongoDB related statuses."""
 
     # STATE statuses:
-    WAITING_FOR_MONGODB_START = StatusObject(status="waiting", message="Waiting to start mongod...")
+    WAITING_FOR_MONGODB_START = StatusObject(
+        status="waiting",
+        message="Waiting for mongod to start...",
```
Gu1nness, line 26: We'll have to check for all statuses changes if it needs a documentation update or not, and they should be validated by @Mehdi-Bendriss

single_kernel_mongo/managers/backups.py
```diff
@@ -246,7 +246,11 @@ def is_valid_s3_integration(self) -> bool:
 
         Only replica sets and config_servers can integrate to s3-integrator.
         """
-        return (self.state.s3_relation is None) or (not self.state.is_role(MongoDBRoles.SHARD))
+        return (
+            self.state.s3_relation is None
+            or self.state.is_role(MongoDBRoles.REPLICATION)
+            or self.state.is_role(MongoDBRoles.CONFIG_SERVER)
+        )
```
Gu1nness, line 253: Good sight!

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -913,17 +913,17 @@ def is_relation_feasible(self, rel_name: str) -> bool:
             )
 
             self.state.statuses.add(
-                MongoDBStatuses.INVALID_DB_REL_ON_SHARD.value, scope="unit", component=self.name
+                MongoDBStatuses.INVALID_DB_REL.value, scope="unit", component=self.name
             )
             return False
-        if not self.state.is_sharding_component and rel_name == RelationNames.SHARDING:
+        if not self.state.is_sharding_component and self.state.has_sharding_integration:
```
Gu1nness: I don't think this change is semantically correct: the `is_relation_feasible` is called on a specific relation (eg as part of the hook checks before proceeding) while this change will add a status that could be unrelated to that specific relation.

There are two things here: On one side the check from the `get_statuses` that is overall, and on this side this status that is returned for another check.
Maybe the best would be to add one more status here, or on the `get_statuses` to iterate on relations.

I'm not sure which one is the best, but I'm sure that this check shouldn't change because the function checks for a specific relation.

Gu1nness (changes requested):
Excellent work!
There's a discussion open on the change on `is_relation_feasible` that needs a decision, but aside from that, I love what I'm reading!

single_kernel_mongo/config/statuses.py
Mehdi-Bendriss: I think this form is better phrased than the previous one, though we'll have to reflect that on the charmhub docs where it applies

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: You are right. I modified the `is_relation_feasible` method to check against `rel_name`. As discussed in a meeting, there is some rework to do over the `get_statuses` and `is_relation_feasible` methods to reduce duplication. This is to be done in a separate PR.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -913,17 +913,19 @@ def is_relation_feasible(self, rel_name: str) -> bool:
...
 
             self.state.statuses.add(
-                MongoDBStatuses.INVALID_DB_REL_ON_SHARD.value, scope="unit", component=self.name
+                MongoDBStatuses.INVALID_DB_REL.value, scope="unit", component=self.name
             )
             return False
-        if not self.state.is_sharding_component and rel_name == RelationNames.SHARDING:
+        if not self.state.is_sharding_component and (
+            rel_name == RelationNames.SHARDING or rel_name == RelationNames.CONFIG_SERVER
+        ):
```
Mehdi-Bendriss: nit:
```suggestion
        if not self.state.is_sharding_component and rel_name in {RelationNames.SHARDING, RelationNames.CONFIG_SERVER}
        ):
```

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -993,21 +995,16 @@ def basic_statuses(self) -> list[StatusObject]:
...
-            self.state.config_server_relation or self.state.shard_relation
-        ):
-            statuses.append(MongoDBStatuses.SHARDING_ON_REPLICA.value)
+            statuses.append(MongoDBStatuses.INVALID_S3_REL.value)
 
         if self.state.client_relations and self.state.is_sharding_component:
-            statuses.append(MongoDBStatuses.INVALID_DB_REL_ON_SHARD.value)
+            statuses.append(MongoDBStatuses.INVALID_DB_REL.value)
 
         if not self.state.is_sharding_component and self.state.has_sharding_integration:
```
Mehdi-Bendriss, line 1007: Is this the equivalent of the now deleted (a few lines earlier)?
```
        if self.state.is_role(MongoDBRoles.REPLICATION) and (
            self.state.config_server_relation or self.state.shard_relation
        )
```

Mehdi-Bendriss (commented):
Thanks Patricia, great work! I only have a question but otherwise LGTM

tests/integration/mongodb/relations/test_sharding_relations.py
```diff
@@ -324,14 +324,14 @@ async def test_shard_s3_relation(ops_test: OpsTest, substrate: Substrate) -> Non
         ops_test,
         substrate,
         SHARD_ONE_APP_NAME,
-        status="Relation to s3-integrator is not support",
+        status="Invalid s3-credentials relation",
```
Gu1nness: Nit: 
```suggestion
        status="Invalid s3-credentials relation.",
```

Gu1nness (dismissed):

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: yes, both conditions are equivalent

single_kernel_mongo/config/statuses.py
patriciareinoso: I will update the documentation to match the changes

Mehdi-Bendriss (approved):
Thanks Patricia, great work!

Gu1nness (approved):

PR #63, opened 2025-08-26
patch: extend statuses messages for mongos, charm, tls, and backup components

patriciareinoso:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

- Extend the status in `MongosStatuses`, `CharmStatuses`, `TLSStatuses` and `BackupStatuses` classes
  - Make the short_message shorter than 40 characters.
  - Add check and action when judged necessary
  - Make the variable names, messages and vocabulary consistent

Functional changes:
- The `TLSStatuses.ENABLING_TLS` becomes `async` as indicated in the existing comment above its declaration. 
- Improve the assert message in the blocked status message check for units

## 📑 Motivation and Context

This PR improves the quality of the message displayed to the operator about the state of the charms

Locally and in integration tests

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/config/statuses.py
+++ b/single_kernel_mongo/config/statuses.py
@@ -50,27 +50,58 @@ class MongoDBStatuses(Enum):
 class MongosStatuses(Enum):
     """Mongos related statuses."""
 
-    INVALID_EXPOSE_EXTERNAL = StatusObject(
-        status="blocked", message="Config option for expose-external not valid."
-    )
     CONNECTING_TO_CONFIG_SERVER = StatusObject(
-        status="waiting", message="Connecting to config-server..."
+        status="waiting",
+        message="Connecting to config-server...",
+        check="mongos process status check.",
     )
     WAITING_FOR_SECRETS = StatusObject(
-        status="waiting", message="Waiting for secrets from config-server"
+        status="waiting",
+        message="Waiting for config-server secrets...",
+        check="Cluster relation validation.",
     )
-    REQUIRES_TLS = StatusObject(status="blocked", message="mongos requires TLS to be enabled.")
-    REQUIRES_NO_TLS = StatusObject(
-        status="blocked", message="mongos has TLS enabled, but config-server does not."
+    WAITING_FOR_MONGOS_START = StatusObject(
+        status="waiting",
+        message="Waiting for mongos to start...",
+        check="mongos process status check.",
+    )
+    INVALID_EXPOSE_EXTERNAL = StatusObject(
+        status="blocked",
+        message="The expose-external config option is invalid. Valid options are `nodeport` and `none`.",
+        short_message="Invalid expose-external config",
+        check="Config validation failed.",
+        action="Set the expose-external config to a valid value: `nodeport` or `none`.",
+    )
+    MISSING_TLS_REL = StatusObject(
+        status="blocked",
+        message="TLS must be enabled in mongos, since it is enabled on the config-server in the cluster relation.",
+        short_message="Certificates relation is missing",
+        check="Relation validation failed.",
+        action="Add the certificates relation (tls-certificates interface) to mongos.",
+    )
+    INVALID_TLS_REL = StatusObject(
+        status="blocked",
+        message="TLS must be disabled in mongos, since it is disabled on the config-server in the cluster relation.",
+        short_message="Invalid certificates relation",
+        check="Relation validation failed.",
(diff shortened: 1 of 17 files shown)


delgod (dismissed):

single_kernel_mongo/config/statuses.py
```diff
@@ -69,27 +69,58 @@ class MongoDBStatuses(Enum):
...
-    REQUIRES_NO_TLS = StatusObject(
-        status="blocked", message="mongos has TLS enabled, but config-server does not."
+    WAITING_FOR_MONGOS_START = StatusObject(
+        status="waiting",
+        message="Waiting for mongos to start...",
+        check="mongos process status check.",
+    )
+    INVALID_EXPOSE_EXTERNAL = StatusObject(
+        status="blocked",
+        message="The expose-external config option is invalid. Valid options are `nodeport` and `none`.",
```
Gu1nness, line 89: I would probably remove the second part of the message if we have the action field explaining what should be done to fix this.

single_kernel_mongo/config/statuses.py
```diff
@@ -100,68 +131,94 @@ class CharmStatuses(Enum):
     """Charm Statuses."""
 
     ACTIVE_IDLE = StatusObject(status="active", message="")
-    FAILED_SERVICES_START = StatusObject(status="blocked", message="Failed to start services.")
-    WAITING_TO_START = StatusObject(status="waiting", message="Starting services.")
-    MONGODB_NOT_INSTALLED = StatusObject(status="blocked", message="MongoDB not installed.")
-    MONGOS_NOT_STARTED = StatusObject(status="waiting", message="Waiting to start mongos...")
+    FAILED_SERVICES_START = StatusObject(
+        status="maintenance", message="Failed to start services. Retrying..."
+    )
```
Gu1nness: I'm wondering if that status should be maintenance or blocked.
On one hand, looks like maintenance, but on the other hand it could be failing because the service is misconfigured and will never start, and in that case maybe it should be blocked?
@Mehdi-Bendriss ?

single_kernel_mongo/config/statuses.py
```diff
@@ -100,68 +131,94 @@ class CharmStatuses(Enum):
...
     # RUNNING Statuses
     INSTALLING_MONGODB = StatusObject(
-        status="maintenance", message="installing MongoDB", running="blocking"
+        status="maintenance", message="Installing MongoDB...", running="blocking"
     )
     DEPLOYED_WITHOUT_TRUST = StatusObject(
-        status="blocked", message="Charm deployed without `trust`", running="async"
+        status="blocked",
+        message="Charm deployed without `trust` option.",
+        action="Run `juju trust <application-name>`.",
```
Gu1nness:
```suggestion
        action="Run `juju trust --scope=cluster <application-name>`.",
```

Gu1nness (changes requested):
A few tiny suggestions / discussions.
One change to be made on the `juju trust` action.

single_kernel_mongo/config/statuses.py
Mehdi-Bendriss: I think we can get away with the Blocked here, for simple cases / regular retries, the `Retrying...` should be indication enough it's temporary.

single_kernel_mongo/config/statuses.py
Mehdi-Bendriss: This is an interesting one, because on one hand it would be useful to show the whole message if the space allows, as the user won't be forced to run the action. 
And in the case where the short message has been set, the action would yield the same action items. 
No duplication here imo

single_kernel_mongo/state/cluster_state.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, relation: Relation | None, data_interface: Data, component: A
 
     @property
     def config_server_uri(self) -> str:
-        """Is TLS enabled."""
+        """Return config-server URI in the databag."""
```
Mehdi-Bendriss, line 41: nice catch

Mehdi-Bendriss (dismissed):
Thanks Patricia!

single_kernel_mongo/config/statuses.py
patriciareinoso: I changed to `blocked` and added an action indicating to check the logs. WDYT?

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #67, opened 2025-08-29
patch: remove auto-delete option

patriciareinoso:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

This PR removes the `auto-delete` config option and its logic. 
Now, the database are never dropped.

## 📑 Motivation and Context

This is not a desired configuration option in release 8.
This prevents unintended data loss in case of misconfiguration.

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/utils/mongo_connection.py
+++ b/single_kernel_mongo/utils/mongo_connection.py
@@ -31,7 +31,7 @@ from single_kernel_mongo.exceptions import (
 from single_kernel_mongo.utils.helpers import hostname_from_hostport, hostname_from_shardname
 from single_kernel_mongo.utils.mongo_config import MongoConfiguration
 from single_kernel_mongo.utils.mongo_error_codes import MongoErrorCodes
-from single_kernel_mongo.utils.mongodb_users import DBPrivilege
+from single_kernel_mongo.utils.mongodb_users import DBPrivilege, SystemDBS
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


single_kernel_mongo/utils/mongo_connection.py
```diff
@@ -319,18 +319,6 @@ def add_replset_member(self, hostname: str) -> None:
...
-        """Return list of all non-default databases."""
-        databases: list[str] = self.client.list_database_names()
-        return {db for db in databases if db not in SystemDBS}
-
-    def drop_database(self, database: str):
-        """Drop a non-default database."""
-        if database in SystemDBS:
-            logger.info(f"Not dropping system DB {database}.")
-            return
-        self.client.drop_database(database)
```
Gu1nness, line 332: Even if they are unused for now, I think it's safe to keep those "just in case" we would need it for the future?

Gu1nness (commented):
I would keep the two methods in `mongo_connection.py`
With regard to the PR name, we have a "system" around PR names that should definitely end up in the developer documentation.
Basically, `1.6.x` is for `6/edge` branch, `1.7.x` is for `8-transition/edge` and `1.8.x` is for `8/edge` branch.
So we only ever use `patch` PR's and `compatible` for new major version of MongoDB.
It's not ideal but it is how we decided a few weeks ago to handle the link between versions and `tracks`

delgod (dismissed):

single_kernel_mongo/utils/mongo_connection.py
Mehdi-Bendriss: I think I would disagree here, I think it makes sense we remove unused code and de-clutter as much as we can. Those methods seem to be simple enough to rewrite should the need ever arises

Mehdi-Bendriss (commented):
Thanks Patricia! This PR is pending the merge of the unit tests port PR - I would also be in favor of deleting unused code

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #83, opened 2025-09-19
patch: migrate from action-based password management to config-option-based passwords

patriciareinoso:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

In `Mongo6` we used to have password management trough actions: `get-password` and `set_password`.

This PR replaces the the action approach with a secret config-option approach. 

How it works:
- `get-password` and `set-password` are removed
- The new config option is called `system-users`
- Only the leader unit of a config-server or a replica set can manage passwords.
- If the config option is set in a shard, a blocked status is added.
- Password update is triggered on `config-changed` or `secret-update` event.
- The user, may set a password for one or multiple users in the secret.
- The password update is deferred if there is an upgrade, a backup in progress or if the secret has not been granted to the application.
- Statuses related to the password management are set at application level.
- The `PasswordManagementContext` defines the a `PasswordManagementState` and extracts the content of the `system-users` secret. This is useful to ensure that the same checks are performed independently of the situation and we can trust that the system_users field in the PasswordManagementContext is always valid if it exists.

## 📑 Motivation and Context

Align Data products for internal users password management.

Locally and existing integration tests. UTs where added.

## ✅ Checklist


tests/charms/mongodb_k8s_test_charm/config.yaml
```diff
@@ -26,3 +26,11 @@ options:
       `{PROVIDED_USER}` should be used only if no value in the `ldap-user-to-do-mapping` config option is provided.
       If this configuration is not provided, a default string will be computed based on the base_dn returned by the GLAuth k8s charm.
       example: “dc=example,dc=com??sub?(&(objectClass=groupOfNames)(member={USER}))"
+  system-users:
```
patriciareinoso, line 29: add that this option is only compatible with replication and config server role

single_kernel_mongo/config/statuses.py
```diff
@@ -93,6 +100,18 @@ class CharmStatuses(Enum):
...
+    INVALID_SYSTEM_USERS = StatusObject(
+        status="blocked",
+        message="Invalid secret in system-users config.",
+        action="Check logs and verify the content and permissions of the secret in the system-users config.",
+        check="Configuration update failure.",
+        running="async",
+    )
+    PASSWORD_UPDATE_FAILED = StatusObject(
+        status="maintenance",
+        message="Failed to update user passwords.",
```
Gu1nness:
```suggestion
        message="Failed to update user passwords.",
        action="Check logs."
```

single_kernel_mongo/events/lifecycle.py
```diff
@@ -161,7 +164,12 @@ def on_config_changed(self, event: ConfigChangedEvent):
         """Config Changed Event."""
         try:
             self.dependent.update_config_and_restart()
-        except (UpgradeInProgressError, WaitingForLeaderError):
+        except (
+            UpgradeInProgressError,
+            WaitingForLeaderError,
+            DeferrableFailedHookChecksError,
```
Gu1nness, line 171: :purple_heart:

single_kernel_mongo/events/lifecycle.py
```diff
@@ -161,7 +164,12 @@ def on_config_changed(self, event: ConfigChangedEvent):
...
+            SetPasswordError,
+        ):
             event.defer()
```
Gu1nness: Realising now, we should log a line here to explain that we're deferring and for what reason (something like [that](https://github.com/canonical/mongo-single-kernel-library/blob/6/edge/single_kernel_mongo/events/sharding.py#L154).

You can use the helper method `defer_event_with_info_log` that you can find in `single_kernel_mongo.utils.event_helpers` that will log a formatted line and defer.

single_kernel_mongo/events/lifecycle.py
```diff
@@ -175,6 +183,8 @@ def on_config_changed(self, event: ConfigChangedEvent):
                 scope="unit",
                 component=self.dependent.name,
             )
+        except (NonDeferrableFailedHookChecksError, WorkloadServiceError):
```
Gu1nness, line 187: Same, better be verbose in logging than leaving an empty return.
If not, the user won't know why the event did not end up in what they where expecting.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -441,9 +443,74 @@ def update_config_and_restart(self) -> None:
...
+            self.charm.status_handler.set_running_status(
+                CharmStatuses.INVALID_SYSTEM_USERS.value,
+                scope="app",
+                statuses_state=self.state.statuses,
+                component_name=self.name,
+            )
+            raise SetPasswordError("The charm has no permission to access the secret.")
+        except (SecretNotFoundError, InvalidPasswordError) as e:
+            logger.error(f"Invalid system-users secret: {e}.")
+            self.charm.status_handler.set_running_status(
```
Gu1nness: I think this does not need to be a running status, but a regular status: you will exit right after, there's no need to display it immediately, and it's easy to recompute it later.
You'll still be able to remove it as you do below, with the same call.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -441,9 +443,74 @@ def update_config_and_restart(self) -> None:
...
+                component=self.name,
+            )
+            return
+        try:
+            user_passwords = self.charm.state.get_secret_from_id(system_users_secret_id)
+            validate_charm_user_password_config(user_passwords)
+        except ModelError as e:
+            logger.error(f"Invalid system-users secret: {e}.")
+            self.charm.status_handler.set_running_status(
+                CharmStatuses.INVALID_SYSTEM_USERS.value,
```
Gu1nness: I think this does not need to be a running status, but a regular status: you will exit right after, there's no need to display it immediately, and it's easy to recompute it later.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -468,8 +535,23 @@ def new_leader(self) -> None:
...
+        if self.state.internal_user_passwords_are_initialized():
+            return
+
+        if system_users_secret_id := self.config.system_users:
+            user_passwords = None
+            try:
+                user_passwords = self.charm.state.get_secret_from_id(system_users_secret_id)
+                validate_charm_user_password_config(user_passwords)
+            except (ModelError, SecretNotFoundError, InvalidPasswordError) as e:
+                logger.error(f"Invalid system-users secret: {e}.")
```
Gu1nness: I feel this should lead to the same kind of message as on config changed: similar problems lead to similar statuses.
It could hence be exported to a helper function that could be used as well in the `get_statuses` method to compute the same statuses on status update events.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -468,8 +535,23 @@ def new_leader(self) -> None:
...
+
+            if user_passwords:
+                for username, password in user_passwords.items():
+                    user = get_user_from_username(username)
+                    self.state.set_user_password(user, password)
+
```
Gu1nness: I would rework this in a less intricate and more natural way (which avoids back and forth):
```
user_passwords : dict[MongoDBUser, str] = {}
if system_users_secret_id := ...:
   // Same logic but store user, password in the dictionary

for user in Charm users:
    if not self.state.get_user_password(user):
		password = user_passwords.get(user, self.workload.generate_password())
        self.state.set_user_password(user, password)
```

This way:
 * You have less nested loops
 * You never iterate twice on the same user
 * You avoid having set password and then get password on the same user in the second loop
 * You never have to worry if `user_passwords is empty or not, they'll all be treated the same way once the map is built.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -685,18 +777,10 @@ def update_status(self) -> None:
             except NotDrainedError:
                 logger.warning("Still draining shard.")
 
-    def set_password(self, username: str, password: str | None = None) -> tuple[str, str]:
-        """Handler for the set password action."""
+    def set_charmed_user_password(self, user: MongoDBUser, new_password: str) -> None:
+        """Handler for the set password."""
         self.assert_pass_password_checks()
```
Gu1nness: This check should be moved way higher in the call stack I think.
It used to be the safety check before settting a password in the `set-password` action. So I would call it in the config-changed / leader-elected events  maybe?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -723,7 +805,7 @@ def assert_pass_password_checks(self) -> None:
                 "Password rotation must be called on leader unit."
             )
         if self.state.upgrade_in_progress:
-            raise NonDeferrableFailedHookChecksError(
+            raise DeferrableFailedHookChecksError(
```
Gu1nness, line 815: Excellent!

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1047,3 +1124,19 @@ def get_statuses(self, scope: DPHScope, recompute: bool = False) -> list[StatusO
...
+            statuses.append(CharmStatuses.PASSWORD_ON_SHARD.value)
+            return statuses
+
+        if system_users_secret_id := self.config.system_users:
+            try:
+                user_passwords = self.charm.state.get_secret_from_id(system_users_secret_id)
+                validate_charm_user_password_config(user_passwords)
+            except (ModelError, SecretNotFoundError, InvalidPasswordError) as e:
+                logger.error(f"Invalid system-users secret: {e}.")
+                statuses.append(CharmStatuses.INVALID_SYSTEM_USERS.value)
```
Gu1nness: As I wrote above:
 * Same errors should lead to same statuses, so you need to align between `leader-elected`, `config-changed` and this piece of code.
 * I have the feeling that this code duplication highlights an issue in the implementation: I would advise you to have a look at other managers where we were facing the same kind of issues, such as the `BackupManager` or `LdapManager`: there, we build a `state` object based on the check, that is then mapped according to the function to state value: set a status + raise / set a status or map directly the state to a status.

This will bring a much more cleaner code and an easier one to understand.

single_kernel_mongo/state/charm_state.py
```diff
@@ -870,3 +875,21 @@ def mongo_config(self) -> MongoConfiguration:
...
+            dict: The content of the secret.
+        """
+        try:
+            secret_content = self.charm.model.get_secret(id=secret_id).get_content(refresh=True)
+        except SecretNotFoundError:
+            raise SecretNotFoundError(f"The secret '{secret_id}' does not exist.")
+        except ModelError:
+            raise
+
+        return secret_content
```
Gu1nness, line 897: Nit: already map each username to its `MongoDBUser` here and discard all other values in the secret from the map. That way you know this function only returns valid users.

single_kernel_mongo/utils/mongodb_users.py
```diff
@@ -171,11 +172,18 @@ def get_hosts(self) -> set[str]:
...
     MonitorUser.username,
     LogRotateUser.username,
+}
+
+CharmUsers = (
+    OperatorUser,
+    BackupUser,
+    MonitorUser,
+    LogRotateUser,
 )
```
Gu1nness: I feel we're duplicating `InternalUsers` here with `CharmUsernames`.
Maybe keep only one (I would go for internal users).

single_kernel_mongo/utils/mongodb_users.py
```diff
@@ -190,3 +198,30 @@ def get_user_from_username(username: str) -> MongoDBUser:
...
+      - Only internal charmed usernames are allowed.
+      - Passwords must not be empty or only whitespace.
+      - Passwords must not exceed MAX_PASSWORD_LENGTH characters.
+
+    Raises:
+        InvalidPasswordError: if any validation rule is violated.
+    """
+    extra_users = set(user_passwords.keys()) - CharmUsernames
+    if extra_users:
+        raise InvalidPasswordError(f"Unexpected usernames provided: {', '.join(extra_users)}")
```
Gu1nness: I'm not sure of the specification but my understanding was that we should support a secret with extra data that we just ignore, not raise if there's any unexpected field.
@reneradoi Could you confirm/infirm that assertion?

tests/charms/mongodb_k8s_test_charm/config.yaml
```diff
@@ -26,3 +26,11 @@ options:
...
+    type: secret
+    description: |
+      Configure the internal system users and their passwords. The passwords will
+      be auto-generated if this option is not set. It is for internal use only
+      and SHOULD NOT be used by applications. This needs to be a Juju Secret URI pointing
+      to a secret that contains the following content: `<username>: <password>`.
```
Gu1nness:
```suggestion
      to a secret that contains a map with the following content: `<username>: <password>`.
```

tests/charms/mongodb_k8s_test_charm/config.yaml
```diff
@@ -26,3 +26,11 @@ options:
       `{PROVIDED_USER}` should be used only if no value in the `ldap-user-to-do-mapping` config option is provided.
       If this configuration is not provided, a default string will be computed based on the base_dn returned by the GLAuth k8s charm.
       example: “dc=example,dc=com??sub?(&(objectClass=groupOfNames)(member={USER}))"
+  system-users:
+    type: secret
+    description: |
+      Configure the internal system users and their passwords. The passwords will
+      be auto-generated if this option is not set. It is for internal use only
+      and SHOULD NOT be used by applications. This needs to be a Juju Secret URI pointing
```
Gu1nness:
```suggestion
      and SHOULD NOT be used by applications. This MUST be a Juju Secret URI pointing
```

On top of that, and it goes also with the `structured_config.py` validation, do we know a precise structure for the secret so we can do some validation on it ?

tests/integration/mongodb/test_charm.py
```diff
@@ -205,119 +207,118 @@ async def test_set_password_action(ops_test: OpsTest, substrate: Substrate) -> N
...
+    # The password has to be maximum 4096-character long
+    await set_password(
+        ops_test,
+        username=MONITOR_USERNAME,
+        password="c" * 4097,
+        app_name=app_name,
+    )
+    app = ops_test.model.applications[app_name]
+    await ops_test.model.block_until(*[lambda: app.status == "blocked"], timeout=1000)
+    assert app.status_message == "Invalid secret in system-users config."
```
Gu1nness: If we're going to repeat this more than 3 times, create a helper in `common.py` helpers file and use it anytime you want to block until a certain status.

Gu1nness (changes requested):
That is really excellent work!
 I have a few points of attention on code duplication and readability, so we adhere to the best standards.
 Aside from that, I love the implementation, it's overall very sturdy and very well tested!

Good job, let's just discuss the few points I have opened and find a common ground on that :)

single_kernel_mongo/utils/mongodb_users.py
patriciareinoso: According to the spec: "If the secret does not contain a correct password or user name, the charm will be blocked and display a message asking the user for a valid secret content."
I am raising to be able to show a meaningful message to the user.
I think that we should report invalid usernames. It will be easier for the user to catch typos and the will get clear feedback about their configuration.

tests/charms/mongodb_k8s_test_charm/config.yaml
patriciareinoso: it has the  `secret:` prefix but I cannot find the length in the documentation

patriciareinoso: I added a check in `get_secret_from_id` to assert the `secret:` prefix

single_kernel_mongo/utils/mongodb_users.py
patriciareinoso: I totally misread this comment. But anyways, now we have `InternalUsernames` and `InternalUsers`. WDYT?

single_kernel_mongo/state/charm_state.py
patriciareinoso: Done in the password_management_context

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: I refactored and created a "PasswordManagementContext" that help centralize operations. Please take a look

single_kernel_mongo/events/lifecycle.py
```diff
@@ -196,6 +207,10 @@ def on_secret_changed(self, event: SecretChangedEvent):
             )
             event.defer()
             return
+        except (DeferrableFailedHookChecksError, SetPasswordError) as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"{str(type(event))} will be skipped: {str(e)}")
```
Mehdi-Bendriss, line 213: nice catch!

single_kernel_mongo/managers/mongo.py
```diff
@@ -103,8 +103,8 @@ def mongod_ready(self, uri: str | None = None, direct: bool = True) -> bool:
         with MongoConnection(EMPTY_CONFIGURATION, actual_uri, direct=direct) as direct_mongo:
             return direct_mongo.is_ready
 
-    def set_user_password(self, user: MongoDBUser, password: str) -> str:
-        """Sets the password for a given username and return the secret id.
+    def set_user_password(self, user: MongoDBUser, password: str):
```
Mehdi-Bendriss: nit:
```suggestion
    def set_user_password(self, user: MongoDBUser, password: str) -> None:
```

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1045,3 +1130,104 @@ def get_statuses(self, scope: DPHScope, recompute: bool = False) -> list[StatusO
...
+
+    def get_password_management_statuses(self) -> list[StatusObject]:
+        """Returns the statuses related to password management."""
+        return self.map_password_management_state_to_status(
+            self.get_password_management_context().state
+        )
+
+    def map_password_management_state_to_status(
+        self, state: PasswordManagementState
+    ) -> list[StatusObject]:
```
Mehdi-Bendriss: why do we need to return a list? since we always only return a single or No statuses

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1045,3 +1130,104 @@ def get_statuses(self, scope: DPHScope, recompute: bool = False) -> list[StatusO
...
+                return True
+        return False
+
+    def get_password_management_statuses(self) -> list[StatusObject]:
+        """Returns the statuses related to password management."""
+        return self.map_password_management_state_to_status(
+            self.get_password_management_context().state
+        )
+
+    def map_password_management_state_to_status(
```
Mehdi-Bendriss: perhaps:

```suggestion
    def get_status_from_password_management_state(
```

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -441,9 +449,80 @@ def update_config_and_restart(self) -> None:
...
+            old_password = self.charm.state.get_user_password(user)
+            # only update user credentials if the password has changed
+            if new_password == old_password:
+                logger.debug(f"{user.username} password unchanged, skipping update.")
+                continue
+            try:
+                self.update_single_user_password(user, new_password)
+            except (SetPasswordError, WorkloadServiceError) as e:
+                logger.error(f"Failed to update password for {user.username}: {e}.")
+                self.charm.status_handler.set_running_status(
```
Mehdi-Bendriss: I'm not sure I understand why this should be a running status, as opposed to a regular status

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -534,24 +625,34 @@ def peer_changed(self) -> None:
...
+            logger.debug("Secret %s changed, but it's unknown", secret_id)
             return
-        logging.debug("Secret %s for scope %s changed, refreshing", secret_id, scope)
+        logger.debug("Secret %s for scope %s changed, refreshing", secret_id, scope)
         self.state.secrets.get(scope)
 
-        # Always update the PBM and mongodb exporter configuration so that if
-        # the secret changed, the configuration is updated and will still work
-        # afterwards.
+        # Update the PBM and mongodb exporter configuration so that ifthe secret changed,
```
Mehdi-Bendriss: nit:
```suggestion
        # Update the PBM and mongodb exporter configuration so that if the secret changed,
```

single_kernel_mongo/utils/mongodb_users.py
```diff
@@ -190,3 +194,31 @@ def get_user_from_username(username: str) -> MongoDBUser:
     if username == LogRotateUser.username:
         return LogRotateUser
     raise ValueError(f"Unknown user: {username}")
+
+
+def validate_charm_user_password_config(user_passwords: dict):
```
Mehdi-Bendriss: nit:
```suggestion
def validate_charm_user_password_config(user_passwords: dict) -> None:
```

single_kernel_mongo/utils/mongodb_users.py
```diff
@@ -190,3 +194,31 @@ def get_user_from_username(username: str) -> MongoDBUser:
...
+      - Only internal charmed usernames are allowed.
+      - Passwords must not be empty or only whitespace.
+      - Passwords must not exceed MAX_PASSWORD_LENGTH characters.
+
+    Raises:
+        InvalidPasswordError: if any validation rule is violated.
+    """
+    internal_usernames = {user.username for user in InternalUsers}
+    extra_users = set(user_passwords.keys()) - internal_usernames
+    if extra_users:
```
Mehdi-Bendriss: nit: walrus operator

single_kernel_mongo/utils/mongodb_users.py
```diff
@@ -190,3 +194,31 @@ def get_user_from_username(username: str) -> MongoDBUser:
...
+        raise InvalidPasswordError(f"Unexpected usernames provided: {', '.join(extra_users)}")
+
+    for username, password in user_passwords.items():
+        pwd = password.strip()
+        if not pwd:
```
Mehdi-Bendriss: nit: walrus

single_kernel_mongo/utils/mongodb_users.py
```diff
@@ -190,3 +194,31 @@ def get_user_from_username(username: str) -> MongoDBUser:
...
         return LogRotateUser
     raise ValueError(f"Unknown user: {username}")
+
+
+def validate_charm_user_password_config(user_passwords: dict):
+    """Validate a mapping of charm usernames to passwords.
+
+    Rules:
+      - Only internal charmed usernames are allowed.
+      - Passwords must not be empty or only whitespace.
```
Mehdi-Bendriss, line 204: is a 1 char password supported by mongodb?

tests/integration/helpers/upgrade.py
```diff
@@ -113,3 +114,46 @@ async def set_fcv(
         ops_test, app_name, substrate, replica_set_uri, admin_mongod_cmd, expecting_output=False
     )
     assert result.succeeded, f"Failed to set fcv to {fcv}."
+
+
+async def set_password_action(
```
Mehdi-Bendriss, line 119: do we still need this?

tests/integration/helpers/upgrade.py
```diff
@@ -113,3 +114,46 @@ async def set_fcv(
...
+    String with the password stored on the peer relation databag.
+    """
+    action = await ops_test.model.units.get(f"{app_name}/{unit_id}").run_action(
+        "set-password", **{"username": username, "password": password}
+    )
+    action = await action.wait()
+    return action.results
+
+
+async def get_password_action(
```
Mehdi-Bendriss, line 138: do we still need this?

Mehdi-Bendriss (commented):
Thank you Patricia, great work! I only have small nits and a couple of questions

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: I agree, there's no need to display it immediately so let's keep a regular status.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: That way it integrates cleanly with all the interfaces for statuses that only ever interact with lists and in case we wanted to report multiple statuses in the future, that would make everything easier

single_kernel_mongo/utils/mongodb_users.py
```diff
@@ -188,7 +180,11 @@ def get_hosts(self) -> set[str]:
 
 
 def get_user_from_username(username: str) -> MongoDBUser:
-    """Returns the key name for the password of the user."""
+    """Return the MongoDBUser instance that matches the given username.
+
+    Raises:
+        ValueError: If the username is not one of the known users.
+    """
```
Gu1nness, line 187: :purple_heart:

Gu1nness (dismissed):
Excellent work!
I like that new version much better and I love the Context object you built!
We should consider using that structure for all checks that we have as it makes the logic so much more understandable!

tests/integration/mongodb/test_charm.py
```diff
@@ -205,119 +208,115 @@ async def test_set_password_action(ops_test: OpsTest, substrate: Substrate) -> N
...
+    new_password = "NEW-PASSWORD"
+    secret_name = "test-secret"
+    secret_id = await ops_test.model.add_secret(
+        name=secret_name, data_args=[f"{OPERATOR_USERNAME}={new_password}"]
+    )
+    app_name = await get_app_name(ops_test)
+    await ops_test.model.applications[app_name].set_config(
+        {INTERNAL_USER_PASSWORD_CONFIG: secret_id}
+    )
+    expected_message = "Secret in system-users not granted."
```
reneradoi, line 224: Nice!

tests/integration/helpers/upgrade.py
reneradoi: I think only for the upgrades tests.

reneradoi (dismissed):
No concerns from my side, really thorough work! Don't forget to remove the actions in subsequent PR's from the actual charms once this is merged.

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: I would keep it aligned with `LDAPManager::map_state_to_statuses` and `BackupManager::map_backup_state_to_status`

tests/integration/helpers/upgrade.py
Mehdi-Bendriss: But this is for MongoDB 8, we don't have upgrades yet - the first revision upgrades will be based on will not have those actions

patriciareinoso: the `test_major_upgrades.py` nad `test_sharding_major_upgrades.py` are still running in this branch. These functions are used only for mongo6 and mongo7. That's why I kept those functions and moved them to the `tests/integration/helpers/upgrade.py` helper.

single_kernel_mongo/utils/mongodb_users.py
patriciareinoso: image

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: set as a regular status

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #91, opened 2025-09-29
patch: Outdated statuses and relation feasible

### LINKED ISSUE #87: Invalid relation between 2 shards does not display an error to the user
In a sharded cluster, if we try to relate 2 shards through the `shards` interface. No error message will be reported to the user.

## Steps to reproduce

Deploy a sharded deploy with two shards: `shard0` and `shard1` and integrate them as followed:
```
 juju relate shard0:sharding shard1:config-server
```

## Expected behavior

`shard1` is acting as a config-server in this relation, which is invalid. I expected:
- Blocked status
- Message indicating to remove the relation

## Actual behavior

`shard1` does not report any status to the user.

Image

Image

## Versions

6/edge, mongodb and mongodb-k8s

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

On backup manager, clear outdated statuses.
This also prevents the credentials from being updated while a backup is in progress, which was not given for granted before that.

On mongodb operator and mongos operator, improves drastically the `is_relation_feasible` method to check for way more cases, fixing #87 and other similar cases.
This PR also ensures that all those cases are reported in a similar fashion, with the same logic so that everything is always properly in sync.

Fixes: #87 

## 📑 Motivation and Context

This is part of the preparations for new stable releases, and fulfilling an ongoing improvement of stability and edge cases handling.

I added unit testing of the methods modified.

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/events/backups.py
+++ b/single_kernel_mongo/events/backups.py
@@ -13,6 +13,7 @@ from botocore.exceptions import ConnectTimeoutError, SSLError
 from ops.charm import ActionEvent, RelationBrokenEvent, RelationJoinedEvent
 from ops.framework import Object
 
+from single_kernel_mongo.config.models import BackupState
 from single_kernel_mongo.config.relations import ExternalRequirerRelations
 from single_kernel_mongo.config.statuses import BackupStatuses, MongoDBStatuses
 from single_kernel_mongo.exceptions import (
@@ -122,6 +123,15 @@ class BackupEventsHandler(Object):
             )
             return
 
+        match self.manager.backup_state():
+            case BackupState.BACKUP_RUNNING | BackupState.RESTORE_RUNNING:
+                defer_event_with_info_log(
+                    logger, event, action, "Backup/Restore running, not changing credentials."
+                )
+                return
+            case _:
+                pass
+
         if not self.manager.validate_s3_config():
             logger.warning(
                 "Relation to S3 charm exists but not all necessary configurations have been set."
@@ -137,6 +147,12 @@ class BackupEventsHandler(Object):
         credentials = self.s3_client.get_s3_connection_info()
 
         try:
+            # We can clear all statuses as they will get rewritten right after if needed.
+            # The only ones we don't want to lose were checked earlier and returned early.
+            self.manager.state.statuses.clear(
+                scope="unit",
+                component=self.manager.name,
+            )
             # First create the bucket if it does not exist.
             self.manager.create_bucket(credentials=credentials)
             # Then set the config options on PBM.
@@ -207,6 +223,7 @@ class BackupEventsHandler(Object):
     def _on_s3_relation_broken(self, event: RelationBrokenEvent) -> None:
         """Proceed on s3 relation broken."""
         self.manager.cleanup_certs_and_restart(event.relation)
+        self.manager.state.statuses.clear(scope="unit", component=self.manager.name)
 
     def _on_create_backup_action(self, event: ActionEvent) -> None:
         action = "backup"
--- a/single_kernel_mongo/events/backups.py
+++ b/single_kernel_mongo/events/backups.py
@@ -13,7 +13,6 @@ from botocore.exceptions import ConnectTimeoutError, SSLError
 from ops.charm import ActionEvent, RelationBrokenEvent, RelationJoinedEvent
 from ops.framework import Object
 
-from single_kernel_mongo.config.models import BackupState
 from single_kernel_mongo.config.relations import ExternalRequirerRelations
 from single_kernel_mongo.config.statuses import BackupStatuses, MongoDBStatuses
 from single_kernel_mongo.exceptions import (
@@ -122,15 +121,6 @@ class BackupEventsHandler(Object):
                 "Set PBM configurations, pbm-agent service not found.",
             )
             return
-
-        match self.manager.backup_state():
-            case BackupState.BACKUP_RUNNING | BackupState.RESTORE_RUNNING:
-                defer_event_with_info_log(
-                    logger, event, action, "Backup/Restore running, not changing credentials."
-                )
-                return
-            case _:
-                pass
 
         if not self.manager.validate_s3_config():
             logger.warning(
--- a/single_kernel_mongo/events/database.py
+++ b/single_kernel_mongo/events/database.py
@@ -122,9 +122,12 @@ class DatabaseEventsHandler(Object):
     def pass_hook_checks(self, event: RelationEvent) -> bool:
         """Runs the pre-hooks checks for MongoDBProvider, returns True if all pass."""
         # First, ensure that the relation is valid, useless to do anything else otherwise
-        if not self.dependent.state.is_role(
-            MongoDBRoles.MONGOS
-        ) and not self.dependent.is_relation_feasible(event.relation.name):
+        if (
+            not self.dependent.state.is_role(MongoDBRoles.MONGOS)
+            and (status := self.dependent.get_relation_feasible_status(self.relation_name))
+            is not None
+        ):
+            self.dependent.state.statuses.add(status, scope="unit", component=self.dependent.name)
             return False
 
         # We shouldn't try to create or update users if the database is not
--- a/single_kernel_mongo/events/database.py
+++ b/single_kernel_mongo/events/database.py
@@ -122,10 +122,8 @@ class DatabaseEventsHandler(Object):
     def pass_hook_checks(self, event: RelationEvent) -> bool:
         """Runs the pre-hooks checks for MongoDBProvider, returns True if all pass."""
         # First, ensure that the relation is valid, useless to do anything else otherwise
-        if (
-            not self.dependent.state.is_role(MongoDBRoles.MONGOS)
-            and (status := self.dependent.get_relation_feasible_status(self.relation_name))
-            is not None
+        if not self.dependent.state.is_role(MongoDBRoles.MONGOS) and (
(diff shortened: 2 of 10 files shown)


patriciareinoso (dismissed):
I am happy with this improvement. When you are starting with mongodb all the possible role/relations can be confusion. Users will have better guidance

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -337,12 +337,14 @@ def restart_charm_services(self, force: bool = False) -> None:
             raise
 
     @override
-    def is_relation_feasible(self, name: str) -> bool:
+    def get_relation_feasible_status(self, name: str) -> StatusObject | None:
         """Checks if the relation is feasible.
 
         In the mongos case, we only allow the mongos proxy client relation.
```
patriciareinoso, line 343: giving it another thought. I find confusing that in the mongodb implementation we will have verifications over multiple role/relations combinations. but in the mongos implementation, "we only allow, mongos proxy". This makes me think that any other relation that we may try to add would be invalid

single_kernel_mongo/events/backups.py
```diff
@@ -207,6 +223,7 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent) -> None:  #
     def _on_s3_relation_broken(self, event: RelationBrokenEvent) -> None:
         """Proceed on s3 relation broken."""
         self.manager.cleanup_certs_and_restart(event.relation)
+        self.manager.state.statuses.clear(scope="unit", component=self.manager.name)
```
Mehdi-Bendriss, line 216: What is the current behavior when a backup/restore is running?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -995,19 +1003,26 @@ def is_removing_last_replica(self) -> bool:
...
+            if (
+                self.model.relations[relation_name.value]
+                and (status := self.get_relation_feasible_status(relation_name)) is not None
+            ):
+                statuses.append(status)
 
         if not self.state.is_sharding_component and self.state.has_sharding_integration:
-            statuses.append(MongoDBStatuses.INVALID_SHARDING_REL.value)
-        elif rev_status := self.cluster_version_checker.get_cluster_mismatched_revision_status():
             # don't bother checking revision mismatch on sharding interface if replica
```
Mehdi-Bendriss, line 1055: should this go?

single_kernel_mongo/managers/sharding.py
```diff
@@ -184,7 +184,8 @@ def assert_pass_sanity_hook_checks(self) -> None:
             raise NonDeferrableFailedHookChecksError("is only executed by config-server")
         if not self.state.db_initialised:
             raise DeferrableFailedHookChecksError("db is not initialised.")
-        if not self.dependent.is_relation_feasible(self.relation_name):
+        if (status := self.dependent.get_relation_feasible_status(self.relation_name)) is not None:
```
Mehdi-Bendriss: nit:

```suggestion
        if status := self.dependent.get_relation_feasible_status(self.relation_name):
```

single_kernel_mongo/managers/sharding.py
```diff
@@ -508,7 +509,8 @@ def assert_pass_sanity_hook_checks(self, is_leaving: bool) -> None:
             raise NonDeferrableFailedHookChecksError("is only executed by shards")
         if not self.state.db_initialised:
             raise DeferrableFailedHookChecksError("db is not initialised.")
-        if not self.dependent.is_relation_feasible(self.relation_name):
+        if (status := self.dependent.get_relation_feasible_status(self.relation_name)) is not None:
```
Mehdi-Bendriss, line 512: nit:
```suggestion
        if status := self.dependent.get_relation_feasible_status(self.relation_name):
```

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -995,19 +1003,26 @@ def is_removing_last_replica(self) -> bool:
...
         statuses = []
-        if not self.cluster_manager.is_valid_mongos_integration():
-            statuses.append(MongoDBStatuses.INVALID_MONGOS_REL.value)
-
         if not self.backup_manager.is_valid_s3_integration():
             statuses.append(MongoDBStatuses.INVALID_S3_REL.value)
-
-        if self.state.client_relations and self.state.is_sharding_component:
-            statuses.append(MongoDBStatuses.INVALID_DB_REL.value)
+        # Add valid statuses for all invalid relations integratedd
```
Mehdi-Bendriss:
```suggestion
        # Add valid statuses for all invalid integrated relations
```

Mehdi-Bendriss (commented):
Thanks Neha! I have a question and a couple of nits

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: The mongos implementation only ever accepts `mongos_proxy` and `cluster` relation.
We also don't use `get_relation_feasible` on other relations in mongos.
Maybe we should add more checks on mongos side as well ?

single_kernel_mongo/events/backups.py
Gu1nness: Excellent question !
I think it would leave a broken backup and maybe the restarts would fail.

I think it would be safer to delay that event until the backup/restore finishes!

I can add that logic in a loop so that we wait until the backup/restore finishes.

Should we handle it the same way the shard waits for draining ?
I think so because we can't defer that event and we really want to wait until the backup is finished.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: It's been moved below with the same comment.
There's no need to check for invalid versions on invalid relations so we early return and otherwise continue with the cluster status checking.

single_kernel_mongo/managers/mongos_operator.py
patriciareinoso: my concern in this function is that in case of a `cluster` relation (which is a valid one), it will return false.

I do not think that we should add checks for the other relations (sharding, config-server...) because we would not be able to add them to the mongos charm.

Gu1nness: Oh yes, it's not used on that relation because there's no doubt, but I can add that check as well

Gu1nness: Done.

patriciareinoso (dismissed):

single_kernel_mongo/events/backups.py
Mehdi-Bendriss: I would agree, if a backup / restore is running, we should protect the user from a broken state - so similar to shards draining handling it is - with a graceful error handling if any of those operations fail

Gu1nness: I've added the loop logic, I'll add some error handling as well.

Gu1nness: "fun fact" we don't have any kind of logic for failed backups for now, what should I do ?

Mehdi-Bendriss: Summary of offline conversation: we wait for any currently backup or restore to complete if any. In case of an error, we do not retry.

single_kernel_mongo/events/database.py
```diff
@@ -122,9 +122,12 @@ def _on_relation_event(self, event: RelationEvent):
     def pass_hook_checks(self, event: RelationEvent) -> bool:
         """Runs the pre-hooks checks for MongoDBProvider, returns True if all pass."""
         # First, ensure that the relation is valid, useless to do anything else otherwise
-        if not self.dependent.state.is_role(
-            MongoDBRoles.MONGOS
-        ) and not self.dependent.is_relation_feasible(event.relation.name):
+        if (
+            not self.dependent.state.is_role(MongoDBRoles.MONGOS)
+            and (status := self.dependent.get_relation_feasible_status(self.relation_name))
+            is not None
```
Mehdi-Bendriss: nit:
```suggestion
```

single_kernel_mongo/events/backups.py
```diff
@@ -122,6 +123,15 @@ def _on_s3_credential_changed(self, event: CredentialsChangedEvent) -> None:  #
             )
             return
 
+        match self.manager.backup_state():
+            case BackupState.BACKUP_RUNNING | BackupState.RESTORE_RUNNING:
```
Mehdi-Bendriss: I'm not sure this can work in this case, if the credentials have changed, there is no guarantee the previous credentials are still valid for the backup / restore to continue

Mehdi-Bendriss (commented):
Thanks Neha! I have 1 comment

single_kernel_mongo/events/backups.py
Gu1nness: The credentials haven't been rotated yet, so it still uses the former credentials until they finish rotating.
Or you mean the credentials that where used have been revoked ?

Mehdi-Bendriss: I mean the latter, the credentials may no longer be accepted by the cloud

Gu1nness: Ok, let's remove it then

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

patriciareinoso (approved):

PR #92, opened 2025-09-30
patch(DPE-8421): Graceful validation of config options

### LINKED ISSUE #66: Deploying a mongodb charm using an invalidad role does not show message to the user
## Steps to reproduce

1. juju deploy mongodb-k8s --config role="sharding" shard0

## Expected behavior

An error is displayed to the user for them to know that a mistake was done in the deployment configuration.

## Actual behavior

When the application is deployed, its status is active

Image

We can even related it to other mongdb charms and a bunch of error appear but nothing related to the mistake in the configuration.

Image

Image

## Versions

mongodb and mongodb-k8s operators from 6/stable and 6/edge

Operating system: noble

Juju CLI: 3.6

## Log output

Juju debug log:

```
unit-shard1-0: 11:48:08 INFO juju.worker.uniter.operation ran "update-status" hook (via hook dispatching script: dispatch)
unit-shard1-0: 11:50:04 INFO unit.shard1/0.juju-log Component MongoDBRoles.UNKNOWN is not a shard, cannot be integrated to a config-server.
``` 

## Additional context

I would be nice to have the valid values in the charm config.yaml files. The description of the role parameter is misleading

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

Proper validation of the role config option: it can be updated before the start event, but not after.
The start event will not run until the role config option is valid.
Adds unit test + integration test for that feature.
Also fixes some missing statuses regarding PBM: in case PBM should be running and is not, report it in the status.

Finally, the get_statuses function is reworked for the MongoDBOperator, so that app statuses are reported both for the app and the units, while the unit statuses will only be reported for the unit.
This allows for a better app report of statuses in the main operator.

## 📑 Motivation and Context

We discovered recently that the role config option was never properly validated in some cases allowing weird deployments.

Closes #66 

This is a bug fix, part of DPE-8421.

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/core/structured_config.py
+++ b/single_kernel_mongo/core/structured_config.py
@@ -121,8 +121,8 @@ class MongoDBCharmConfig(MongoConfigModel):
 
     @field_validator("role", mode="before")
     @classmethod
-    def invalid_role_to_unknown(cls, v: str) -> MongoDBRoles:
-        """If the value is neither replication, config-server, or shard, returns unknown."""
+    def invalid_role_to_invalid(cls, v: str) -> MongoDBRoles:
+        """If the value is neither replication, config-server, or shard, returns invalid."""
         if v not in MongoDBRoles.valid_roles():
             return MongoDBRoles.INVALID
         return MongoDBRoles(v)
--- a/single_kernel_mongo/core/structured_config.py
+++ b/single_kernel_mongo/core/structured_config.py
@@ -44,6 +44,7 @@ class MongoDBRoles(str, Enum):
     """The different accepted roles for a charm."""
 
     UNKNOWN = ""
+    INVALID = "invalid"
     REPLICATION = "replication"
     CONFIG_SERVER = "config-server"
     SHARD = "shard"
@@ -123,7 +124,7 @@ class MongoDBCharmConfig(MongoConfigModel):
     def invalid_role_to_unknown(cls, v: str) -> MongoDBRoles:
         """If the value is neither replication, config-server, or shard, returns unknown."""
         if v not in MongoDBRoles.valid_roles():
-            return MongoDBRoles.UNKNOWN
+            return MongoDBRoles.INVALID
         return MongoDBRoles(v)
 
 
--- a/single_kernel_mongo/events/lifecycle.py
+++ b/single_kernel_mongo/events/lifecycle.py
@@ -51,6 +51,7 @@ from single_kernel_mongo.config.statuses import (
 from single_kernel_mongo.core.operator import OperatorProtocol
 from single_kernel_mongo.exceptions import (
     ContainerNotReadyError,
+    InvalidConfigRoleError,
     InvalidLdapQueryTemplateError,
     InvalidLdapUserToDnMappingError,
     UpgradeInProgressError,
@@ -121,6 +122,10 @@ class LifecycleEventsHandler(Object):
             logger.info("Not ready to start.")
             event.defer()
             return
+        except InvalidConfigRoleError:
+            logger.info("Missing a valid role.")
+            event.defer()
+            return
         except WorkloadNotReadyError:
             logger.info("Still starting service.")
             self.dependent.state.statuses.add(
@@ -163,6 +168,16 @@ class LifecycleEventsHandler(Object):
             self.dependent.update_config_and_restart()
         except (UpgradeInProgressError, WaitingForLeaderError):
             event.defer()
+        except InvalidConfigRoleError:
+            logger.info("Invalid config role.")
+            if self.charm.unit.is_leader():
+                self.dependent.state.statuses.add(
+                    MongoDBStatuses.INVALID_ROLE.value,
+                    scope="app",
+                    component=self.dependent.name,
+                )
+            event.defer()
+            return
         except InvalidLdapUserToDnMappingError:
             self.dependent.state.statuses.add(
                 LdapStatuses.INVALID_LDAP_USER_MAPPING.value,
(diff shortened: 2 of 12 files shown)


single_kernel_mongo/core/structured_config.py
```diff
@@ -123,7 +124,7 @@ class MongoDBCharmConfig(MongoConfigModel):
     def invalid_role_to_unknown(cls, v: str) -> MongoDBRoles:
         """If the value is neither replication, config-server, or shard, returns unknown."""
```
patriciareinoso:
```suggestion
        """If the value is neither replication, config-server, or shard, returns invalid."""
```

patriciareinoso: actually, the function name does no longer correspond

single_kernel_mongo/state/app_peer_state.py
```diff
@@ -78,10 +78,11 @@ def role(self) -> MongoDBRoles:
...
         Either from the app databag or from the default from config.
         """
-        if not (databag_role := self.relation_data.get(AppPeerDataKeys.ROLE.value)):
+        if (
+            not (databag_role := self.relation_data.get(AppPeerDataKeys.ROLE.value))
+            or not self.relation
+        ):
             return MongoDBRoles.UNKNOWN
-        if not self.relation:
-            return self._role
```
patriciareinoso, line 84: it seems like _role is no longer used and can be removed

single_kernel_mongo/state/app_peer_state.py
```diff
@@ -78,10 +78,11 @@ def role(self) -> MongoDBRoles:
 
         Either from the app databag or from the default from config.
```
patriciareinoso:
```suggestion
        Either from the app databag or UNKNOWN.
```

patriciareinoso (commented):
Thanks for fixing this. This was one of the first issues that I run into when I incorrectly set the role to "sharding".

LGTM. I just added minor comments to clean up and keep the comments aligned with the code. 

You can add the https://github.com/canonical/mongo-single-kernel-library/issues/66 to the PR description. 

And also, as I said in the bug description: 'it would be nice to have the valid values in the charm config.yaml files" (specially for the replica set)

single_kernel_mongo/core/structured_config.py
Gu1nness: Fixed

single_kernel_mongo/events/lifecycle.py
```diff
@@ -163,6 +168,16 @@ def on_config_changed(self, event: ConfigChangedEvent):
             self.dependent.update_config_and_restart()
         except (UpgradeInProgressError, WaitingForLeaderError):
             event.defer()
+        except InvalidConfigRoleError:
+            logger.info("Invalid config role.")
+            if self.charm.unit.is_leader():
+                self.dependent.state.statuses.add(
+                    MongoDBStatuses.INVALID_ROLE.value,
+                    scope="app",
```
Mehdi-Bendriss, line 176: nice, I believe this would be our first non-refresh related `app` status in mongodb :)

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -298,9 +299,7 @@ def install_workloads(self) -> None:
         # Truncate the file.
         self.workload.write(self.workload.paths.config_file, "")
 
-    @override
-    def prepare_for_startup(self) -> None:
-        """Handler on start."""
+    def _prepare_checks(self):
```
Mehdi-Bendriss: I think we can rename this method, as it does quite a few things, and add a docstring + return type hint

perhaps: `_run_startup_checks`, and offload the status logic up in the call chain

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1019,29 +1046,37 @@ def get_statuses(self, scope: DPHScope, recompute: bool = False) -> list[StatusO
         if not recompute:
             return self.state.statuses.get(scope=scope, component=self.name).root
 
-        if scope == "app":
-            return charm_statuses
+        if scope == "unit" and not self.workload.workload_present:
+            return [CharmStatuses.MONGODB_NOT_INSTALLED.value]
```
Mehdi-Bendriss, line 1056: to be clear, this should never run - right? As we should be error'ing the install hook should it fail to install, so no subsequent hook will ever display the status

tests/integration/helpers/common.py
```diff
@@ -854,6 +854,12 @@ async def check_status_detail(ops_test: OpsTest, app_name: str, status: str, mes
         assert unit_statuses[0]["Message"] == message
 
 
+async def check_app_status(ops_test: OpsTest, app_name: str, status: str, message: str) -> None:
```
Mehdi-Bendriss, line 857: missing docstrings

tests/integration/mongodb/test_charm_invalid_config.py
```diff
@@ -0,0 +1,61 @@
...
+)
+from ..helpers.types import Substrate
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(
+    ops_test: OpsTest, mongodb_charm: str, substrate: Substrate, mongod_resource, base_app_name
+):
```
Mehdi-Bendriss, line 29: missing docstrings

Mehdi-Bendriss (commented):
Thanks Neha! Looks good, I have a few minor comments

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: On VM this should never run.
On K8s, it will happen if the container is not available.

single_kernel_mongo/events/lifecycle.py
Gu1nness: I think it is indeed!

Mehdi-Bendriss (approved):

patriciareinoso (approved):

PR #96, opened 2025-10-01
patch: Implement TLS v4

patriciareinoso:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

This PR replace TLS v3 with TLS v4. Internal and external certificates will be handled separately.

We now have 2 requirer objects (`TLSCertificatesRequiresV4`): one for the peer certificates (internal) and one for client certificates (external)

When the TLS relations are created, the creation of a CSR and a private key is automatically done in the TLS library. Once the certificates are ready, a `certificate_available` is emitted by the library. We do not need to manage the CSR anymore, this is done automatically by the library.

The certificate, CA, CA chain and private key received are written when `certificate_available` event is handled. 

Edge cases: (behavior from V3 was kept)
1. Shard integrated with TLS before integrating with the config-server:
   The shard stores the certificates using the shard application name as a subject for the CSR. When the shard is integrated to the config-server, a refresh of the certificates is triggered, this time, a new CSR is automatically created by the library, using the config-server as subject. Once, the certificates are available, the new certs will be stored.

2. Mongos integrated with TLS before integrating with the config-server:
  Mongos does not store the certificates (`certificate_available`) are ignored until the relation with the config-server exists. Once Mongos is integrated with the config-server,  a refresh of the certificates is triggered, this time, a new CSR is automatically created by the library, using the config-server as subject. Once, the certificates are available, the new certs will be stored.
 

## ✅ Checklist


single_kernel_mongo/managers/config.py
```diff
@@ -333,7 +333,7 @@ def audit_options(self) -> dict[str, Any]:
     def auth_parameter(self) -> dict[str, Any]:
         """The auth mode."""
         cmd = {"security": {"authorization": "enabled"}} if self.auth else {}
-        if self.state.tls.internal_enabled and self.state.tls.external_enabled:
+        if self.state.tls.peer_enabled and self.state.tls.client_enabled:
```
patriciareinoso: maybe only needs peer to be enabled

single_kernel_mongo/managers/tls.py
```diff
@@ -231,92 +159,68 @@ def enable_certificates_for_unit(self):
...
-        scope = "int" if internal else "ext"
-        label_name = f"{scope}-{WAIT_CERT_UPDATE}"
-
-        return json.loads(self.state.unit_peer_data.get(label_name) or "false")
+        logger.info(
+            f"{TLSType.PEER.value if internal else TLSType.CLIENT.value} certificate secrets updated."
+        )
 
-    def is_waiting_for_both_certs(self) -> bool:
+    def is_waiting_for_a_cert(self) -> bool:
```
patriciareinoso, line 268: on a second though, this works only for mongos before being integrated with the config-server. 

on  `tls_events.refresh_certificates()` we do not remove the certificate secret, as we did before when creating the CSR. And we shouldn't, because if we trigger the `refresh_tls_certificates_event`  and the library detects that the the CSR did not change, we won't get a new one -> no certificates_available event -> the certificate secret will stay as None

Gu1nness: I updated that.

single_kernel_mongo/managers/tls.py
```diff
@@ -233,153 +175,235 @@ def enable_certificates_for_unit(self):
...
-            current_sans = self.get_current_sans(internal)
-            current_sans_ip = set(current_sans["sans_ips"]) if current_sans else set()
-            expected_sans_ip = set(self.get_new_sans()["sans_ips"]) if current_sans else set()
-            sans_ip_changed = current_sans_ip ^ expected_sans_ip
-
-            if not sans_ip_changed:
-                continue
+        if tls_peer_private_key_id := self.dependent.config.tls_peer_private_key_id:
+            if peer_private_key := self.update_private_key(tls_peer_private_key_id, internal=True):
+                self.dependent.tls_events.peer_certificate._private_key = peer_private_key
```
Gu1nness, line 313: Opened [this](https://github.com/canonical/tls-certificates-interface/issues/382) so I don't have to use private attribute to set the new key immediately.

single_kernel_mongo/events/tls.py
```diff
@@ -9,205 +9,225 @@
...
+        self.framework.observe(self.charm.on.secret_changed, self._on_secret_changed)
+
+    @property
+    def tls_mapping(self) -> dict[bool, TLSCertificatesRequiresV4]:
+        """Mapping of boolean to a TLS requirer instance.
+
+        The boolean value is True if the requirer is for internal certificates,
+        and False for the client certificates.
+        """
+        return {True: self.peer_certificate, False: self.client_certificate}
```
Mehdi-Bendriss, line 97: This is an odd data structure, why not a dict[str, ..] ?

single_kernel_mongo/events/tls.py
```diff
@@ -9,205 +9,225 @@
...
+
+    def _on_tls_relation_created(self, event: RelationCreatedEvent) -> None:
+        """Handler for relation created."""
+        if event.relation.name != ExternalRequirerRelations.PEER_TLS.value:
```
Mehdi-Bendriss: Can you explain this check?

single_kernel_mongo/events/tls.py
```diff
@@ -9,205 +9,225 @@
...
             return
-        if self.manager.state.upgrade_in_progress:
```
Mehdi-Bendriss: is the upgrades reconciliation happening later?

single_kernel_mongo/events/tls.py
```diff
@@ -9,205 +9,225 @@
...
-            return
+        """Handle the relation broken event."""
+        state = self.manager.get_tls_management_state()
+        match state:
+            case (
+                TlsManagementState.DB_NOT_INTIALIZED
+                | TlsManagementState.MONGOS_DB_NOT_INITIALIZED
+                | TlsManagementState.UPGRADE_IN_PROGRESS
+            ):
+                defer_event_with_info_log(logger, event, str(type(event)), state.value)
```
Mehdi-Bendriss: why should we defer if the TLS relation is broken and the DB is not initialized?

single_kernel_mongo/events/tls.py
```diff
@@ -9,205 +9,225 @@
...
+    def _on_tls_relation_created(self, event: RelationCreatedEvent) -> None:
+        """Handler for relation created."""
+        if event.relation.name != ExternalRequirerRelations.PEER_TLS.value:
             return
-        if self.manager.state.upgrade_in_progress:
-            logger.warning(
-                "Enabling TLS is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
+        if self.manager.state.is_role(MongoDBRoles.MONGOS):
+            self.manager.state.statuses.delete(
+                MongosStatuses.MISSING_PEER_TLS_REL.value,
```
Mehdi-Bendriss, line 103: I'm not sure I understand how we could infer the type here?

single_kernel_mongo/managers/config.py
```diff
@@ -332,16 +332,20 @@ def audit_options(self) -> dict[str, Any]:
...
         cmd = {"security": {"authorization": "enabled"}} if self.auth else {}
-        if self.state.tls.internal_enabled and self.state.tls.external_enabled:
+        if self.state.tls.peer_enabled:
             return always_merger.merge(
                 cmd,
                 {
                     "security": {"clusterAuthMode": "x509"},
                     "net": {
                         "tls": {
+                            "mode": "preferTLS",
```
Mehdi-Bendriss, line 342: Didn't we go with `requireTLS` in a previous PR?

single_kernel_mongo/managers/config.py
```diff
@@ -332,16 +332,20 @@ def audit_options(self) -> dict[str, Any]:
...
                             "allowInvalidCertificates": True,
-                            "clusterCAFile": f"{self.workload.paths.int_ca_file}",
+                            "certificateKeyFile": f"{self.workload.paths.int_pem_file}",
+                            "CAFile": f"{self.workload.paths.int_ca_file}",
+                            "disabledProtocols": "TLS1_0,TLS1_1",
```
Mehdi-Bendriss, line 346: Is this impose TLS1_2?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1254,3 +1248,18 @@ def clear_password_management_statuses(self) -> None:
                 scope="app",
                 component=self.name,
             )
+
+    def _should_skip_because_of_tls(self) -> bool:
```
Mehdi-Bendriss: 1. Missing docstring
2. I would perhaps clarify the name, that it's skipping due to an incomplete TLS setup

single_kernel_mongo/managers/sharding.py
```diff
@@ -959,24 +961,49 @@ def should_skip_shard_status(self) -> bool:
 
         return False
 
-    def get_tls_status(self) -> StatusBase | None:
-        """Returns the TLS status of the sharded deployment."""
-        shard_has_tls, config_server_has_tls = self.tls_status()
-        match (shard_has_tls, config_server_has_tls):
+    def get_tls_status(self, internal: bool):
```
Mehdi-Bendriss: missing return type hint

single_kernel_mongo/managers/tls.py
```diff
@@ -233,153 +173,237 @@ def enable_certificates_for_unit(self):
...
-            self.workload.write(self.workload.paths.int_ca_file, internal_ca)
-        if internal_pem is not None:
-            self.workload.write(self.workload.paths.int_pem_file, internal_pem)
+        logger.info(
+            f"Pushing {TLSType.PEER.value if internal else TLSType.CLIENT.value} TLS certificates to filesystem"
+        )
+        ca, pem = self.get_tls_file_contents(internal=internal)
+
+        if internal:
+            if ca is not None:
```
Mehdi-Bendriss: nit: for this one and the following ones
```suggestion
            if ca:
```

tests/integration/helpers/tls.py
```diff
@@ -57,6 +61,48 @@ def internal_cert_path(substrate: Substrate):
     return f"{MONGODB_ROCK_CONF_DIR}/internal-ca.crt"
 
 
+async def integrate_apps_with_tls(
```
Mehdi-Bendriss, line 64: It seems like we are not testing when either client or peer (not both) are integrated

Mehdi-Bendriss (commented):
Thank you both Neha and Patricia! Great work! :)

I left a few comments, and I have 2 main points:
- reconciling this PR with refresh v3 - with all the safeguards added to that PR
- missing integration tests testing either client or peer

single_kernel_mongo/events/tls.py
Gu1nness: Because we always identify the internal/external by a boolean, and this structure (even if odd) was the easiest for me if I wanted a dict-like access.
However I could bring that to a different shape

single_kernel_mongo/events/tls.py
Gu1nness: It's something that needs rework, I did not work on that part.
But I think I understand what was the goal, just not updated yet to two relations

single_kernel_mongo/events/tls.py
Gu1nness: We can't defer the relation created events but we can defer the certificate available, which is exactly what we do later.

single_kernel_mongo/events/tls.py
Gu1nness: I think this is a miss, we should directly remove the relation.

single_kernel_mongo/events/tls.py
Gu1nness: WDYM ?

single_kernel_mongo/managers/config.py
Gu1nness: We talked about it: if you have only internal, you must use `preferTLS` otherwise all clients will be mandated to use TLS relations (which is not what we want).
MongoDB has only one setting for TLS mode that works for both internal and external, with the following values and semantics:
```
disabled: The server does not use TLS.
allowTLS: Connections between servers do not use TLS. For incoming connections, the server accepts both TLS and non-TLS.
preferTLS: Connections between servers use TLS. For incoming connections, the server accepts both TLS and non-TLS.
requireTLS: The server uses and accepts only TLS encrypted connections.
```

single_kernel_mongo/managers/config.py
Gu1nness: It disables the use of 1.0 and 1.1 so at least 1.2 yes.

single_kernel_mongo/config/statuses.py
```diff
@@ -182,11 +224,28 @@ class CharmStatuses(Enum):
...
+    INVALID_PEER_PRIVATE_KEY = StatusObject(
+        status="blocked",
+        message="Invalid peer private key",
+        check="Peer private key format validation failed",
+        action="Update the peer private key secret.",
+    )
+    INVALID_CLIENT_PRIVATE_KEY = StatusObject(
+        status="blocked",
+        message="Invalid client private key",
+        check="CLient private key format validation failed",
```
patriciareinoso:
```suggestion
        check="Client private key format validation failed.",
```

single_kernel_mongo/config/statuses.py
```diff
@@ -343,12 +402,39 @@ def waiting_for_shard_upgrade(
...
+        action="Align the peer TLS configuration in all the cluster components: remove the client-certificates relation from the shard.",
     )
-    CA_MISMATCH = StatusObject(
-        status="blocked", message="Shard CA and Config-Server CA don't match."
+    PEER_CA_MISMATCH = StatusObject(
+        status="blocked",
+        message="Shard internal CA and Config-Server internal CA don't match.",
+        short_message="Peer CA mismatch.",
+        check="Relation validation failed.",
+        action="Verify the certificates relations. Use the same CA for all cluster components.",
```
patriciareinoso:
```suggestion
        action="Verify the peer-certificates relations. Use the same CA for all cluster components.",
```

single_kernel_mongo/config/statuses.py
```diff
@@ -343,12 +402,39 @@ def waiting_for_shard_upgrade(
...
+    )
+    CLIENT_CA_MISMATCH = StatusObject(
+        status="blocked",
+        message="Shard client CA and Config-Server client CA don't match.",
+        short_message="Client CA mismatch.",
+        check="Relation validation failed.",
+        action="Verify the certificates relations. Use the same CA for all cluster components.",
```
patriciareinoso:
```suggestion
        action="Verify the client-certificates relations. Use the same CA for all cluster components.",
```

single_kernel_mongo/state/cluster_state.py
```diff
@@ -74,6 +75,13 @@ def internal_ca_secret(self) -> str | None:
             return None
         return self.relation_data.get(ClusterStateKeys.INT_CA_SECRET.value, None)
 
+    @property
+    def external_ca_secret(self) -> str | None:
+        """Returns the internal CA secret."""
```
patriciareinoso:
```suggestion
    def external_ca_secret(self) -> str | None:
        """Returns the external CA secret."""
```

patriciareinoso (commented):
small improvements

single_kernel_mongo/events/tls.py
Mehdi-Bendriss: I meant that we are targeting `MISSING_PEER_TLS_REL` and not `MISSING_CLIENT_TLS_REL`

Gu1nness: Oh it was a miss here: this function was checking only for the peer relation to cleanup some statuses in case of sharding integration.
It's now removed and updated to remove all statuses systematically (remove a non existing status is a noop)

Mehdi-Bendriss (approved):

Gu1nness (approved):

PR #97, opened 2025-10-02
patch: Upgrades V3 implementation

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

This is the implementation of  [refresh-v3](canonical-charm-refresh.readthedocs-hosted.com/latest/).
It replaces the old implementation of refreshes V2, being more stable and having a real implementation as a library.
The biggest changes:
 * In the pre-refresh checks, the charm doesn't connect directly to a shard to see that the data was migrated, because MongoDB 8 forbids so. Instead, it checks that the primary was indeed changed and that the data was accessible
 * Removal of the `post_app_upgrade` and `post_cluster_upgrade` custom ops events. They don't work well with refresh V3, and have hence been replaced by a logic that runs automatically on all events under some conditions.
 * The Upgrade Manager is a compatibility layer between refresh statuses and the charm statuses, it comes directly from charmed-etcd-operator.
 * Tagging: Add a new tag to each commit so that it respects the versioning system, it will make it easier to test for upgrades.
 * Updates to the script building the test charms to take the new versioning into account.
 * Rework of the snap installation system, to respect `charm-refresh`
 * Start events run only when no refresh is in progress, otherwise the start logic gets run in the `__init__` still following some checks. This is a workflow similar to what postgres has been implementing in refresh V3, and allows for post-upgrade checks.

## 📑 Motivation and Context

As we prepare MongoDB 8 stable, we allow ourselves to bring as much breaking changes as we can, since there is no upgrade path from 6 to 8 without a redeployment.
Migrating to refreshes V3 is a way to modernize upgrades, have more stable upgrades, and will make new changes in the upgrades easier.

This has been tested manually for the pre-refresh-check, and unit tested.

## ✅ Checklist


single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,587 @@
...
+
+        if self.dependent.name == CharmKind.MONGOS:
+            if not self.is_mongos_able_to_read_write():
+                raise charm_refresh.PrecheckFailed("mongos is not able to read/write")
+            return
+
+        backup_state = self.dependent.backup_manager.backup_state()
+
+        if backup_state in (BackupState.BACKUP_RUNNING, BackupState.RESTORE_RUNNING):
+            raise charm_refresh.PrecheckFailed("Backup/Restore in progress.")
```
carlcsaposs-canonical: nit: could this be split into "Backup in progress" and "Restore in progress"?

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,587 @@
...
+        for attempt in Retrying(stop=stop_after_attempt(10), wait=wait_fixed(1)):
+            with attempt:
+                if not self.is_cluster_healthy():
+                    raise ClusterNotHealthyError()
+
+    def is_cluster_healthy(self) -> bool:
+        """Returns True if all nodes in the replica set / cluster are healthy."""
+        if not self.dependent.mongo_manager.mongod_ready():
+            logger.error("Cannot proceed with refresh. Service mongod is not running.")
+            return False
```
carlcsaposs-canonical: this is currently being retried in wait_for_cluster_healthy when it will never succeed

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,587 @@
...
+
+        return True
+
+    def move_primary_to_last_upgrade_unit(self) -> None:
+        """Moves the primary to last unit that gets upgraded (the unit with the lowest id).
+
+        Raises FailedToMovePrimaryError
+        """
+        # no need to move primary in the scenario of one unit
+        if len(self.state.units_upgrade_peer_data) < 2:
```
carlcsaposs-canonical: question: where is `units_upgrade_peer_data` defined?

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,587 @@
...
+
+        backup_state = self.dependent.backup_manager.backup_state()
+
+        if backup_state in (BackupState.BACKUP_RUNNING, BackupState.RESTORE_RUNNING):
+            raise charm_refresh.PrecheckFailed("Backup/Restore in progress.")
+
+        self._mongodb_checks()
+
+        if not self.are_pre_upgrade_operations_config_server_successful():
+            raise charm_refresh.PrecheckFailed("Pre-refresh operations on config-server failed.")
```
carlcsaposs-canonical: can `are_pre_upgrade_operations_config_server_successful` raise PrecheckFailed with more specific messages instead? imo pre-check failures should be more specific

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,587 @@
...
+
+    def _mongodb_checks(self) -> None:
+        """MongoDB specific checks."""
+        try:
+            self.wait_for_cluster_healthy()
+        except RetryError as e:
+            logger.error(
+                "Cannot proceed with refresh. Failed to check cluster health, error: %s", e
+            )
+            raise charm_refresh.PrecheckFailed("Cluster is not healthy")
```
carlcsaposs-canonical, line 99: imo could be worth being a bit more specific in the message

fyi PrecheckFailed is designed as an exception so you can bubble it up through multiple function calls without needing to pass the message back through the function return values

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,587 @@
...
+    def are_pre_upgrade_operations_config_server_successful(self) -> bool:
+        """Runs pre-upgrade operations for config-server and returns True if successful."""
+        if not self.state.is_role(MongoDBRoles.CONFIG_SERVER):
+            return True
+
+        if (
+            isinstance(charm_api.event, charm_api.ActionEvent)
+            and charm_api.event.action == "pre-refresh-check"
+        ):
+            return True
```
carlcsaposs-canonical: we already discussed this; leaving comment for historical reference

- this is not a supported usage of charm-refresh (supported usage documented here: https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/pre-refresh-checks/)
- as such, this may break in any release of charm-refresh
- will keep this use case in mind & try to avoid breaking, but no commitments—and the stability of the public interface will have priority
- discussing how to officially support this (mid to long term) in https://chat.canonical.com/canonical/pl/moz3ynrczigb8ncktyrh1z7zro

single_kernel_mongo/core/machine_upgrades_v3.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@dataclasses.dataclass(eq=False)
+class MachineMongoDBRefresh(
+    MongoDBRefresh,
+    charm_refresh.CharmSpecificMachines,
+):
+    """Machine specific refresh code."""
+
+    dependent: OperatorProtocol
```
carlcsaposs-canonical: is this redundant since it's already specified in MongoDBRefresh?

single_kernel_mongo/core/machine_upgrades_v3.py
```diff
@@ -0,0 +1,57 @@
...
+    def refresh_snap(
+        self, *, snap_name: str, snap_revision: str, refresh: charm_refresh.Machines
+    ) -> None:
+        """Refreshes a snap."""
+        self.dependent.stop_charm_services()
+
+        if self.dependent.name == CharmKind.MONGOD:
+            try:
+                if self.dependent.charm.unit.name == self.dependent.primary_unit_name:
+                    self.step_down_primary_and_wait_reelection()
```
carlcsaposs-canonical: question: where is `step_down_primary_and_wait_reelection` defined?

single_kernel_mongo/core/machine_upgrades_v3.py
```diff
@@ -0,0 +1,57 @@
...
+            except FailedToElectNewPrimaryError:
+                logger.error("Failed to reelect primary before upgrading unit.")
+                return
+
+        revision_before_refresh = self.dependent.workload.snap_revision()
+        if snap_revision == revision_before_refresh:
+            logger.warning("current snap revision and target revision are equal")
+            return
```
carlcsaposs-canonical: suggest raising error in this case; this should never happen & if it happens there's a major bug somewhere (that we should be aware of & fix)

single_kernel_mongo/core/operator.py
```diff
@@ -215,6 +215,13 @@ def check_relation_broken_or_scale_down(self, event: RelationDepartedEvent) -> N
                 "Scaling down the application, no need to process removed relation in broken hook."
             )
 
+    @property
+    def refresh_in_progress(self) -> bool:
+        """Check if charm-refresh is currently in progress."""
+        # If charm_refresh.UnitTearDown or charm_refresh.PeerRelationNotReady
+        # we consider a refresh to NOT be in progress.
+        return not self.refresh or self.refresh.in_progress
```
carlcsaposs-canonical, line 225: I think the comment contradicts the implementation

if `self.refresh is None`, I believe `refresh_in_progress` returns `True`

carlcsaposs-canonical: if the intention is that `refresh_in_progress` returns `False` if PeerRelationNotReady or UnitTearingDown is raised:

will mongo user docs clearly state scale up & scale down not supported while refresh in progress?

if so, please confirm this tradeoff with EMs and PMs per "Alternative: Declare scaling up as not supported" at https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/peer-relation-not-ready/

also, is there any reason why scale down can't be supported (i.e. why this value can't be True if `not self.refresh`)?

single_kernel_mongo/core/vm_workload.py
```diff
@@ -186,23 +186,42 @@ def active(self) -> bool:
...
+                persists.
 
         Returns:
-            True if successfully installed. False otherwise.
+            True if successfully installed, False if errors occur and `retry_and_raise` is False.
         """
         try:
+            if not revision:
+                versions = self.load_toml_file(VERSIONS_FILE)
+                revision = versions["snap"]["revisions"][machine()]
```
carlcsaposs-canonical: if possible, suggest using charm-refresh api instead of reading from refresh_versions.toml directly: https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/remove-duplicate-hardcoded-snap/

single_kernel_mongo/core/workload.py
```diff
@@ -295,29 +307,15 @@ def get_version(self) -> str:
...
     def get_charm_revision(self) -> str:
         """Get the charm revision.
 
         Returns:
             String of charm revision
         """
         try:
-            version = Path("charm_version").read_text().strip()
+            versions = self.load_toml_file(VERSIONS_FILE)
+            version = versions["charm"].split("/")[1]
```
carlcsaposs-canonical: would discourage using charm refresh version outside of `is_compatible`—this is likely to cause headaches

also, if you're going to use it, any reason to remove the track from the version? versions are not guaranteed to be unique if you remove track from the version

single_kernel_mongo/events/database.py
```diff
@@ -137,7 +137,7 @@ def pass_hook_checks(self, event: RelationEvent) -> bool:
         if not self.charm.unit.is_leader():
             return False
 
-        if self.dependent.state.upgrade_in_progress:
+        if self.dependent.refresh_in_progress:
```
carlcsaposs-canonical, line 141: it looks like this is running on relation-changed events—is that intentional?

single_kernel_mongo/events/tls.py
```diff
@@ -106,7 +106,8 @@ def _on_tls_relation_joined(self, event: RelationJoinedEvent) -> None:
             )
             event.defer()
             return
-        if self.manager.state.upgrade_in_progress:
+
+        if self.dependent.refresh_in_progress:
             logger.warning(
                 "Enabling TLS is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
             )
```
carlcsaposs-canonical, line 113: I think this may run when the tls application is scaled up, not just when the tls relation is created

single_kernel_mongo/events/tls.py
```diff
@@ -162,7 +163,7 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
             logger.info(f"Deferring {str(type(event))}: db is not initialised")
             event.defer()
             return
-        if self.manager.state.upgrade_in_progress:
+        if self.dependent.refresh_in_progress:
             logger.warning(
                 "Enabling TLS is not supported during an upgrade. The charm may be in a broken, unrecoverable state."
             )
```
carlcsaposs-canonical: are certificate rotations not handled while a refresh is in progress? this may affect HA while a refresh is in progress if a cert is going to expire

single_kernel_mongo/managers/cluster.py
```diff
@@ -77,7 +77,7 @@ def assert_pass_hook_checks(self) -> None:
         if not self.charm.unit.is_leader():
             raise NonDeferrableFailedHookChecksError("Not leader")
 
-        if self.state.upgrade_in_progress:
+        if self.dependent.refresh_in_progress:
```
carlcsaposs-canonical: does this block HA functions? e.g. ip address change that leads to `update_config_server_db` call

single_kernel_mongo/managers/cluster.py
```diff
@@ -269,7 +269,7 @@ def assert_pass_hook_checks(self) -> None:
             raise DeferrableFailedHookChecksError(
                 "Mongos was waiting for config-server to enable TLS. Wait for TLS to be enabled until starting mongos."
             )
-        if self.state.upgrade_in_progress:
+        if self.dependent.refresh_in_progress:
```
carlcsaposs-canonical, line 272: does this block HA functions (e.g. uri change [ip address change?], secret rotation)?

single_kernel_mongo/managers/ldap.py
```diff
@@ -76,7 +76,7 @@ def assert_pass_hook_checks(self) -> None:
             raise DeferrableFailedHookChecksError("DB is not initialised")
         if self.state.is_role(MongoDBRoles.SHARD):
             raise InvalidLdapWithShardError("Cannot integrate LDAP with shard.")
-        if self.state.upgrade_in_progress:
+        if self.dependent.refresh_in_progress:
```
carlcsaposs-canonical: same HA question (cert rotation, change in urls/secrets)

single_kernel_mongo/managers/mongo.py
```diff
@@ -555,3 +555,16 @@ def get_leader_statuses(self) -> list[StatusObject]:
...
+        """Sets the mongos feature compatibility version."""
+        if self.state.is_role(MongoDBRoles.REPLICATION):
+            config = self.state.mongo_config
+        elif self.state.is_role(MongoDBRoles.CONFIG_SERVER):
+            config = self.state.mongos_config
+        else:
+            return
+        with MongoConnection(config) as mongos:
+            mongos.client.admin.command(
+                "setFeatureCompatibilityVersion", value=feature_version, confirm=True
```
carlcsaposs-canonical, line 577: discussed before but double checking: downgrades will not be supported on the 8.0 track?

if there's an intention to support downgrades in the future, this could prevent that

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -220,6 +217,42 @@ def __init__(self, charm: AbstractMongoCharm):
...
+                workload_name="MongoDB",
+                charm_name=self.charm.name,
+            )
+            refresh_class = charm_refresh.Machines
+        else:
+            upgrade_backend = KubernetesMongoDBRefresh(
+                dependent=self,
+                state=self.state,
+                workload_name="MongoDB",
+                charm_name=self.charm.name,
```
carlcsaposs-canonical, line 238: not sure if I traced this correctly, but it looks like `self.charm.name` is `mongod` instead of `mongodb-k8s`

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -220,6 +217,42 @@ def __init__(self, charm: AbstractMongoCharm):
             ExternalRequirerRelations.LDAP_CERT,
         )
 
+        # Upgrades
+        if self.substrate == Substrates.VM:
+            upgrade_backend = MachineMongoDBRefresh(
+                dependent=self,
+                state=self.state,
+                workload_name="MongoDB",
+                charm_name=self.charm.name,
```
carlcsaposs-canonical, line 229: not sure if I traced this correctly, but it looks like `self.charm.name` is `mongod` instead of `mongodb`

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -220,6 +217,42 @@ def __init__(self, charm: AbstractMongoCharm):
...
+            refresh_class = charm_refresh.Kubernetes
+
+        try:
+            self.refresh = refresh_class(upgrade_backend)  # type: ignore[argument-type]
+        except (charm_refresh.UnitTearingDown, charm_refresh.PeerRelationNotReady):
+            self.refresh = None
+        except charm_refresh.KubernetesJujuAppNotTrusted:
+            logger.error(
+                f"Application is not trusted, please run `juju trust --scope=cluster {self.app.name}"
+            )
```
carlcsaposs-canonical:
```suggestion
```
charm-refresh already logs instructions

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -220,6 +217,42 @@ def __init__(self, charm: AbstractMongoCharm):
...
+            self.refresh = None
```
carlcsaposs-canonical:
```suggestion
```
no effect since raise immediately after

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -229,12 +262,94 @@ def __init__(self, charm: AbstractMongoCharm):
...
+
+        Checks if unit is healthy and allow the next unit to update.
+        """
+        if not self.state.db_initialised:
+            return
+
+        logger.info("Restarting workloads")
+        # always apply the current charm revision's config -> no need to "migrate" configuration
+        # this charm revision's config is the one supported by the targeted workload version
+        self.dependent._configure_workloads()
```
carlcsaposs-canonical, line 343: question: is this safe to run if the compatibility checks, pre-refresh checks, or OCI container check fails?

wondering if this is can run anything potentially destructive
e.g. an incompatible version of the charm code sets the directory permissions incorrectly and we lose permissions to change the permissions back, file gets written and old contents can't be recovered, etc.

would strongly recommend running only after workload_allowed_to_start

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -229,12 +262,94 @@ def __init__(self, charm: AbstractMongoCharm):
...
+        """Post refresh checks and actions.
+
+        Checks if unit is healthy and allow the next unit to update.
+        """
+        if not self.state.db_initialised:
+            return
+
+        logger.info("Restarting workloads")
+        # always apply the current charm revision's config -> no need to "migrate" configuration
+        # this charm revision's config is the one supported by the targeted workload version
```
carlcsaposs-canonical: I don't think this is true in some cases

- on k8s, if user refreshes to incompatible oci revision (workload_allowed_to_start will be False, but hasn't been checked here)
- on machines, if snap was refreshed but next unit not allowed to refresh (potentially because uncaught exception reverted databag changes) and the user refreshes again (e.g. to rollback)—charm code will get updated to latest refresh but snap revision will be old [e.g. on a unit that is not first to refresh]

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -229,12 +262,94 @@ def __init__(self, charm: AbstractMongoCharm):
...
+        self.dependent._configure_workloads()
+        self.dependent.start_charm_services()
+
+        self.state.unit_peer_data.current_revision = self.cross_app_version_checker.version
+
+        if self.dependent.name == CharmKind.MONGOD:
+            self.dependent._restart_related_services()  # type: ignore[attr-defined]
+
+        if self.dependent.mongo_manager.is_healthy():
```
carlcsaposs-canonical: question: where is `is_healthy` defined?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -229,12 +262,94 @@ def __init__(self, charm: AbstractMongoCharm):
...
+            refresh.next_unit_allowed_to_refresh = True
```
carlcsaposs-canonical: should any of this run if workload_allowed_to_start is False?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -463,7 +563,9 @@ def update_config_and_restart(self) -> None:
                 "Invalid LDAP Query template, please update your config."
             )
 
-        if self.state.upgrade_in_progress:
+        if self.refresh is None:
+            logger.warning("Warning update_config_and_restart: Refresh could be in progress.")
+        if self.refresh_in_progress:
             logger.warning(
                 "Changing config options is not permitted during an upgrade. The charm may be in a broken, unrecoverable state."
             )
```
carlcsaposs-canonical, line 575: fyi I believe config-changed is called on ip address change
and after upgrade, regardless of if config changes

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -112,10 +111,32 @@ def __init__(self, charm: AbstractMongoCharm):
...
             self, self.workload, self.state, self.substrate, RelationNames.CLUSTER
         )
-        upgrade_backend = MachineUpgrade if self.substrate == Substrates.VM else KubernetesUpgrade
-        self.upgrade_manager = MongosUpgradeManager(
-            self, upgrade_backend, key=RelationNames.UPGRADE_VERSION.value
+        if self.substrate == Substrates.VM:
+            upgrade_backend = MachineMongoDBRefresh(
+                dependent=self,
+                state=self.state,
+                workload_name="MongoDB",
```
carlcsaposs-canonical: is the workload mongodb or mongos?

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -112,10 +111,32 @@ def __init__(self, charm: AbstractMongoCharm):
...
+                charm_name=self.charm.name,
+            )
+        else:
+            upgrade_backend = KubernetesMongoDBRefresh(
+                dependent=self,
+                state=self.state,
+                workload_name="MongoDB",
+                charm_name=self.charm.name,
```
carlcsaposs-canonical, line 130: I think `self.charm.name` is `mongos` instead of `mongos-k8s`

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -132,13 +153,41 @@ def __init__(self, charm: AbstractMongoCharm):
...
+    def _post_refresh(self, refresh: charm_refresh.Common):
+        """Post refresh checks and actions.
+
+        Checks if unit is healthy and allow the next unit to update.
+        """
+        logger.info("Restarting workloads")
+        # always apply the current charm revision's config -> no need to "migrate" configuration
+        # this charm revision's config is the one supported by the targeted workload version
+        self._configure_workloads()
+        self.start_charm_services()
```
carlcsaposs-canonical, line 182: same as comments in mongodb_operator.py

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -132,13 +153,41 @@ def __init__(self, charm: AbstractMongoCharm):
...
+        if not self.state.db_initialised:
+            refresh.next_unit_allowed_to_refresh = True
+            return
+
+        if not self.is_mongos_running():
+            logger.error("Waiting for mongos router to be ready before finalising refresh.")
+
+        if not self.upgrades_manager.is_mongos_able_to_read_write():
+            logger.error("mongos is not able to read/write after refresh")
+            raise DeferrableError("mongos is not able to read/write after refresh.")
```
carlcsaposs-canonical, line 195: if these checks succeed, should next_unit_allowed_to_refresh be set to True?

single_kernel_mongo/managers/upgrade_v3.py
```diff
@@ -0,0 +1,130 @@
...
+        if not self.refresh:
+            return [UpgradeStatuses.ACTIVE_IDLE.value]
+
+        if scope == "app" and (refresh_app_status := self.refresh.app_status_higher_priority):
+            app_status = self._convert_ops_status_to_advanced_status(refresh_app_status)
+            status_list.append(app_status)
+            return status_list
+
+        if self.refresh.in_progress and not self.refresh.next_unit_allowed_to_refresh:
+            status_list.append(UpgradeStatuses.HEALTH_CHECK_FAILED.value)
```
carlcsaposs-canonical: the status should describe what health check failed (so that the user can try to resolve the issue)

tests/charms/mongos_k8s_test_charm/actions.yaml
```diff
@@ -10,13 +10,63 @@ set-tls-private-key:
...
+
+    Must run with at least one of the parameters `=false`
+  params:
+    check-compatibility:
+      type: boolean
+      default: true
+      description: |
+        Potential of data loss and downtime
+
+        If `false`, force refresh if new version of MongoDB and/or charm is not compatible with previous version
```
carlcsaposs-canonical: is workload name mongodb or mongos?

single_kernel_mongo/managers/upgrade_v3.py
```diff
@@ -0,0 +1,130 @@
...
+        For more information: see https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/status/
+        """
+        status_list: list[StatusObject] = []
+
+        # Check if Juju app was deployed with `--trust` (needed to patch StatefulSet partition)
+        if self.state.substrate == Substrates.K8S:
+            try:
+                self.state.k8s_manager.get_pod()
+            except DeployedWithoutTrustError:
+                return [CharmStatuses.DEPLOYED_WITHOUT_TRUST.value]
```
carlcsaposs-canonical: this duplicates/overrides https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/status/#kubernetes-app-not-trusted

single_kernel_mongo/managers/upgrade_v3.py
carlcsaposs-canonical, line 1: higher priority refresh statuses (e.g. refresh incompatible or pre-check failed, where the user's next action should be to rollback) should never be overridden

currently, it appears action running statuses override higher priority refresh statuses: https://github.com/canonical/data-platform-helpers/issues/40

single_kernel_mongo/managers/upgrade_v3.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if not self.refresh:
+            return [UpgradeStatuses.ACTIVE_IDLE.value]
```
carlcsaposs-canonical: in the case of scale up while refresh in progress (disaster recovery) and leader switchover to the new unit before that unit stops raising PeerRelationNotReady (which can be awhile on machines, if things are unhealthy), this will override the app status set by the previous leader

scripts/build_lib_for_integration.sh
carlcsaposs-canonical, line 1: could this be replaced with charmcraftlocal (and thus, data-platform-workflows)?

tag_charm_edge.yaml is the only supported way to create the charm refresh compatibility tag and contains a number of safeguards
https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/charm-version/

refresh_versions.toml
```diff
@@ -0,0 +1,12 @@
+# https://canonical-charm-refresh.readthedocs-hosted.com/latest/refresh-versions-toml/
+
+charm = "v8/1.10.0"
```
carlcsaposs-canonical: `charm` should not be manually set in refresh_versions.toml

https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/refresh-versions-toml/

single_kernel_mongo/core/version_checker.py
```diff
@@ -36,19 +36,20 @@ def get_cluster_mismatched_revision_status(self) -> StatusObject | None:
...
         local_identifier = (
             "-locally built" if self.version_checker.is_local_charm(self.charm.app.name) else ""
         )
         try:
-            # This part needs some explanation: If we are running this during
-            # the pre-refresh hook that happens after the upgrade, we want to
-            # check our version against the already upgraded config server, so
+            # This part needs some explanation: On VM, if we are running this
+            # during the pre-refresh hook that happens after the upgrade, all
+            # charm codes have been refreshed.
```
carlcsaposs-canonical, line 47: is this running while a refresh is in progress?

single_kernel_mongo/core/abstract_upgrades_v3.py
Gu1nness: It was removed but I forgot to update here.
It was referring to the former peer relation.
I'll update that code.

single_kernel_mongo/core/abstract_upgrades_v3.py
Gu1nness: I can add the full error that was raied in that turn off and wait for balancer call.

single_kernel_mongo/core/machine_upgrades_v3.py
Gu1nness: Yes it is.

single_kernel_mongo/core/machine_upgrades_v3.py
Gu1nness: I had forgotten to port this call, thanks for the feedback

single_kernel_mongo/core/operator.py
Gu1nness: Because the workload has to wait until the refresh object is ready to start (as it has to check that `workload_allowed_to_start`) it felt better to consider that not having the refresh relation meant that refresh would be in progress. But maybe I'm wrong.

single_kernel_mongo/core/workload.py
Gu1nness: The version is shared between the applications in a multi-application cluster for local application.
It's heavily needed to maintain cluster stability and safety checks.
No reasons to remove the track no, I'll add it back.

single_kernel_mongo/core/operator.py
carlcsaposs-canonical: > it felt better to consider that not having the refresh relation meant that refresh would be in progress. But maybe I'm wrong.

oh yes definitely—better to assume a refresh might be in progress

I got tripped up by the comment—the comment says it will be considered that a refresh is not in progress, where the code seems to consider that a refresh is in progress

single_kernel_mongo/events/database.py
Gu1nness: Yes, we don't support adding/updating new client application during an upgade

single_kernel_mongo/events/tls.py
Gu1nness: This might be a non problem as we're migrating to v4 in another PR (+ we don't scale up certificate application) but ok, noted.
If that's still and issue after the V4 migration I'll come back to it.

single_kernel_mongo/core/workload.py
carlcsaposs-canonical: > The version is shared between the applications in a multi-application cluster for local application.
> It's heavily needed to maintain cluster stability and safety checks.

do the mongo charms require all charms to be on an identical version?

I can see how there might be a requirement on the workload version (e.g. for mysql router, mysql router has to be >= mysql server), but requiring the charm versions to be identical I think might be incompatible with product requirements/expectations and can cause more issues than it solves

but seems like that assumption was already made in mongo 6 so far out of scope of this PR—but would suggest maybe reconsidering this approach

single_kernel_mongo/core/operator.py
Gu1nness: I'll update the comment it's a miss :)

single_kernel_mongo/events/database.py
carlcsaposs-canonical: could relation-changed be fired in any other cases? (e.g. ip change, scale up/down of client application, update-status on client, etc.)

single_kernel_mongo/events/tls.py
Gu1nness: In the current implementation no.
The initial implementation of refreshes blocks almost everything during an upgrade not to affect cluster stability and I think this falls into that scenario.
This is also the case on other charms that integrated charm-refresh v3 like etcd.

single_kernel_mongo/managers/cluster.py
Gu1nness: If refresh is in progress, not all applications have the same version so we rather not change any kind of config if possible.
Let me ask @Mehdi-Bendriss first.

single_kernel_mongo/managers/cluster.py
Gu1nness: If refresh is in progress, not all applications have the same version so we rather not change any kind of config if possible.
Let me ask @Mehdi-Bendriss first.

single_kernel_mongo/managers/ldap.py
Gu1nness: Less important as it's ldap users authentication but yes

single_kernel_mongo/managers/mongo.py
Gu1nness: There's no plan to support downgrades on 8.0 track, confirmed with @Mehdi-Bendriss

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: No, it will be `mongodb-k8s` for the charm `mongodb-k8s`
You must be confusing it with the operator name (not the same)

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: No, it will be mongodb-k8s for the charm mongodb-k8s
You must be confusing it with the operator name (not the same)

single_kernel_mongo/managers/mongodb_operator.py
carlcsaposs-canonical: ah yes that's what I mixed it up with; thanks!

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Updated adding the check on workload allowed to start.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: I can add that check yes.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Small typo, thanks for the warning! It's mongod_ready.

single_kernel_mongo/events/tls.py
carlcsaposs-canonical: Letting a certificate expire also affects cluster stability

I think etcd handles cert rotations while a refresh is in progress: https://github.com/canonical/charmed-etcd-operator/pull/139#discussion_r2290079550

Gu1nness: Oh indeed, I had misunderstood their code.
I'll check to do something better.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Updated that function to prevent all the code from running if it's not allowed to start.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: In that case the event is deferred so we're safe and we don't lose it.
All important configuration is duplicated in the databag to ensure that we don't use an invalid value/update the used value until we're in a safe state.

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: It is mongos, I'll update that.

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Definitely !

single_kernel_mongo/managers/upgrade_v3.py
Gu1nness: Similar layer to what is done in charmed-etcd.

tests/charms/mongos_k8s_test_charm/actions.yaml
Gu1nness: mongos

single_kernel_mongo/managers/upgrade_v3.py
Gu1nness: We should never end up in this situation hopefully (because we raise on the operator) but I kept it as a safeguard.

single_kernel_mongo/managers/upgrade_v3.py
Gu1nness: This is by design, and has been required by EMs/PMs so that if the user decide to run an action even while an upgrade is in progress, they can have immediate feedback on what is running in the action.

single_kernel_mongo/managers/upgrade_v3.py
Gu1nness: This layer of compatibility is copied from Charmed ETCD where it was already approved.
It's not perfect but it works and has already been approved.
If there are huge issues to fix, we can fix it but let's sync with Etcd as well.

scripts/build_lib_for_integration.sh
Gu1nness: We're not building charms that are released here yet (for many reasons).
In the future yes.
For now, we can't use `tag_charm_edge.yaml` as this codebase is released as a library the sematic tagging of pip packages fails if there are also some tags with the charm version syntax (I tried, it failed).

So in the future yes, for now, no

refresh_versions.toml
Gu1nness: This is only for unit testing.
I tried as much as I could to have mocking that would prevent me from using a toy file but I could not.

single_kernel_mongo/core/version_checker.py
Gu1nness: Yes it does and that's why we have to use this tiny hack.

single_kernel_mongo/core/abstract_upgrades_v3.py
Gu1nness: That's true and I'll move things around so that it does not happen anymore.

single_kernel_mongo/managers/upgrade_v3.py
carlcsaposs-canonical: > This is by design, and has been required by EMs/PMs

I think this deviates from the agreement in DA147

> they can have immediate feedback on what is running in the action

I think if the refresh is incompatible or a precheck failed, it's more important that the user rolls back instead of getting feedback about the action. Also, would these actions be allowed while a refresh is in progress?

single_kernel_mongo/managers/upgrade_v3.py
carlcsaposs-canonical: I think it's worth telling the user what failed (e.g. network issue) so that they have some guidance on how to recover (or whether to wait for the situation to recover by itself)

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -435,11 +443,11 @@ def move_primary_to_last_upgrade_unit(self) -> None:
         Raises FailedToMovePrimaryError
         """
         # no need to move primary in the scenario of one unit
-        if len(self.state.units_upgrade_peer_data) < 2:
+        if len(self.state.peer_units) < 2:
```
carlcsaposs-canonical: where is peer_units defined?

single_kernel_mongo/core/vm_workload.py
```diff
@@ -202,8 +201,7 @@ def install(self, revision: str | None = None, retry_and_raise: bool = True) ->
         """
         try:
             if not revision:
-                versions = self.load_toml_file(VERSIONS_FILE)
-                revision = versions["snap"]["revisions"][machine()]
+                revision = charm_refresh.Machines().pinned_snap_revision
```
carlcsaposs-canonical: the refresh class must not be instantiated more than once https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/instantiate/machines/#where

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -347,8 +345,12 @@ def _post_refresh(self, refresh: charm_refresh.Common):  # noqa: C901
         if self.dependent.name == CharmKind.MONGOD:
             self.dependent._restart_related_services()  # type: ignore[attr-defined]
 
-        if self.dependent.mongo_manager.is_healthy():
-            refresh.next_unit_allowed_to_refresh = True
+        if self.dependent.mongo_manager.mongod_ready():
+            try:
+                refresh.wait_for_cluster_healthy()
```
carlcsaposs-canonical: is wait_for_cluster_healthy defined on `refresh`?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -229,12 +258,100 @@ def __init__(self, charm: AbstractMongoCharm):
...
+
+        if (
+            self.state.is_role(MongoDBRoles.CONFIG_SERVER)
+            and not self.cross_app_version_checker.are_related_apps_valid()
+        ):
+            # Early return if not all apps are valid.
+            return
+
+        try:
+            self.refresh._charm_specific.wait_for_cluster_healthy()  # type: ignore[attr-defined]
```
carlcsaposs-canonical: accessing private attributes in `charm-refresh` is not supported

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -132,13 +153,46 @@ def __init__(self, charm: AbstractMongoCharm):
...
+
+        refresh.next_unit_allowed_to_refresh = True
```
carlcsaposs-canonical, line 197: if mongos is not running, next_unit_allowed_to_refresh can be set to True

refresh_versions.toml
carlcsaposs-canonical: I think it should be possible to mock

some approaches:
https://github.com/canonical/mysql-router-operators/blob/fc0cd58d66c9fa17056ecbf4c6c83fc611d8bfe5/kubernetes/tests/unit/conftest.py#L28
https://github.com/canonical/postgresql-operator/blob/d38d7656916aad2ed07e62820ff0a35deffa6a11/tests/unit/conftest.py#L42

single_kernel_mongo/core/version_checker.py
carlcsaposs-canonical: I'm not sure if I'm understanding the code correctly

if you follow the refresh instructions, will this not display a blocked status saying "not up-to-date" on all non-upgraded units when you refresh a shard app?

scripts/build_lib_for_integration.sh
carlcsaposs-canonical: you're likely to run into issues and charm-refresh may completely break—tag_charm_edge.yaml and charm-refresh-build-version have safeguards to ensure the tag is created correctly and refresh_versions.toml is in the format that charm-refresh expects

tag_charm_edge.yaml is the only supported way to create the refresh compatibility tag
and charm-refresh-build-version is the only supported way to write `charm` in refresh_versions.toml

single_kernel_mongo/managers/upgrade_v3.py
Gu1nness: it has been decided after DA147 was fully approved, it's present in [DA161](https://docs.google.com/document/d/1Yg7w7N-S1STbluk3SttZCQx1waZW_e_yOuKWeyahy20) though.
For now there are no actions that are allowed to run during an upgrade but that could change in the future and this is why this `is_action` was added, not to my decision but required by @delgod .

scripts/build_lib_for_integration.sh
Gu1nness: So because I have needs that don't fit with the overly specific ways to generate the correct charm version I can't use charm-refresh ?
Or would it be possible to have a specified format in the documentation so that people can have alternative implementations.
What would fix this would be to update [this script](https://github.com/canonical/data-platform-workflows/blob/main/_cli/data_platform_workflows_cli/create_semantic_version_tag.py#L21) so that  the filter on git-describe excludes `/` from the tags and then I could do what you want. But in the meanwhile I can't so I found a workaround (AGAIN only for test charms) that generates the right versions for all the charms so I can use charm-refresh.

refresh_versions.toml
Gu1nness: Yes but it goes further than just the pinned charm revision.
It goes for the name of the workload, and for everything.
I'll give it another try but I'm not sure that it will work.

single_kernel_mongo/core/version_checker.py
Gu1nness: Well because they are not up to date (refresh is in progress) so yes it will display that they are not up to date yet until they are fully upgraded.

single_kernel_mongo/core/abstract_upgrades_v3.py
Gu1nness: Typo again -_-

single_kernel_mongo/core/vm_workload.py
Gu1nness: Rolled back that part because otherwise it was not working and I don't want to have to inject my Machine refresh class in the workloads.
I'll go with the toml file parsing for the initial install.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: NO it's in the cluster specific code you are right.

scripts/build_lib_for_integration.sh
carlcsaposs-canonical: do you have a requirement that all the charms need to have the same refresh compat version?

if not, `refresh_versions.toml` shouldn't be present in this repo and the charm refresh compat version can be created on each of the charm repos using the standard workflows

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Yes well I need to share some logic between pre refresh checks and post refresh checks so at some point it has to live somewhere where I can call it and because of the way charm-refresh is developed it cannot live where I would put it by default (the upgrade manager) because the upgrade manager requires the `Machine / Kubernetes` class to be instantiated (and the pre-refresh logic runs in the `__init__` of that class).

It's not that I want to use private attributes TBH, I don't enjoy having to do it but at some point I have to live with the way charm-refresh is implemented, which really makes the implementation hard.

I'll try to find another way to implement that but really, charm-refresh doesn't make it easy.

single_kernel_mongo/core/version_checker.py
carlcsaposs-canonical: I don't think that's a good UX if the user is following the instructions

blocked status on a unit implies that a user should take action on that unit
where they should be taking action on the app (or no action at all)

if the user is following the happy path, it's expected that these statuses are displayed: https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/status/#unit-lower-priority

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: fixed.

single_kernel_mongo/managers/mongodb_operator.py
carlcsaposs-canonical: you can add any fields you want to the charmspecific class: https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/add-charm-specific-fields/

so if you want to share code between the charmspecific calls and your other charm code, it should be fairly straightforward without using charm-refresh's private internal interface

single_kernel_mongo/core/version_checker.py
Gu1nness: However, charm-refresh is about one single app refresh and we're dealing with multiple app refreshes.
It's important to display which applications are still outdated and are going to be refresh.
The blocked status is displayed before the refresh starts (and hence requires an action) and will disappear when the refresh finishes (when there's no action left to take).
It feels to me to be a sensible way of exposing the current status of the cluster during an upgrade.

scripts/build_lib_for_integration.sh
Gu1nness: > do you have a requirement that all the charms need to have the same refresh compat version?

In multi cluster application, this is expected yes, and checked.
If they don't, the statuses will be blocked and many actions will be prevented from happening.
We check for revision parity in the cluster when coming form charmhub, and local charm version when using local deployment.
So yes we really need this. Also, those charms that we're building, the 4 test charms are using charm-refresh (because they are plain copies of the 4 charms that are deployed) and won't start unless each one has its `refresh_versions.toml`.

single_kernel_mongo/core/version_checker.py
carlcsaposs-canonical: I think it would be sensible to display an app status before the refresh starts if the app is outdated

I don't think it aligns with Juju's UX to display a blocked status (while a refresh is in progress) if the user should not take any action (i.e. if pause-after-unit-refresh is none or pause-after-unit-refresh is first and the refresh has already been resumed): https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/status/#in-progress-no-pause

scripts/build_lib_for_integration.sh
carlcsaposs-canonical: to check my understanding, you're checking that charm revision is identical between different charms? if so, how are you syncing revisions in charmhub? it looks like mongodb, mongodb-k8s, mongos, and mongos-k8s all have different revisions on stable and edge right now

carlcsaposs-canonical: I think I may have misunderstood initially

on the actual charm repos, will tag_charm_edge.yaml and release_charm_edge.yaml be used?
and on this repo, the `charm` value in refresh_versions.toml is only used for testing?

if so, I think there wouldn't be any issues as long as `refresh_versions.toml` is only present in the test charms & not at the top level of this repo (besides issues from possible differences in testing on this repo and the charm repos themselves)

Gu1nness: > to check my understanding, you're checking that charm revision is identical between different charms? if so, how are you syncing revisions in charmhub? it looks like mongodb, mongodb-k8s, mongos, and mongos-k8s all have different revisions on stable and edge right now

Between multiple `mongodb` / `mongodb-k8s` apps yes (each one with a different role).

> on the actual charm repos, will tag_charm_edge.yaml and release_charm_edge.yaml be used?
and on this repo, the charm value in refresh_versions.toml is only used for testing?

Yes.

> if so, I think there wouldn't be any issues as long as refresh_versions.toml is only present in the test charms & not at the top level of this repo (besides issues from possible differences in testing on this repo and the charm repos themselves)

I'll try to make that happen (removing the top level `refresh_versions.toml` if I can find a solution for that !

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: I've updated the code to have something that works and does not require private access.

single_kernel_mongo/events/database.py
Gu1nness: It should not as data interfaces prevents the client from updating it's requests after the initial one.

single_kernel_mongo/events/tls.py
Gu1nness: I'll update that.

single_kernel_mongo/managers/cluster.py
Gu1nness: I'll update that after we talked with Mehdi

single_kernel_mongo/managers/cluster.py
Gu1nness: I'll update that after we talked with Mehdi

single_kernel_mongo/managers/ldap.py
Gu1nness: I'll update that.

single_kernel_mongo/events/database.py
carlcsaposs-canonical: If the client is scaled up I think juju will automatically fire relation-changed after relation-joined

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,595 @@
...
+        """Return a mongos configuration for the sharded cluster."""
+        return (
+            self.state.mongos_config
+            if self.state.is_role(MongoDBRoles.CONFIG_SERVER)
+            else self.state.mongos_config_for_user(
+                OperatorUser, hosts=set(self.state.shard_state.mongos_hosts)
+            )
+        )
+
+    def are_shards_healthy(self, mongos_config: MongoConfiguration) -> bool:
```
Mehdi-Bendriss: Shouldn't all these methods be in a manager?

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,595 @@
...
+        return self.state.mongodb_config_for_user(
+            OperatorUser, parsed_ips, replset=shard_entry[SHARD_NAME_INDEX]
+        )
+
+    def get_random_write_and_collection(self) -> tuple[str, str, str]:
+        """Returns a tuple for a random collection name and a unique write to add to it."""
+        choices = string.ascii_letters + string.digits
+        collection_name = "collection_" + "".join([secrets.choice(choices) for _ in range(32)])
+        write_value = "unique_write_" + "".join([secrets.choice(choices) for _ in range(16)])
+        db_name = "db_name_" + "".join([secrets.choice(choices) for _ in range(32)])
```
Mehdi-Bendriss: nit, maybe an uuidv4 makes sense here?

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,595 @@
...
+        secondary_config: MongoConfiguration,
+    ):
+        """Returns True if the replica contains the expected write in the provided collection."""
+        secondary_config.hosts = {host}
+        with MongoConnection(secondary_config, direct=True) as direct_secondary:
+            db = direct_secondary.client[db_name]
+            test_collection = db[collection]
+            query = test_collection.find({}, {WRITE_KEY: 1})
+            if query[0][WRITE_KEY] != expected_write_value:
+                raise ClusterNotHealthyError
```
Mehdi-Bendriss: I think we can move this directly in `is_write_on_secondaries` and return False instead of raising?

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -0,0 +1,595 @@
...
+        Raises FailedToMovePrimaryError
+        """
+        # no need to move primary in the scenario of one unit
+        if len(self.state.units) < 2:
+            return
+
+        with MongoConnection(self.state.mongo_config) as mongod:
+            unit_with_lowest_id = self.state.reverse_order_peer_units[-1]
+            unit_host = self.state.peer_unit_data(unit_with_lowest_id).internal_address
+            if mongod.primary() == unit_host:
```
Mehdi-Bendriss: nice

single_kernel_mongo/core/machine_upgrades_v3.py
```diff
@@ -0,0 +1,54 @@
...
+        """Refreshes a snap."""
+        self.dependent.stop_charm_services()
+
+        if self.dependent.name == CharmKind.MONGOD:
+            try:
+                if self.dependent.charm.unit.name == self.dependent.primary_unit_name:
+                    self.dependent.mongo_manager.step_down_primary_and_wait_reelection()
+            except FailedToElectNewPrimaryError:
+                logger.error("Failed to reelect primary before upgrading unit.")
+                return
```
Mehdi-Bendriss, line 34: should we not restart the charm services in this case?

single_kernel_mongo/core/vm_workload.py
```diff
@@ -186,23 +185,42 @@ def active(self) -> bool:
...
             self.mongod_snap.hold()
         except snap.SnapError as err:
-            logger.error(f"Failed to install {self.snap.name}. Reason: {err}.")
-            raise WorkloadNotReadyError("Failed to install mongodb")
+            logger.error(f"Failed to install {SNAP_NAME}. Reason: {err}.")
+            if retry_and_raise:
+                raise WorkloadNotReadyError("Failed to install mongodb")
+            return False
+        else:
+            return True
```
Mehdi-Bendriss: nit: for readability, we could perhaps return after the snap.hold

single_kernel_mongo/events/tls.py
```diff
@@ -162,7 +163,9 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
             logger.info(f"Deferring {str(type(event))}: db is not initialised")
             event.defer()
             return
-        if self.manager.state.upgrade_in_progress:
+        # Check if refresh is in progress and this is the initial integration, delay.
+        # Otherwise it's a rotation and we're safe to continue.
```
Mehdi-Bendriss, line 167: Should this comment be set a few lines below before the try?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -229,12 +258,100 @@ def __init__(self, charm: AbstractMongoCharm):
...
+
+        if self.dependent.name == CharmKind.MONGOD:
+            self.dependent._restart_related_services()  # type: ignore[attr-defined]
+
+        if self.dependent.mongo_manager.mongod_ready():
+            try:
+                self.upgrade_backend.wait_for_cluster_healthy()
+                refresh.next_unit_allowed_to_refresh = True
+            except RetryError:
+                return
```
Mehdi-Bendriss, line 357: Could you explain how is this case handled? in the event where the cluster is unable to stabilize its health

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -463,7 +565,9 @@ def update_config_and_restart(self) -> None:
                 "Invalid LDAP Query template, please update your config."
             )
 
-        if self.state.upgrade_in_progress:
+        if self.refresh is None:
+            logger.warning("Warning update_config_and_restart: Refresh could be in progress.")
```
Mehdi-Bendriss: should we not return here? if refresh is none

Mehdi-Bendriss (commented):
Thank you Neha! Really good work, I only have minor comments and about the methods placements (manager), otherwise I'm up-to-date with the decisions taken here.

And thank you @carlcsaposs-canonical for the thorough review!

single_kernel_mongo/core/abstract_upgrades_v3.py
Gu1nness: The big issue is: I wanted to use the `UpgradesManager` to do that, but I can't. Reason is the `UpgradesManager` is set using the `refresh` object, which is the one running all this logic in it's init.
I can probably move all this in a different manager for upgrade checks but semantically speaking, it bugs me a bit?

single_kernel_mongo/core/abstract_upgrades_v3.py
Gu1nness: I'm not sure it would be accepted by MongoDB as valid db_name, / collection_name but I can try.

single_kernel_mongo/core/machine_upgrades_v3.py
Gu1nness: I would say no: If we failed to elect a primary in time, it means that the service is probably overloaded by writes, and this would probably add even more load on the cluster ?

single_kernel_mongo/events/tls.py
Gu1nness: This comment explains the check right after: "are we in progress and is this the initial integration ?" so I think it lives in the right place.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: If the cluster is unable to stabilize its health, a message will be displayed so that the operator decides if they want to rollback or continue.
Until the cluster stabilizes, the upgrade will not be allowed to continue.
I should add a log here however to help the operator understanding what's going on.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: I think we want to defer because otherwise we would lose the event.
However, after reading the code, this line is useless as `self.refresh_in_progress` considers that not having the `self.refresh` object indicates a refresh could be in progress.
I'll remove this line.

single_kernel_mongo/core/abstract_upgrades_v3.py
Gu1nness: Fixed.

Mehdi-Bendriss (dismissed):

single_kernel_mongo/managers/mongodb_operator.py
carlcsaposs-canonical: ^ on machines

if the charm assumes that the installed charm code version matches the installed workload version whenever this is called, that assumption will break in one of the most common rollback cases

would suggest at least updating comment

single_kernel_mongo/managers/upgrade_v3.py
carlcsaposs-canonical: ^

single_kernel_mongo/core/version_checker.py
carlcsaposs-canonical: ^

.github/workflows/tag_refresh_v3.yaml
carlcsaposs-canonical, line 1: looks like this is unused, can it be removed?

scripts/build_lib_for_integration.sh
```diff
@@ -4,12 +4,18 @@
...
 LIB_PATH="./single_kernel_mongo"
 
 CHARMS_PATH="./tests/charms"
 
+# We need packaging to parse the version correctly
+eval $(poetry env activate)
+poetry install --only build-refresh-version
+VERSION=$(poetry version --short)
+# We compute a version based on the tag of the pip package version.
+VERSION_TAG=$(python3 scripts/tags/src/compute_tag.py --track 8 --tag $VERSION --testing)
```
carlcsaposs-canonical: could the version be set to `test/0.0.0+dirty` to
- avoid confusion
- if someone accidentally refreshes to a test charm build artifact in production, to mark the refresh as incompatible

?

Gu1nness: This would make it hard to have integration tests for upgrades against charmhub charms.
I think it's good that we keep a real version number for this kind of tests.

single_kernel_mongo/core/version_checker.py
Gu1nness: I check with @Mehdi-Bendriss and he asked to keep the statuses as is.

single_kernel_mongo/managers/upgrade_v3.py
Gu1nness: How can I know what health check failed ? That is the question.
There will be logs explaining the checks that failed.

scripts/build_lib_for_integration.sh
carlcsaposs-canonical: > I think it's good that we keep a real version number for this kind of tests.

it's already not a real version number—any version on this repo will be out of sync with the actual version number since the actual charm refresh compatibility version number reflects releases to edge

long-term, if the charms are in a monorepo, it would be possible to test with the real version number
currently, it's only possible to test the real version number in the repo containing the real charm

fyi any charm refresh compatibility version on a PR is marked as incompatible, so force-refresh-start check-compatibility=false will have to be used. The only difference is on other charms, tests running after merge can skip that (whereas here they couldn't). more details: https://github.com/canonical/mysql-router-operators/blob/5cef65a3e1608bb380d17d83daf4375f6398263d/kubernetes/tests/integration/test_upgrade.py#L122-L129

single_kernel_mongo/managers/upgrade_v3.py
carlcsaposs-canonical: in the place where `next_unit_allowed_to_refresh` is not set to `True`, (e.g. https://github.com/canonical/mongo-single-kernel-library/blob/c0ed076a3af98964873d2916e0658f9b92aa57b6/single_kernel_mongo/managers/mongodb_operator.py#L351-L357) would you know the reason you didn't set it to True?

scripts/tags/README.md
carlcsaposs-canonical, line 1: I think everything in scripts/tags is unused & can be removed

carlcsaposs-canonical (commented):
looks good except for https://github.com/canonical/mongo-single-kernel-library/pull/97#discussion_r2416195887 (which I believe contradicts Juju's UX design & unnecessarily deviates from [our standardized unit statuses](https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/status/#unit-lower-priority))

scripts/tags/README.md
Gu1nness: Done.

Mehdi-Bendriss (dismissed):
Thanks Neha! 

@carlcsaposs-canonical - thanks! I think we can reiterate on the status spec later on - perhaps in Gothenburg, but as things currently stand, this is a sensible approach for charms managing "manual" upgrades of multiple components where other apps are blind to the overall cluster state

carlcsaposs-canonical:
> @carlcsaposs-canonical - thanks! I think we can reiterate on the status spec later on - perhaps in Gothenburg, but as things currently stand, this is a sensible approach for charms managing "manual" upgrades of multiple components where other apps are blind to the overall cluster state

My suggestion would be to display https://github.com/canonical/mongo-single-kernel-library/blob/6f8c264149675ccda6ba231cd716def7da9f7283/single_kernel_mongo/config/statuses.py#L397-L400 as an app blocked status while a refresh is not in progress

While a refresh is in progress, IMO, this should not be displayed as a unit status in order to
- give the user information about the workload health (is it running) on each unit
- clearly inform the user about the workload version and charm version on each unit (since revision in `juju status` is confusing and will otherwise mislead them; and to explain the divergence in behavior between vm/k8s about charm code & workload being refreshed at different times in case something goes wrong and they need to troubleshoot a failed refresh)
- clearly inform the users what units will restart
- clearly inform users about the progress of a rollback
- avoid distracting and confusing the user with a `blocked` status when they don't need to take action (either they should pay attention to the blocked app status, blocked status on another unit, or there is no action for them to take)
- be consistent with our other charms
- inform the user when the refresh is stuck because of a juju bug (https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/status/#machines-no-upgrade-charm-event)

Here are the statuses that are getting overwrriten that I think should be displayed instead, if the workload is healthy on that unit: https://canonical-charm-refresh.readthedocs-hosted.com/latest/user-experience/status/#unit-lower-priority

carlcsaposs-canonical (approved):

Mehdi-Bendriss (approved):

PR #98, opened 2025-10-03
patch(DPE-8147): Require TLS

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

This PR implements the require TLS feature:
 * Ensures that the connection in the cluster is encrypted
 * Ensures all client connections are encrypted
 * Disallow non client connections

## 📑 Motivation and Context

CIS hardening

Manually ensured the cluster was using TLS, trying to find a way to do it automatically.

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/managers/config.py
+++ b/single_kernel_mongo/managers/config.py
@@ -340,9 +340,11 @@ class MongoConfigManager(FileBasedConfigManager, ABC):
                     "security": {"clusterAuthMode": "x509"},
                     "net": {
                         "tls": {
-                            "allowInvalidCertificates": True,
                             "clusterCAFile": f"{self.workload.paths.int_ca_file}",
                             "clusterFile": f"{self.workload.paths.int_pem_file}",
+                            "clusterAuthX509": {
+                                "attributes": f"O={self.state.get_subject_name()}",
+                            },
                         }
                     },
                 },
@@ -366,7 +368,7 @@ class MongoConfigManager(FileBasedConfigManager, ABC):
                     "tls": {
                         "CAFile": f"{self.workload.paths.ext_ca_file}",
                         "certificateKeyFile": f"{self.workload.paths.ext_pem_file}",
-                        "mode": "preferTLS",
+                        "mode": "requireTLS",
                         "disabledProtocols": "TLS1_0,TLS1_1",
                     }
                 },
--- a/single_kernel_mongo/managers/config.py
+++ b/single_kernel_mongo/managers/config.py
@@ -18,7 +18,6 @@ from typing_extensions import override
 from yaml import safe_dump, safe_load
 
 from single_kernel_mongo.config.literals import (
-    LOCALHOST,
     PBM_RESTART_DELAY,
     CharmKind,
     MongoPorts,
@@ -496,7 +495,7 @@ class MongosConfigManager(MongoConfigManager):
             return {"sharding": {"configDB": uri}}
         return {
             "sharding": {
-                "configDB": f"{self.state.app_peer_data.replica_set}/{LOCALHOST}:{MongoPorts.MONGODB_PORT.value}"
+                "configDB": f"{self.state.app_peer_data.replica_set}/{self.state.unit_peer_data.internal_address}:{MongoPorts.MONGODB_PORT.value}"
             }
         }
 
--- a/single_kernel_mongo/managers/config.py
+++ b/single_kernel_mongo/managers/config.py
@@ -339,6 +339,7 @@ class MongoConfigManager(FileBasedConfigManager, ABC):
                     "security": {"clusterAuthMode": "x509"},
                     "net": {
                         "tls": {
+                            "allowInvalidCertificates": True,
                             "clusterCAFile": f"{self.workload.paths.int_ca_file}",
                             "clusterFile": f"{self.workload.paths.int_pem_file}",
                             "clusterAuthX509": {
--- a/single_kernel_mongo/managers/mongo.py
+++ b/single_kernel_mongo/managers/mongo.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 import json
 import logging
 from typing import TYPE_CHECKING
+from urllib.parse import urlencode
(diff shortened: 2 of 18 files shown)


single_kernel_mongo/managers/config.py
```diff
@@ -340,9 +339,11 @@ def auth_parameter(self) -> dict[str, Any]:
                     "security": {"clusterAuthMode": "x509"},
                     "net": {
                         "tls": {
-                            "allowInvalidCertificates": True,
```
Mehdi-Bendriss, line 343: This would prevent invalid certificates, and from the [docs](https://www.mongodb.com/docs/manual/core/security-x.509/#std-label-client-x509-certificates-requirements), a valid certificate can be defined amongst others as:
- A single Certificate Authority (CA) must issue the certificates for both the client and the server.

Which would defeat the purpose of 2 relations with v4

single_kernel_mongo/managers/mongo.py
```diff
@@ -265,7 +272,7 @@ def add_user(self, relation: Relation):
                 return
 
             data_interface.set_endpoints(relation.id, ",".join(sorted(config.hosts)))
-            data_interface.set_uris(relation.id, config.uri)
+            data_interface.set_uris(relation.id, config.uri_without_tls)
```
Mehdi-Bendriss, line 275: Could you explain this?

single_kernel_mongo/managers/mongo.py
```diff
@@ -466,7 +472,7 @@ def process_added_units(self) -> None:
 
             for member in config_hosts - replset_members:
                 logger.debug("Adding %s to replica set", member)
-                if not self.mongod_ready(uri=member):
+                if not self.mongod_ready(uri=f"mongodb://{member}"):
```
Mehdi-Bendriss, line 475: Could you explain this, how could it work previously without the scheme?

single_kernel_mongo/managers/tls.py
```diff
@@ -252,6 +270,16 @@ def push_tls_files_to_workload(self) -> None:
         if internal_pem is not None:
             self.workload.write(self.workload.paths.int_pem_file, internal_pem)
 
+        if self.substrate == Substrates.VM:
+            return
+
+        if external_ca is not None:
+            self.state.paths.ext_ca_file.write_text(external_ca)
+            self.state.paths.ext_ca_file.chmod(600)
+        if external_pem is not None:
```
Mehdi-Bendriss: nit:
```suggestion
        if external_ca:
            self.state.paths.ext_ca_file.write_text(external_ca)
            self.state.paths.ext_ca_file.chmod(600)
        if external_pem:
```

single_kernel_mongo/state/charm_state.py
```diff
@@ -722,6 +733,10 @@ def is_shard_added_to_cluster(self) -> bool:
         if not self.config_server_name or not self.app_peer_data.mongos_hosts:
             return False
 
+        # We can't check if we don't have a valid certificate
+        if self.shard_state.internal_ca_secret is not None and not self.tls.external_enabled:
```
Mehdi-Bendriss, line 737: can this not be the case in the case where we're moving from No TLS, to adding TLS to a single sharded component?

single_kernel_mongo/state/charm_state.py
```diff
@@ -778,8 +793,9 @@ def mongodb_config_for_user(
             hosts=hosts or user.hosts,
             port=MongoPorts.MONGODB_PORT.value,
             roles=user.roles,
-            tls_external=self.tls.external_enabled,
-            tls_internal=self.tls.internal_enabled,
```
Mehdi-Bendriss, line 782: Why are we unsetting TLS internal?

tests/integration/helpers/tls.py
```diff
@@ -145,6 +180,42 @@ async def check_tls(
         return False
 
 
+async def cannot_connect_without_tls(
```
Mehdi-Bendriss, line 183: This is a matter of taste ultimately, personally I find it easier to check against and reason with positive functions than with negatives

Mehdi-Bendriss (commented):
Thank you Neha! Great work! I left a few questions

single_kernel_mongo/managers/mongo.py
Gu1nness: We're providing a URI that is to be used by the client database: we don't control where they will put the certificates, so we don't want to add the certificates paths as URI params so that we don't enforce the place where they should write it. It will be their responsibility to chose that.

single_kernel_mongo/managers/mongo.py
Gu1nness: It was working because we where never providing URI params after the host.
If we provide URI params after the host (in our case because of require TLS we MUST use certificates in case of TLS, so we have to provide a scheme to generate a valid URI.

single_kernel_mongo/state/charm_state.py
Gu1nness: Because we use require TLS, until we get a valid certificate we can't log to the config server: the config server is using TLS and we have to connect to the remote mongos so we must wait to have a certificate to be able to log in to their instance.
So until we get a valid certificate, we won't be able to check if the shard is added to the cluster.
The good thing is that we should theoretically never end up in that case, I added that as a safeguard so we don't raise later and we don't try to generate invalid URIs.
When we generate URIs with TLS parameters, it expects a valid path for the certificates and checks on the system so we protect ourselves by ensuring that the certificates are on the file system before trying to use it.

single_kernel_mongo/state/charm_state.py
Gu1nness: Because we never need that information for mongodb clients, we're never connecting as a member of the cluster, only as a client.
A client only requires external enabled, not internal enabled.

This code was added a long time ago and never used so I just cleaned it along the way.

Mehdi-Bendriss (approved):

PR #106, opened 2025-10-08
patch(DPE-8495): Enable Transparent Huge Pages if available

Gu1nness:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

MongoDB 8 strongly recommends enabling THP to have the best performances.
This enables it if available.

## 📑 Motivation and Context

Prepare to provide the best support for MongoDB 8 as a preparation of stable.

Manual testing + updated unit tests.

## ✅ Checklist


single_kernel_mongo/templates/enable-transparent-huge-pages.service.j2
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Enable Transparent Hugepages (THP)
+DefaultDependencies=no
+After=sysinit.target local-fs.target
+Before={{service_file}}
+
+[Service]
+Type=oneshot
+ExecStart=/bin/sh -c 'echo always | tee /sys/kernel/mm/transparent_hugepage/enabled > /dev/null && echo defer+madvise | tee /sys/kernel/mm/transparent_hugepage/defrag > /dev/null && echo 0 | tee /sys/kernel/mm/transparent_hugepage/khugepaged/max_ptes_none > /dev/null && echo 1 | tee /proc/sys/vm/overcommit_memory > /dev/null'
```
Mehdi-Bendriss: A few points:
1. could you apply `/proc/sys/vm/overcommit_memory` through sysctl directly? (and handle failure if any in the likes of LXD) - I'd like to keep the service file to the strict minimum
2. could you break down the the `ExecStart` into multiple `ExecStartPre` commands for readability?
3. It may make sense to add the `-` character after the equal to tolerate failures in case they don't apply, think LXD ?

Mehdi-Bendriss (commented):
Thanks Neha! Looks good! I have a comment on the service file

single_kernel_mongo/templates/enable-transparent-huge-pages.service.j2
Gu1nness: 1. I think it makes sense to keep all of it in a single file so that all operations happens in only one logical unit (otherwise it means 2 places to clean up every time, etc).
2. Sure
3. We already catch the error on startup and disable the service if it fails, do you think we should do more ?

Mehdi-Bendriss: For 1, the problem is that we alreadyapply sysctl using the sysctl charm lib for the `vm.max_map_count`: https://github.com/canonical/mongo-single-kernel-library/blob/174b9c80f44fd576d5646db7cb1b156c207ae91e/single_kernel_mongo/managers/mongodb_operator.py#L849

Gu1nness: Ok

Mehdi-Bendriss (approved):

PR #121, opened 2025-10-22
patch: Upgrades v3 - integration tests

Gu1nness:
## 🏷 Type of changes

- [x] Documentation Update
- [x] Tooling and CI changes

## 📝 Description

Introduce integration tests for refreshes v3.
This also brings some fixes to the codebase, more accurate pre-refresh checks
This also fixes a bug in the mongos charms where some charms would not startup
This also fixes an issue in the mongodb charms where the updated versions where not shared to the shards/config-server until the upgrade was finished.

## 📑 Motivation and Context

Finish refresh v3

This is the tests.

## ✅ Checklist


single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -108,6 +107,12 @@ def _mongodb_checks(self) -> None:
             logger.error("Cluster cannot read/write to replicas")
             raise charm_refresh.PrecheckFailed("Cluster is not able to read/write to replicas")
 
+        if not (
+            isinstance(charm_api.event, charm_api.ActionEvent)
+            and charm_api.event.action == "pre-refresh-check"
+        ):
+            return
```
Mehdi-Bendriss, line 117: can you add a log before the return?

single_kernel_mongo/core/operator.py
```diff
@@ -393,3 +393,10 @@ def build_local_tls_directory(self) -> None:
             return
 
         Path(self.state.paths.conf_path).mkdir(exist_ok=True)
+
+    def instantiate_keyfile(self):
+        """Instantiate the keyfile."""
+        if not (keyfile := self.state.get_keyfile()):
+            raise Exception("Waiting for leader unit to generate keyfile contents")
```
Mehdi-Bendriss, line 400: this is okay for the current PR, as it's been simply relocated from elsewhere in the code, but we should scan those in the future to have more specific exceptions

single_kernel_mongo/managers/upgrade_v3.py
```diff
@@ -242,6 +262,8 @@ def is_write_on_secondaries(
         secondary_config = copy.deepcopy(mongodb_config)
 
         for replica_ip in mongodb_config.hosts:
+            if replica_ip == self.state.unit_peer_data.internal_address:
+                continue
```
Mehdi-Bendriss, line 266: do we have the guarantee that the current unit is the primary?

Mehdi-Bendriss (commented):
Thanks Neha, great work! I left a couple of questions

single_kernel_mongo/core/operator.py
Gu1nness: Agreed, I'll create an issue for that

single_kernel_mongo/managers/upgrade_v3.py
Gu1nness: No.
However, this test runs in between refreshes, before the unit is restarted.
This means that this test fails in the case we're during a refresh.
We can't check if a refresh is in progress while this code runs because it runs in the real refresh object.
I think it's a best effort to check on every other node than our node if there's the data.

Mehdi-Bendriss (dismissed):

patriciareinoso (dismissed):

Mehdi-Bendriss (approved):

patriciareinoso (approved):

PR #122, opened 2025-10-23
patch: user renaming, TLS v4 and secret-based documentation update

patriciareinoso:
## 🏷 Type of changes

- [x] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

Update documentation for:
- Password management using secrets instead of actions
- TLS V4: enable and disable TLS, private key management using secrets instead of actions
- Rename of users to include the `charmed-` prefix

## ✅ Checklist


docs/how-to/tls/disable-tls.md
Mehdi-Bendriss, line 1: I think we could merge this one with the `Enable TLS` page no? or can we guarantee that this page gets rendered after the `enable TLS` for a logical flow?
thoughts @a-velasco ?

docs/explanation/users.md
```diff
@@ -3,22 +3,22 @@
...
 
-| user        | function                                              |
-|-------------|-------------------------------------------------------|
-| `operator`  | Admin user that manages database/cluster (i.e. admin) |
-| `monitor`   | Manages COS integration                               |
-| `backup`    | Manages all backup operations                         |
-| `logrotate` | Manages log rotation                                  |
+| user                | function                                              |
+|---------------------|-------------------------------------------------------|
+| `charmed-operator`  | Admin user that manages database/cluster (i.e. admin) |
```
Mehdi-Bendriss: Could you add that this is the internal user of the charm?

docs/how-to/tls/manage-private-keys.md
```diff
@@ -0,0 +1,64 @@
+(manage-private-keys)=
+# How to manage private keys
+
```
Mehdi-Bendriss, line 3: Could you add a note that by the charm generates private keys (similar to what you wrote for the passwords)

Mehdi-Bendriss (commented):
Thanks Patricia! Looks good, I left a few small comments

docs/how-to/tls/disable-tls.md
Gu1nness: the index makes it that enable will appear before disable

Gu1nness (dismissed):

docs/how-to/tls/disable-tls.md
patriciareinoso: This will be shown like this : (same behavior a in etcd)
image

docs/explanation/users.md
patriciareinoso: I changed the page name to internal user and add a sentence above the table to say that all of them are managed by the  charm ? is that what you meant?

docs/how-to/back-up-and-restore/migrate-a-cluster.md
```diff
@@ -34,15 +34,23 @@ old cluster
...
 
 ```shell
-juju run <app>/leader set-password username=<username> password=<new password>
+juju grant-secret <secret-name> <application-name>
+juju config <application-name> system-users=<secret_URI>
 ```
 
-Where  `<username>` is `operator`, `backup`, `monitor`, and `logrotate`. 
+```{seealso}
+[How to > Manage passwords](/how-to/manage-passwords)
```
a-velasco:
```suggestion
{ref}`manage-passwords`
```

Not all of our charm docs do this yet, but best practice is to link to internal sections with the `{ref}` directive, instead of standard markdown syntax `[]()`. It's also best to reference the target ID (i.e `(manage-passwords)=`) instead of the file path.

No need to specify custom text because it'll automatically render the title of the page :)

For future reference, if there is some place where you need custom link text, the syntax would be
```
{ref}`Custom text <manage-passwords>`
```

docs/how-to/manage-passwords.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+In this example,
+* `username-1`’s password was updated from `password-1` to `new-password-1`
+* `username-3`’s password was updated from an auto-generated password to `password-3`
+* `username-2`’s password remains as it was when the secret was added, but **`username-2` is no longer part of the secret**.
+
+See also: [Explanation > Users](/explanation/users)
+
+See also: [How to > Manage client connections](/how-to/manage-client-connections) to manage client passwords
```
a-velasco:
````suggestion
```{seealso}
* {ref}`users`
* {ref}`manage-client-connections`
```
````

File: docs/how-to/tls/disable-tls.md
Comment (a-velasco) [reply docs/how-to/tls/disable-tls.md:1]: @Gu1nness and @patriciareinoso are correct, we hardcode the order of the pages in the `index.md` file, so they are guaranteed to appear in that order.

Since there's quite a bit of content in the "Disable TLS" page, I suggest we keep it separate for now.

File: docs/how-to/tls/enable-tls.md [hunk 5]
```diff
@@ -0,0 +1,103 @@
...
+```{admonition} Caution
+:class: warning
+
+**[Self-signed certificates](https://en.wikipedia.org/wiki/Self-signed_certificate) are not recommended for a production environment.**
+
+See [this guide](https://charmhub.io/topics/security-with-x-509-certificates) for an overview of the available TLS certificates charms and how to choose the right one for your use-case.
+```
+
+Deploy the `self-signed-certificates` charm.
+```shell
```
a-velasco, line 30:
````suggestion
Deploy the `self-signed-certificates` charm.

```shell
````

File: docs/how-to/tls/enable-tls.md [hunk 6]
```diff
@@ -0,0 +1,103 @@
...
+juju deploy self-signed-certificates
+```
+
+## Enable TLS in a replica set
+Integrate your replica set with the TLS provider according the required encryption.
```
a-velasco, line 36:
```suggestion
## Enable TLS in a replica set

Integrate your replica set with the TLS provider according the required encryption.
```

a-velasco (changes requested):
Thank you @patriciareinoso! Just suggested a few small changes.

docs/how-to/tls/manage-private-keys.md
```diff
@@ -0,0 +1,68 @@
...
+By default, Charmed MongoDB 8 automatically manages private keys when TLS is enabled.
+
+If preferred, you can use [Juju secrets](https://documentation.ubuntu.com/juju/latest/reference/secret/#secret) to manage the private keys used by the charm to generate the certificate signing requests (CSR).
+
+You can manage the private keys used by the charm to generate the certificate signing requests (CSR), by storing the private key in a [juju secret](https://canonical-juju.readthedocs-hosted.com/en/latest/user/reference/secret/) and then referencing the secret in the [charm configuration](https://documentation.ubuntu.com/juju/latest/user/howto/manage-applications/#configure-an-application).
```
a-velasco:
```suggestion
You can manage the private keys used by the charm to generate the certificate signing requests (CSR), by storing the private key in a [juju secret](https://canonical-juju.readthedocs-hosted.com/en/latest/user/reference/secret/) and then referencing the secret in the [charm configuration](https://documentation.ubuntu.com/juju/latest/howto/manage-applications/index.html#configure-an-application).
```

(old URL was failing the link check)

Mehdi-Bendriss (dismissed):

Gu1nness (approved):

Mehdi-Bendriss (approved):

a-velasco (approved):

PR #133, opened 2025-11-07
patch: use secret for TLS private key management

patriciareinoso:
## 🏷 Type of changes

- [ ] Documentation Update
- [ ] Tooling and CI changes

## 📝 Description

This PR backports https://github.com/canonical/mongo-single-kernel-library/pull/96

This PR does not implement TLS V4. It only uses actions to set and get the private (client and peer) keys are removed and replaced by a config option (type: secret)

- Remove private key actions
- Add 2 config options to set the private keys (one for peer TLS and one for client TLS)

Logic:

- On config changed and secret update events:
  - The private key is read from the configuration and if the value changed from the one previously stored, a new CSR is generated and a new certificate is requested. IF the TLS relation exists.
  - If the TLS relation does not exists, key is validated, and if it's invalid, the status is reported to the user.

- On relation joined event:
  - The private key is read from the configuration (if the config option is set). Otherwise, a new private key is generated. Then, a new CSR is generated and a new certificate is requested

## 📑 Motivation and Context

Use juju secrets

Manual testing:

1. Happy path
```
juju deploy self-signed-certificates
juju deploy mongodb-k8s
juju scale-application mongodb-k8s 3
openssl genrsa -out my-key.pem 3072
juju add-secret my-secret private-key=$(base64 -w0 my-key.pem)
juju grant-secret my-secret  mongodb-k8s
juju config  mongodb-k8s tls-peer-private-key=secret:d4dgb0vmp25c7d068u30
juju integrate  mongodb-k8s:certificates self-signed-certificates
``` 
2. Invalid key: modify the `my-key.pem` to make it invalid. And set the invalid key to the configuration option: `tls-peer-private-key` and `tls-client-private-key`

3. Remove the certificates relation

Integration tests:
- Existing TLS tests

UT: 
- Given config changed event -> the private keys are properly updated and a new CSR is generated
- Given config changed event with invalid keys -> Blocked status is reported and private key is not changed
- Given config changed event with invalid keys and not TLS relation -> Blocked status is reported and private key is None

## ✅ Checklist


single_kernel_mongo/events/tls.py
```diff
@@ -64,36 +59,24 @@ def __init__(self, dependent: OperatorProtocol):
...
-            self.manager.state.is_role(MongoDBRoles.MONGOS)
-            and self.manager.state.config_server_name is None
-        ):
-            logger.info(
-                "mongos is not running (not integrated to config-server) deferring renewal of certificates."
-            )
-            event.fail("Mongos cannot set TLS keys until integrated to config-server.")
+    def _on_config_changed(self, event: ConfigChangedEvent) -> None:
+        """Validate private keys and refresh certs if needed."""
+        if self.manager.state.tls_relation is None:
```
Gu1nness, line 67: I think we should always run the update of the key, otherwise we lose that key if we relate after we set the private key.

single_kernel_mongo/events/tls.py
```diff
@@ -64,36 +59,24 @@ def __init__(self, dependent: OperatorProtocol):
...
-                "Setting TLS keys during an upgrade is not supported.",
-            )
+
+        self.manager.update_private_key_from_config(internal=True)
+        self.manager.update_private_key_from_config(internal=False)
+
+    def _on_secret_changed(self, event: ConfigChangedEvent) -> None:
+        """Validate private keys and refresh certs if needed."""
+        if self.manager.state.tls_relation is None:
             return
```
Gu1nness, line 80: Same here

single_kernel_mongo/events/tls.py
patriciareinoso: On relation joined event, the private key is read from the config (if set) or is generated (if not set), then a CSR is created and a new certificate is requested. So we are not loosing the key.

Since in V3 the management of private keys and CSRs is not done by the library as in V4, we need to handle the new CSR update manually and there are some flags around the code that we need to handle if the config option is set before the TLS relation exists. So returning on config changed event and on secret update event if the TLS relation does not exists is less complex than handling the flags around.

Gu1nness (dismissed):

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #159, opened 2025-12-19
patch: fix TICS job

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [x] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

TICS job is currently failing for mongo-single-kernel-library.

This PR enables the manual triggering of the job only on the 8/edge and 6/edge branches.
Project name is computed to match the name in the TICS dashboard:

mongo-single-kernel-library_8_edge
mongo-single-kernel-library_6_edge

Changes in the tox file:
Modify the `unit` enviroment to run unit tests on both substrates and generate a unified coverage report

## ✅ Checklist

- [x] I have added or updated any relevant documentation.

### DIFF
--- a/.github/workflows/tics_run_sh_ghaction_test.yml
+++ b/.github/workflows/tics_run_sh_ghaction_test.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Compute TICS project name
         id: tics_project
         run: |
-          BRANCH="${{ inputs.target_branch || github.ref_name }}"
+          BRANCH="${{ inputs.target_branch }}"
           PROJECT="mongo-single-kernel-library_${BRANCH//\//_}"
           echo "project=${PROJECT}" >> "$GITHUB_OUTPUT"
 
--- a/.github/workflows/tics_run_sh_ghaction_test.yml
+++ b/.github/workflows/tics_run_sh_ghaction_test.yml
@@ -3,7 +3,16 @@ name: TICS run self-hosted test (github-action)
 on:
   schedule:
     - cron: "0 2 * * 6" # Every Saturday 2:00 AM UTC
-  workflow_dispatch: # Allows manual triggering
+  workflow_dispatch:
+    inputs:
+      target_branch:
+        description: "Branch to run TICS on"
+        required: true
+        default: "8/edge"
+        type: choice
+        options:
+          - 6/edge
+          - 8/edge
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tics_run_sh_ghaction_test.yml
```diff
@@ -1,45 +1,66 @@
...
 
       - name: move results to necessary folder for TICS
         run: |
           mkdir .cover
           mv coverage.xml .cover/coverage.xml
 
+      - name: Compute TICS project name
+        id: tics_project
+        run: |
+          BRANCH="${{ inputs.target_branch }}"
```
Gu1nness: This won't work on a scheduled run I'm afraid, because you're missing the input.
You probably need one variant that uses the input if it's a workflow dispatch and the other one fetches the branch from the workflow in case of cron.

tox.ini
```diff
@@ -47,16 +47,22 @@ commands =
...
-    poetry run coverage run --source={[vars]src_path} \
-    -m pytest -v --tb native -s {posargs} {[vars]tests_path}/unit
+    poetry run coverage erase
+    # Run LXD tests
+    poetry run coverage run --append --source={[vars]src_path} \
+        -m pytest -v --tb native -s {posargs} {[vars]tests_path}/unit --substrate lxd
+    # Run MicroK8s tests
+    poetry run coverage run --append --source={[vars]src_path} \
+        -m pytest -v --tb native -s {posargs} {[vars]tests_path}/unit --substrate microk8s
+    # Generate combined report
```
Gu1nness, line 65: Hope we do get the combined report and not the second coverage run overwriting the first one?

patriciareinoso: That's the reason behind the `--append`

I also tested it running the whole lxd suite followed by only test_basics.py on microk8s and the coverage was combined

.github/workflows/tics_run_sh_ghaction_test.yml
patriciareinoso: fixed

tox.ini
Gu1nness: Ok thanks !

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #188, opened 2026-01-14
patch: improve PR generated by the bump mongo single kernel lib in charms workflow

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [x] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

- Include PAT during checkout the charm repo
- Use data platform bot user name and email in git config
- Remove the `v` prefix from the version use to update the lib version
- Automatically add us as reviewers
- Include the commit list in the PR description 
   - The commit list corresponds only to the bumping version. Ex. if we are bumping to version v.1.8.24 it contains the commits on that tag release. If the charm is not up to date and it holds v.1.8.20 version. The commit message will not show the diff between v.1.8.20 - v.1.8.24 (this can be a future improvement if we want)
- Add a sleep for 60 seconds to avoid race condition immediately after releasing to pypi
- Add commit signing using a SSH key

image

## ✅ Checklist


.github/workflows/bump_dependent_charms.yaml
```diff
@@ -1,35 +1,39 @@
...
     strategy:
       matrix:
         repo:
           - canonical/mongodb-operator
-          - canonical/mongodb-k8s-operator
-          - canonical/mongos-k8s-operator
-          - canonical/mongos-operator
+          #- canonical/mongodb-k8s-operator
+          #- canonical/mongos-k8s-operator
+          #- canonical/mongos-operator
```
Gu1nness: Remember to uncomment this

.github/workflows/bump_dependent_charms.yaml
```diff
@@ -39,15 +43,30 @@ jobs:
...
+        run: |
+          git fetch --tags
+          PREV_TAG=$(git describe --tags --abbrev=0 "${VERSION}"^)
+          VERSION_CONTENT=$(git log --pretty=format:"- %h %s" "${PREV_TAG}..${VERSION}" | sed -E 's/\(#([0-9]+)\)/[#\1](https:\/\/github.com\/canonical\/mongo-single-kernel-library\/pull\/\1)/')
+
+          {
+            echo "VERSION_CONTENT<<EOF"
+            echo "${VERSION_CONTENT}"
+            echo "EOF"
+          } >> "${GITHUB_ENV}"
```
Gu1nness: echo "VERSION_CONTENT=$((VERSION_CONTENT))" >> "$GITHUB_ENV" is a bit cleaner and easier to understand I think (also I think your version is missing a `=` ?)

.github/workflows/bump_dependent_charms.yaml
```diff
@@ -66,25 +85,32 @@ jobs:
...
+
+          git config --global gpg.format ssh
+          git config --global user.signingkey ~/.ssh/dp-bot-ssh-key
+          git config --global commit.gpgsign true
+
       - name: Commit changes
         run: |
-          git config user.name "GitHub Actions"
-          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git checkout -b "${PR_BRANCH}"
```
Gu1nness, line 128: If the branch already exists, this will fail loudly.
Maybe we should have a check / catch the error so that we can handle it properly ?

.github/workflows/bump_dependent_charms.yaml
```diff
@@ -96,8 +122,13 @@ jobs:
...
+            ${{ env.VERSION_CONTENT }}"
+
           gh pr create \
             --title "patch: Bump ${DEP_NAME} to ${VERSION}" \
-            --body "Automated bump of \`${DEP_NAME}\` to version \`${VERSION}\` after PyPI release." \
+            --body "${BODY}" \
             --base ${{ env.BASE_BRANCH }}  \
-            --head "${PR_BRANCH}"
+            --head "${PR_BRANCH}" \
+            --reviewer Mehdi-Bendriss,Gu1nness,patriciareinoso
```
Gu1nness: This list should be fetched better IMHO.
We could use CODEOWNERS for that, or a file in this repo so that we can update it easier and we don't leak the reviewers directly in the workflow.

.github/workflows/release.yaml
```diff
@@ -62,11 +62,24 @@ jobs:
...
+    name: Sleep after release to PyPI
+    runs-on: ubuntu-latest
+    needs:
+      - release-part1
+      - release-part2
+    steps:
+      - name: Sleep
+        run: |
+          echo "Waiting 60 seconds to avoid PyPI race conditions..."
+          sleep 60
```
Gu1nness, line 75: Hope this is enough. Should we find a way to have retries to specific steps of the bump_dependent_charm workflow ?

.github/workflows/bump_dependent_charms.yaml
patriciareinoso: I tested this and if the branch exists on remote, we can still checkout locally. The problem would be at push. 

I added a check to fail the workflow if the branch exists in remote. The result is like this:

Image

WDYT? 

... or we could do `git checkout -B` and then push force.

.github/workflows/release.yaml
patriciareinoso: I think we could try like this and see how it goes. Start with a very simple solution and if it does not work we find a another one. 
We could always poll/retry during the `poetry add` command until it finds the newest version (with a time out)

.github/workflows/bump_dependent_charms.yaml
patriciareinoso: I struggled with this one. Redirecting a multi-line string does not work well. I tried several stuff and this is the one that worked. 
Your proposal shows this error:
image
I think is a problem during the interpretation of the commit number as a string

.github/workflows/bump_dependent_charms.yaml
patriciareinoso: I was thinking we could use the groups defined in the repo automation. But I do not know how that works and we would need to make the update there as well. 
WDYT of this CODEOWNERS? I think it's easier adding it here than on each charm repo and using that

image

.github/workflows/bump_dependent_charms.yaml
Gu1nness: Why would you need a multiline string env is set with `var=value`.
So maybe there's some character escaping to bring, but we should be able to do it.
Something similar to what you're doing below:
```
          BODY="Automated bump of \`${DEP_NAME}\` to version \`${VERSION}\` after PyPI release.
            ### What's Changed
            ${{ steps.get-tag.outputs.VERSION_CONTENT }}"
```

Could we try something like this maybe ?
```
echo "VERSION_CONTENT=\'$((VERSION_CONTENT))\'" >> "$GITHUB_ENV"
```

.github/workflows/bump_dependent_charms.yaml
Gu1nness: I think it's good;
Automated force push looks risky, I don't want to overwrite some perfectly valid changes on a branch not yet pushed.

.github/workflows/bump_dependent_charms.yaml
Gu1nness: CODEOWNERS is perfect yes (and can probably be automated with canonical repo automation btw)

.github/workflows/release.yaml
Gu1nness: Agreed.
Let's see if it's enough in the future or not.

Gu1nness (approved):
Approved.
If we need any kind of better way of writing the REVISION_CONTENT, I would love it, but it's no blocker.

.github/workflows/bump_dependent_charms.yaml
patriciareinoso:
```
echo "VERSION_CONTENT=\'$((VERSION_CONTENT))\'" >> "$GITHUB_ENV"
``` 
-> I get the same error of before.

`var=value`
When there are multiple commits in one git tag, the content of `value` will have multiple lines. So we would need the escape the new lines there. Or process the lines into a list and then process them back when creating the PR description.

So I would rather keep it like this even though is not the most expressive code

.github/workflows/bump_dependent_charms.yaml
```diff
@@ -39,15 +43,40 @@ jobs:
...
+          } >> "${GITHUB_OUTPUT}"
+
+      - name: Get reviewers
+        id: get-reviewers
+        run: |
+          REVIEWERS=$(grep ".github/workflows/bump_dependent_charms.yaml" .github/CODEOWNERS \
+                      | awk '{for(i=2;i<=NF;i++) print substr($i,2)}' \
+                      | paste -sd "," -)
+
+          echo "REVIEWERS=${REVIEWERS}" >> "$GITHUB_OUTPUT"
```
Mehdi-Bendriss, line 72: nice

Mehdi-Bendriss (approved):
Thanks Patricia! 
for the 60sec, I'm not sure it's an entirely viable approach, as i had multiple occurrences last week where the new version was only resolved after 5-10min, but we can reiterate on this one

PR #258, opened 2026-03-19
patch(DPE-9339):  MongoDB Encryption At Rest

Gu1nness:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

This is the implementation of MongoDB encryption at rest.
It uses Vault (Vault-k8s on kubernetes) as a backend to store the encryption keys.
It uses a config option to enable encryption at rest at boot time.
It uses an action to rotate the master key.

## 🧪 Manual testing steps

:sun_with_face: Sunny test
```text
1. juju deploy <my-app> --config enable-encryption-at-rest=true
2. follow [the tutorial](https://canonical-vault-charms.readthedocs-hosted.com/en/latest/tutorial/getting_started_k8s/) to deploy and unseal vault/vault-k8s
3. juju integrate <my-app>:vault-kv vault:vault-kv
4. Profit (you can check the startup options of mongodb to ensure that it contains enableEncryption: True)
5. juju run <my-app>/<a-unit> rotate-encryption-master-key
6. Check in log file that it contains `Rotated master encryption key`.
```
:cloud_with_rain: Rainy test:
```text
1. juju deploy <my-app> --config enable-encryption-at-rest=false (the default)
2. follow [the tutorial](https://canonical-vault-charms.readthedocs-hosted.com/en/latest/tutorial/getting_started_k8s/) to deploy and unseal vault/vault-k8s
3. juju integrate <my-app>:vault-kv vault:vault-kv
4. It goes to blocked BUT does not prevent the charm from operating.
```

## 🔬 Automated testing steps

Positive checks that:
 * Deployment goes to blocked until integrated with vault
 * Integration restarts with correct options
 * Rotation works
 * Removing the relation goes to blocked.
 
Negative checks that:
 * Deploying with encryption disabled and integrating with vault goes to blocked
 * Trying to rotate master key in that scenario fails.

## ✅ Checklist


single_kernel_mongo/config/statuses.py
```diff
@@ -722,3 +722,58 @@ class PasswordManagementStatuses(Enum):
...
+    VAULT_INTEGRATED = StatusObject(
+        status="blocked",
+        message="Can't integrate with vault.",
+        check="Encryption at rest is disabled.",
+        action="Remove vault integration",
+    )
+    VAULT_NOT_INTEGRATED = StatusObject(
+        status="blocked",
+        message="Must be integrated with vault to enable encryption at rest.",
+        short_message="integrate with vault.",
```
patriciareinoso:
```suggestion
        short_message="Integrate with vault.",
```

single_kernel_mongo/workload/vault_agent_workload.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @property
+    def role_id(self) -> Path:
+        """Vault configuration file path."""
+        return Path(f"{self.etc_path}/vault/role_id")
+
+    @property
+    def role_secret_id(self) -> Path:
+        """Vault configuration file path."""
+        return Path(f"{self.etc_path}/vault/role_secret_id")
```
patriciareinoso, line 35: all of them have the same docstrings

single_kernel_mongo/workload/vault_agent_workload.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class VaultAgentWorkload(WorkloadBase, ABC):
+    """MongoDB Workload definition."""
```
patriciareinoso: fix docstring

tests/integration/helpers/vault.py
```diff
@@ -0,0 +1,316 @@
...
+        """Check if the vault unit is active."""
+        response = self.client.sys.read_health_status()
+        return response.status_code == VAULT_STATUS_ACTIVE
+
+    def unseal(self, unseal_key: str) -> None:
+        """Unseal a vault unit.
+
+        Args:
+            endpoint (str): The endpoint of the vault unit
+            ca_file_location (str): The path to the CA file
```
patriciareinoso:
```suggestion
```

tests/integration/helpers/vault.py
```diff
@@ -0,0 +1,316 @@
...
+            return
+        self.client.sys.submit_unseal_key(unseal_key)
+        logger.info("Unsealed vault unit: %s.", self.url)
+
+    async def wait_for_node_to_be_unsealed(self) -> None:
+        """Wait for the vault unit to be unsealed.
+
+        Args:
+            endpoint (str): The endpoint of the vault unit
+            ca_file_location (str): The path to the CA file
```
patriciareinoso:
```suggestion
```

patriciareinoso (commented):
extremely quick pass over the code

single_kernel_mongo/config/statuses.py
```diff
@@ -722,3 +722,58 @@ class PasswordManagementStatuses(Enum):
...
+        status="blocked",
+        message="The enable-encryption-at-rest config option is invalid.",
+        short_message="Wrong enable-encryption-at-rest config.",
+        check="Config validation failed.",
+        action="Revert config option enable-encryption-at-rest value.",
+        approved_critical_component=True,
+    )
+    VAULT_INTEGRATED = StatusObject(
+        status="blocked",
+        message="Can't integrate with vault.",
```
patriciareinoso: I find this message confusing. Even thought  `blocked` means there should be an intervention from the user. It makes me think that the charm tried to integrate and something when wrong during the process (missing information in the databag or that kind of stuff). Maybe something around invalid integration found? to align with the other status messages that we already have

single_kernel_mongo/events/lifecycle.py
```diff
@@ -156,6 +156,9 @@ def on_start(self, event: StartEvent):
             )
             event.defer()
             return
+        except DeferrableError:
+            logger.info("Waiting for vault to be integrated.")
```
patriciareinoso, line 161: Is the lack of vault relation the only deferrable error that can happen here? 
why not catching `WaitingForVaultError`? 

Here we are deferring because we need the integration since boot. Otherwise we could do what ever is necessary on relation-created/joined event instead of deferring. right?

single_kernel_mongo/managers/vault.py
```diff
@@ -0,0 +1,247 @@
...
+    def set_status(self, status: StatusObject, scope: Literal["both"] | Scope):
+        """Sets a status for scope app or unit, or both."""
+        if scope == "unit" or scope == "both":
+            self.state.statuses.set(status, component=self.name, scope="unit")
+        if not self.charm.model.unit.is_leader():
+            return
+        if scope == "app" or scope == "both":
+            self.state.statuses.set(status, component=self.name, scope="app")
+
+    def clear_statuses(self, scope: Literal["both"] | Scope):
```
patriciareinoso: why not adding "both" to `Scope`?

single_kernel_mongo/events/lifecycle.py
```diff
@@ -156,6 +156,9 @@ def on_start(self, event: StartEvent):
...
+            event.defer()
```
patriciareinoso, line 162:
```suggestion
            event.defer()
            return
```

single_kernel_mongo/events/vault.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def _on_ready(self, event: vault_kv.VaultKvReadyEvent) -> None:
+        """Handler for on ready event that starts vault."""
+        if not self.manager.assert_should_integrate():
+            self.manager.set_status(VaultStatuses.VAULT_INTEGRATED.value, scope="both")
+            return
+        self.manager.clear_statuses(scope="both")
+        # First, get the credentials from the interface
+        unit_credentials = self.interface.get_unit_credentials(event.relation)
+        secret = self.model.get_secret(id=unit_credentials)
```
patriciareinoso, line 100: what happens if this secret is updated? I do not see a secret-changed event handling.

single_kernel_mongo/managers/vault.py
patriciareinoso: ahh I see, it comes from `data_platform_helpers`.

It feels like we should be using constants for  `unit`, `app` and `both`

single_kernel_mongo/managers/config.py
```diff
@@ -250,6 +259,85 @@ def configure_and_restart(self):
                 raise
 
 
+@final
+class VaultConfigManager(FileBasedConfigManager):
+    """Config manager for mongodb-exporter."""
```
patriciareinoso:
```suggestion
    """Config manager for Vault."""
```

single_kernel_mongo/managers/config.py
```diff
@@ -250,6 +259,85 @@ def configure_and_restart(self):
...
+        return {
+            "vault_url": self.state.vault_state.vault_url or "",
+            "ca_path": f"{self.workload.paths.vault_cert}",
+            "role_id_path": f"{self.workload.paths.role_id}",
+            "role_secret_id_path": f"{self.workload.paths.role_secret_id}",
+            "token_file_path": f"{self.workload.paths.vault_token_file_path}",
+        }
+
+    def configure_and_restart(self, force: bool = False):
+        """Exposes the endpoint to mongodb_exporter."""
```
patriciareinoso: this is not mongodb_exporter

single_kernel_mongo/state/app_peer_state.py
```diff
@@ -101,12 +102,20 @@ def db_initialised(self) -> bool:
...
-            self.update({AppPeerDataKeys.DB_INITIALISED.value: json.dumps(value)})
-        else:
-            raise ValueError(
-                f"'db_initialised' must be a boolean value. Provided: {value} is of type {type(value)}"
-            )
+        self.update({AppPeerDataKeys.DB_INITIALISED.value: json.dumps(value)})
+
+    @property
+    def enable_encryption_at_rest(self) -> bool | None:
+        """Whether the db is initialised or not yet."""
```
patriciareinoso: this docstring is incorrect

single_kernel_mongo/state/vault_state.py
```diff
@@ -0,0 +1,161 @@
...
+    CERT = "vault-cert"
+    ROLE_ID = "vault-role-id"
+    ROLE_SECRET_ID = "vault-role-secret-id"  # nosec: B105
+    MOUNT_POINT = "vault-secret-mount-point"
+    NONCE = "vault-nonce"
+
+
+@final
+class VaultState:
+    """The stored state for the TLS relation."""
```
patriciareinoso:
```suggestion
    """The stored state for Vault."""
```

single_kernel_mongo/state/vault_state.py
```diff
@@ -0,0 +1,161 @@
...
+        self.unit_name = charm.unit.name.replace("/", "-")
+        self.secrets = SecretCache(charm=charm, relation=ExternalRequirerRelations.VAULT.value)
+
+    @property
+    def vault_enabled(self) -> bool:
+        """Is the vault relation enabled."""
+        return self.vault_relation is not None
+
+    def set_from(self, provider_data: VaultKvProviderSchema) -> None:
+        """Sets everything from LdapProviderData."""
```
patriciareinoso:
```suggestion
        """Sets everything from VaultKvProviderSchema."""
```

single_kernel_mongo/workload/mongodb_workload.py
```diff
@@ -46,3 +47,26 @@ def layer(self) -> Layer:
                 },
             }
         )
+
+    def mongod_command_standalone(
+        self,
+        bin_keyword: str,
+        bin_args: list[str] | None = None,
```
patriciareinoso, line 54: when is `bin_args` None? I see only one call that includes a non empty list

single_kernel_mongo/config/statuses.py
Gu1nness: Yes probably indeed !

single_kernel_mongo/events/lifecycle.py
Gu1nness: You're right.

single_kernel_mongo/managers/vault.py
Gu1nness: I think we should mainly bring `both` to the advanced statuses helpers at some point.
In the meanwhile, with respect for typing I think it's good to keep it that way.

single_kernel_mongo/events/vault.py
Gu1nness: Hum, it does not seem to be something handled by the upstream charm, and they don't document it either.
I don't know how they handle it exactly but as the library does not listen to it, probably it's not something that should happen.
I'll ask the team!

single_kernel_mongo/workload/mongodb_workload.py
Gu1nness: For now, it is not, but I prefer to assume that it could be in some future

patriciareinoso (dismissed):

single_kernel_mongo/events/vault.py
```diff
@@ -0,0 +1,163 @@
...
+    """Event Handler for the vault relation and actions."""
+
+    def __init__(self, dependent: MongoDBOperator):
+        self.dependent: MongoDBOperator = dependent
+        self.charm: AbstractMongoCharm[MongoDBCharmConfig, MongoDBOperator] = dependent.charm
+        self.manager = dependent.vault_manager
+        self.relation_name = ExternalRequirerRelations.VAULT.value
+
+        super().__init__(parent=dependent, key=self.relation_name)
+        self.interface = vault_kv.VaultKvRequires(self.charm, self.relation_name, self.charm.name)
```
Mehdi-Bendriss: nit
```suggestion
        self.kv_interface = vault_kv.VaultKvRequires(self.charm, self.relation_name, self.charm.name)
```

single_kernel_mongo/events/vault.py
```diff
@@ -0,0 +1,163 @@
...
+        try:
+            self.manager.ensures_value_is_not_updated()
+        except WaitingForLeaderError as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+            return
+        except InvalidConfigError:
+            self.manager.set_status(VaultStatuses.INVALID_CONFIG.value, scope="both")
+            return
+
+    def _on_gone_away(self, event: vault_kv.VaultKvGoneAwayEvent) -> None:
```
Mehdi-Bendriss, line 142: I have 2 questions here, 
1. Did we test the flow when the Vault relation is removed and then re-established? 
2. What is the mongodb and vault agent behaviors regarding the expiration of the token?

single_kernel_mongo/events/vault.py
```diff
@@ -0,0 +1,163 @@
...
+        """Generates a nonce for that unit and store it."""
+        self.manager.generate_nonce()
+
+    def _on_connected(self, event: vault_kv.VaultKvConnectedEvent) -> None:
+        """Handler for on connected event that requests for approle."""
+        if not self.manager.assert_should_integrate():
+            self.manager.set_status(VaultStatuses.VAULT_INTEGRATED.value, scope="both")
+            return
+        self.manager.clear_statuses(scope="both")
+        egress_subnets = self.manager.get_subnets()
```
Mehdi-Bendriss: probably worth renaming `get_subnets` to `get_egress_subnets` for clarity

single_kernel_mongo/events/vault.py
```diff
@@ -0,0 +1,163 @@
...
+    def _on_update_status(self, event: UpdateStatusEvent) -> None:
+        """Handler for on connected event that requests for approle."""
+        if not self.manager.assert_should_integrate():
+            # No need for a status here, it will all be recomputed anyway.
+            return
+        if not self.manager.state.vault_relation:
+            return
+        egress_subnets = self.manager.get_subnets()
+        nonce = self.manager.get_nonce()
+        self.interface.request_credentials(
```
Mehdi-Bendriss: Could you explain this?

single_kernel_mongo/managers/config.py
```diff
@@ -250,6 +259,85 @@ def configure_and_restart(self):
...
+        self.state = state
+        self.file = self.workload.paths.vault_config
+
+    def _render_template(self) -> str:
+        new_content = self.build_config()
+        template = jinja2.Template(VAULT_AGENT_TEMPLATE.read_text())
+        return template.render(**new_content)
+
+    @override
+    def set_environment(self):
```
Mehdi-Bendriss: missing return type hint and in configure_and_restart

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -480,6 +504,14 @@ def prepare_for_startup(self) -> None:
         if self.charm.unit.is_leader():
             self.state.statuses.clear(scope="app", component=self.name)
 
+        # If encryption at rest should be enabled, we won't start until it's ready.
+        if self.state.enable_encryption_at_rest and (data := self.state.vault_state.get()):
+            self.vault_manager.prepare_vault_agent(data)
+
+        if self.state.enable_encryption_at_rest and not self.vault_manager.is_ready():
+            self.vault_manager.set_status(VaultStatuses.VAULT_NOT_INTEGRATED.value, scope="both")
+            raise WaitingForVaultError()
```
Mehdi-Bendriss, line 532: This is not caught in the constructor during refreshes

single_kernel_mongo/managers/vault.py
```diff
@@ -0,0 +1,247 @@
...
+        if self.state.enable_encryption_at_rest != self.dependent.config.enable_encryption_at_rest:
+            return VaultConfigurationState.INVALID_CONFIG
+        if not self.state.enable_encryption_at_rest and self.state.vault_relation:
+            return VaultConfigurationState.VAULT_INTEGRATED
+        if not self.state.enable_encryption_at_rest:
+            return VaultConfigurationState.DISABLED
+        if not self.state.vault_relation:
+            return VaultConfigurationState.VAULT_NOT_INTEGRATED
+        data = self.state.vault_state.get()
+        if not data:
```
Mehdi-Bendriss:
```suggestion
        
        if not (data := self.state.vault_state.get()):
```

single_kernel_mongo/managers/vault.py
```diff
@@ -0,0 +1,247 @@
...
+            case VaultConfigurationState.DISABLED:
+                raise ImpossibleToRotateMasterKeyError(
+                    "Encryption at rest not enabled on this application."
+                )
+            case _:
+                raise ImpossibleToRotateMasterKeyError(
+                    self.map_state_to_status(vault_state).message
+                )
+
+        self.dependent.workload.stop()
```
Mehdi-Bendriss, line 351: question, assume this is not rolling restarted, what would happen if all vault workloads were to shutdown at the same time in the cluster?

single_kernel_mongo/managers/vault.py
```diff
@@ -0,0 +1,247 @@
...
+                | VaultConfigurationState.VAULT_UNREACHABLE
+                | VaultConfigurationState.VAULT_AGENT_FAILED
+            ):
+                charm_statuses["unit"].append(status)
+            case _:
+                charm_statuses["app"].append(status)
+                charm_statuses["unit"].append(status)
+        return charm_statuses[scope]
+
+    def set_status(self, status: StatusObject, scope: Literal["both"] | Scope):
```
Mehdi-Bendriss: missing return type hint in this and following and few other methods

single_kernel_mongo/events/vault.py
```diff
@@ -0,0 +1,163 @@
...
+            self.manager.state.vault_relation,
+            egress_subnet=egress_subnets,
+            nonce=nonce,
+        )
+
+    def _on_rotate_master_key(self, event: ActionEvent):
+        """Rotates the master key."""
+        try:
```
Mehdi-Bendriss, line 181: We are missing a few guardrails here, e.g: refresh or backup in progress

tests/integration/mongodb/encryption/test_encryption_invalid_scenario.py
Mehdi-Bendriss, line 1: missing docstrings in tests, and could you add more inline comments to document the flow and expectations?

tests/integration/helpers/vault.py
Mehdi-Bendriss, line 1: missing docstrings in some helpers - and could you add more inline comments to explain some of the flows?

tests/integration/mongodb/encryption/test_encryption_valid_scenario.py
Mehdi-Bendriss, line 1: Missing docstrings and some explanations inline comments here and there

tests/integration/mongodb/encryption/test_encryption_valid_scenario.py
```diff
@@ -0,0 +1,142 @@
...
+        ops_test,
+        app_name,
+        status="blocked",
+        message="Must be integrated with vault to enable encryption at rest.",
+    )
+
+
+@pytest.mark.abort_on_fail
+async def test_integration_goes_to_active(
+    ops_test: OpsTest, substrate: Substrate, vault_charm_name: str
```
Mehdi-Bendriss, line 78: can we add continuous writes on all tests from this one on? I'd like to see the write/read side of things and expectations on the data side for every Vault related operation.

tests/integration/mongodb/encryption/test_encryption_valid_scenario.py
```diff
@@ -0,0 +1,142 @@
...
+        substrate,
+        app_name,
+        status="Must be integrated with vault to enable encryption at rest.",
+    )
+    await check_app_status(
+        ops_test,
+        app_name,
+        status="blocked",
+        message="Must be integrated with vault to enable encryption at rest.",
+    )
```
Mehdi-Bendriss, line 196: Can we add 2 more tests about the restoring of the relation: 
1. after just a quick time while the token is still alive
2. after the token expiration time

and assert that the health is restored and data can be read / written

Mehdi-Bendriss (commented):
Thanks Neha, Great work! :)
I left a few comments, but my main concern is on upgrades and see how it all behaves if for ex the token expired (we'll need to followup with another PR for an integration test about it)

single_kernel_mongo/events/vault.py
Gu1nness: This is the method that is used to request the creation of the credentials for a unit.
Cf [vault-kv documentation](https://charmhub.io/vault-k8s/libraries/vault_kv).

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: In the `_post_refresh` method ? Indeed, it should an early return.
I'm changing that.

single_kernel_mongo/managers/vault.py
Gu1nness: This is not the vault workload that is restarted here, this is mongodb.
If all vault workloads were to shut down, the token would still be valid until it expires, and unless MongoDB was to restart while vault agent is shut down and the token has expired, there is no issue.
Issue would happen if mongodb was to restart while the token is expired (meaning vault has been stopped for around an hour or so).

tests/integration/mongodb/encryption/test_encryption_valid_scenario.py
Gu1nness: Well, we can't until the charm is fully integrated because MongoDB can only start after vault is integrated.
However I can add some for the next steps (key rotation, and maybe relation removal).

tests/integration/mongodb/encryption/test_encryption_valid_scenario.py
Gu1nness: Second one would take a while, the token is valid for a full hour.
I can do the first one for sure, for the second one I could for example remove the token file from the machine and then restore the relation?

single_kernel_mongo/events/vault.py
Gu1nness: 1. The flow was not tested, it is now.
2. Vault agent is responsible for rotation the token automatically. Its sole purpose is to maintain a valid token from the role that it is configured with. MongoDB needs token when it starts to get the master key and decrypt the database. Upon restart of MongoDB, it will find a valid token (because vault agent maintains one). This covers both expected restarts (which would not be the hard scenario without vault agent) and mongodb unexpected restart (because it will always find a valid token thanks to vault agent). If MongoDB does not find a valid token it will fail to start with a log message indicating that.

tests/integration/mongodb/encryption/test_encryption_valid_scenario.py
Mehdi-Bendriss: If the lieftime of the token cannot be easily tweaked by config or relation, then yes that's a sensible approach

single_kernel_mongo/managers/vault.py
Mehdi-Bendriss: Yes, my point is more about the relation is severed for a duration greater than the expiration time, how do we ever recover?

single_kernel_mongo/core/abstract_upgrades_v3.py
```diff
@@ -149,6 +149,14 @@ def run_pre_refresh_checks_before_any_units_refreshed(self):
...
             return
         if self.dependent.name == CharmKind.MONGOD:
+            if (
+                self.dependent.state.enable_encryption_at_rest
+                and not self.dependent.vault_manager.is_ready()  # type: ignore[attr-defined]
+            ):
+                logger.error(
+                    "Encryption at rest is not working properly. This must be fixed first."
+                )
+                raise charm_refresh.PrecheckFailed("Encryption at rest is not working properly.")
```
Mehdi-Bendriss: (applies to all relevant occurrences in this and other modules) 
I think we could provide a more detailed message through `vault_state()`, and make a reference to the vault agent.
Something along the lines of: `Encryption at rest may be degraded. Vault agent state: <state>. This must be fixed first.`

single_kernel_mongo/events/tls.py
```diff
@@ -131,7 +131,7 @@ def _on_tls_relation_broken(self, event: RelationBrokenEvent) -> None:
         """Handle the relation broken event."""
         state = self.manager.get_tls_management_state()
         match state:
-            case TlsManagementState.UPGRADE_IN_PROGRESS:
+            case TlsManagementState.UPGRADE_IN_PROGRESS | TlsManagementState.ENCRYPTION_NOT_ACTIVE:
```
Mehdi-Bendriss: does `ENCRYPTION_NOT_ACTIVE` imply: encryption "enabled but inactive"?

single_kernel_mongo/managers/config.py
```diff
@@ -285,11 +285,12 @@ def _render_template(self) -> str:
...
 
     @override
-    def set_environment(self):
+    def set_environment(self) -> None:
         """Write update parameters in the file."""
         current_config = {}
-        if self.file.exists():
-            current_config = hcl2.loads(self.file.read_text())
+        data = "\n".join(self.workload.read(self.file))
+        if data:
```
Mehdi-Bendriss: nit, walrus

single_kernel_mongo/managers/config.py
```diff
@@ -317,14 +318,24 @@ def build_config(self) -> dict[str, str]:
...
         if not self.state.enable_encryption_at_rest:
             return
 
-        current_config = hcl2.loads(f"{self.file}")
+        if not self.state.vault_relation:
+            return
+
+        current_config = {}
+        data = "\n".join(self.workload.read(self.file))
+        if data:
```
Mehdi-Bendriss: nit, walrus

single_kernel_mongo/events/tls.py
Mehdi-Bendriss: nvm, found it later:
```
if self.state.enable_encryption_at_rest and not self.dependent.vault_manager.is_ready():
```

WHat do you think of a name that implies "enabled but non functional/active"?

single_kernel_mongo/managers/observability.py
```diff
@@ -75,9 +82,42 @@ def __init__(
...
+        return {
+            "job_name": "vault-agent",
+            "metrics_path": "agent/v1/metrics",
+            "static_configs": [
+                {
+                    "targets": [f"{self.state.unit_peer_data.internal_address}:8200"],
+                    "labels": {
+                        "cluster": self.state.config_server_name or self.charm.app.name,
+                    },
+                    "scheme": "https",
```
Mehdi-Bendriss, line 109: thank you :)

single_kernel_mongo/events/vault.py
```diff
@@ -142,13 +154,17 @@ def _on_update_status(self, event: UpdateStatusEvent) -> None:
             return
         if not self.manager.state.vault_relation:
             return
-        egress_subnets = self.manager.get_subnets()
+        egress_subnets = self.manager.get_egress_subnets()
         nonce = self.manager.get_nonce()
```
Mehdi-Bendriss: we need to catch `ValueError` and defer(?)

single_kernel_mongo/events/vault.py
```diff
@@ -77,9 +79,9 @@ def _on_connected(self, event: vault_kv.VaultKvConnectedEvent) -> None:
             self.manager.set_status(VaultStatuses.VAULT_INTEGRATED.value, scope="both")
             return
         self.manager.clear_statuses(scope="both")
-        egress_subnets = self.manager.get_subnets()
+        egress_subnets = self.manager.get_egress_subnets()
         nonce = self.manager.get_nonce()
```
Mehdi-Bendriss: we need to catch `ValueError` and defer(?)

single_kernel_mongo/managers/vault.py
```diff
@@ -109,19 +264,32 @@ def get_nonce(self) -> str:
...
+
+        # Write the approle files to the workload fs
         self.workload.write(self.workload.paths.role_id, data.credentials["role-id"])
         self.workload.write(self.workload.paths.role_secret_id, data.credentials["role-secret-id"])
-        self.dependent.vault_config_manager.set_environment()
+        self.configure_self_signed_certificates()
+
+        # Run the connectivity checks
+        if not self._check_connectivity(data):
+            raise ValueError("Connectivity failed.")
```
Mehdi-Bendriss, line 283: This exception needs to be caught `post_refresh` and `prepare_for_startup`

tests/integration/mongodb/encryption/test_encryption_valid_scenario.py
```diff
@@ -118,25 +162,54 @@ async def test_rotate_master_key(ops_test: OpsTest, substrate: Substrate):
...
         status="blocked",
         message="Must be integrated with vault to enable encryption at rest.",
     )
+
+    # verify that no writes were skipped
+    await verify_writes(ops_test, substrate, app_name)
+
+
+@pytest.mark.abort_on_fail
+async def reintegrate_goes_to_regular(
```
Mehdi-Bendriss, line 203: This is only testing the happy path where we the Token doesn't expire, right? I think we had another test case we wanted to add, about token expiration that can be mimicked through token removal

Mehdi-Bendriss (commented):
Thanks Neha, great work! Only few comments remaining from my side

single_kernel_mongo/events/vault.py
Gu1nness: I quite dislike deferrals on update-status events because they will happen sooner or later and because of the risk of building very long defer chains. I'd rather exit with a log line

single_kernel_mongo/events/vault.py
Gu1nness: Yes, done

tests/integration/mongodb/encryption/test_encryption_valid_scenario.py
Gu1nness: I'll add this scenario as well, indeed I forgot.

single_kernel_mongo/events/tls.py
Gu1nness: `ENCRYPTION_DEGRADED`

single_kernel_mongo/events/vault.py
Mehdi-Bendriss: that would be assuming a default update-status interval, in the case where the interval is way higher we may end up stuck in a wait time for something that could have been resolved on a peer relation changed event that came earlier

Gu1nness: True.
I'll add the deferral.

Mehdi-Bendriss (approved):
Thank you Neha!

patriciareinoso (approved):

patriciareinoso (approved):

PR #299, opened 2026-04-17
patch: test rollingops

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

### Rolling Ops Integration for MongoDB

This PR introduces rolling-ops based asynchronous restarts for mongod and mongos components.
It uses the cccccccccccccccccccccccccccccccccccccccccc

The main change is:

Before: components restarted immediately within hooks
Now: restarts are queued via rolling ops (async lock) and executed in a rolling fashion.

This ensures safer, ordered restarts across units.

This PR adds the relation to etcd but fully etcd integration is not implemented.

### LDAP 

**Former behavior:** 
- Immediate restart 
- LDAP statuses recomputed after restart

**New behavior:**
- Trigger async restart via rolling ops
- LDAP statuses are set immediately and remain unchanged
- Restart applies to mongod component only
- Restart-related status is applied to MongoD, not LDAP

**Restart code sections**
1. `restart_when_ready` 
    - Triggered by `restart-if-ready` if leader or `relation-changed` if follower

2. `clean_ldap_credentials_and_uri` and `remove_ldap_certificates`:
   - Triggered on LDAP `relation-broken` or unavailable.

### TLS

**Former behavior**
- Immediate restart
- SHARD Manager / Mongos statuses recomputed after restart

**New behavior**
- Trigger async restart
- Recompute SHARD Manager / Mongos statuses immediately
- Restart applies to MongoD / Mongos components
- SHARD Manager statuses remain unchanged

**Restart code sections**
1. `enable_certificates_for_unit`  triggered on certificate available

4. `disable_certificates_for_unit` : triggered on tls relation-broken

### SHARD MANAGER

**Former behavior**
On DB created:
- If keyfile changed and cluster auth uses keyfile -> immediate restart
- If the PBM CA certificate does not exist but it exists in the trust store -> immediate restart
- If mondod is not ready -> defer

**New behavior**
On DB created:
- If keyfile changed and cluster auth uses keyfile -> async restart
- If the PBM CA certificate does not exist but it exists in the trust store -> async restart
- If a restart is pending or mondod is not ready -> defer

**Restart code sections**
1. `update_member_auth`
2. `update_pbm_certificate_in_trust_store`

### CLUSTER (mongos)

**Former behavior:** 
For mongos, on relation-changed:
- If the config changed, keyfile changed or mongos is not running ->  immediate restart
- If mongos is not running -> defer

**New behavior:**
For mongos, on relation-changed:
- If the config changed, keyfile changed or mongos is not running ->  async restart
- If a restart is pending -> defer 
- If mongos is not running -> defer

### MongoDB 
**Former behavior:** 
On config-changed:
- Immediate restart (it is actually restarted if the IPs changed)
- Continue configuration

**New behavior:**
On config-changed:
- Async restart (it is actually restarted if the IPs changed)
- Continue configuration

### Operator:
**Former behavior:** 
- Trigger immediate restart on shard relation broken
- Trigger immediate restart on s3/gsc relation broken

**New behavior:**
Same behavior kept but using async restarts

**Restart code sections**
1. `remove_ca_cert_from_trust_store`: 

## ✅ Checklist


single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -675,7 +689,7 @@ def update_config_and_restart(self) -> None:
             )
 
         # If we had an IP change, we must restart.
-        self.restart_charm_services()
+        self.rolling_restart_charm_services(force=False)
```
patriciareinoso: as discussed offline this restart should be protected by a `self.state.db_initialised` check.

context: we have a config-changed event triggered after install, meaning before start.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1216,25 +1230,51 @@ def stop_charm_services(self):
         self.workload.stop()
 
     @override
-    def restart_charm_services(self, force: bool = False):
+    def restart_charm_services(self, force: bool = False) -> OperationResult:
```
patriciareinoso: ass discussed offline. 

This callback should add guards about LDAP and vault state because restarting if we do not have the appropriate state can break the charm.

single_kernel_mongo/events/cluster.py
```diff
@@ -166,12 +165,17 @@ def _on_relation_changed(self, event: RelationChangedEvent) -> None:
         """
         try:
             self.manager.update_mongos_and_restart()
+            if self.dependent.is_waiting_for_rolling_restart():
+                msg = "Waiting for mongos to be restarted"
+                defer_event_with_info_log(logger, event, str(type(event)), str(msg))
+                return
+            self.manager._reconcile_after_mongos_restart()
```
Gu1nness: I don't think we should defer here.
I would rather have a dedicated callback or move the code around so that we don't need to do that.
The reason:
1. relation-changed => request lock
2. lock granted => execute first the deferred event that re-requests the lock then does the restart
3. We still haven't run the reconciliation but we'll need to wait until later to run that.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1216,25 +1230,51 @@ def stop_charm_services(self):
...
+            component=self.name,
+        )
+        self.rollingops_manager.request_async_lock(
+            callback_id="restart_charm_services", kwargs={"force": force}
+        )
+        logger.info("Requested and async lock to restart MongoDB.")
+
+    def is_waiting_for_rolling_restart(self) -> bool:
+        """Returns whether Mongos has pending rolling operations."""
+        return self.rollingops_manager.state.status == RollingOpsStatus.WAITING
```
Gu1nness, line 1308: Maybe this could be exposed by rolling ops as an interface ? It could make sense
`is_waiting_for_lock(callback_id)` ?

single_kernel_mongo/managers/sharding.py
Gu1nness, line 635: We're messing with the workflow here.
Workflow is:
1. Update keyfile on filesystem.
2. Restart.
3. Wait for readiness.
4. Set auth-updated so that config-server can add the shard to the cluster (BECAUSE we have restarted and we now have the same keyfile as the config-server.
5. Config server adds us and sets a flag to indicate that shard has been added to cluster.
6. Shard restarts PBM

With the new workflow:
1. Update keyfile on filesystem
2. Ask for restart
3. Wait for readiness (we're waiting for restart, of course we're raising.
4. On lock granted, we run the deferred event that defers again. We restart. We don't send auth-updated flag so config server is still waiting. We wait until next event for relation-changed to re run (it's been deferred). Which means we need 3 iterations of the relation-changed event (Including the 2 deferrals) before we can signal to the config server to add us.
5. Config server adds us.
6. We restart PBM.

So way more waiting while it should make it easier to do this kind of scenarios.

We need a dedicated callback for this one IMHO.

single_kernel_mongo/managers/tls.py
```diff
@@ -164,7 +163,7 @@ def disable_certificates_for_unit(self, internal: bool):
             self.dependent.state.update_client_ca_secrets(new_ca=None)
 
         self.delete_certificates_from_workload(internal)
-        self.dependent.restart_charm_services(force=True)
```
Gu1nness, line 167: Certificate deletion should happen inside the lock.
Otherwise in the meantime between certificate deletion and restart, if mongod restarts unexpectedly, it will fail that is a disruption of service.
Needs dedicated callback probably, or improve the restart_charm_services to do the cert setting / deletion in a systematic way (if nothing in secret, remove the file if it exists, else write the file with the data from the secret.

Gu1nness (commented):
We need to be careful in the complex relations (sharding + cluster) as they are heavily stateful and the current implementation adds unnecessary delay.
We should take advantage of the asynchronous locking to do something better where we don't have to exchange so much data because we can retry the critical path.
Eg: config-server <-> shard
```
Config server sends keyfile and request lock
On lock of config server: tries to add shard to cluster, if it fails (auth not updated yet) it asks for retry

Shard:
Upon receiving keyfile, asks for two locks (restart with keyfile, restart PBM).
On lock for restart with keyfile: restart with new config
On lock for restart PBM: check if shard has been added to cluster. If yes, start/restart PBM.

That way we rely on the asynchronicity to ensure that we end up in the correct state eventually and we have removed the flag for `auth-updated` and the one for `shard-added-to-cluster` checks.

PR #315, opened 2026-04-21
patch(DPE-9770): Race condition + handle secret events

Gu1nness:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

Two issues are solved by this PR:
 * Race condition when the Mongos K8S would try to reconcile users while it hasn't received credentials yet
 * Handling of secret changes in the cluster relation. For the TLS it would be seen in the next update-status event but that makes it faster to check. Other updates could be missed (keyfile update if it ever happens).

## 🧪 Manual testing steps

The first one is flaky so hard to reproduce and ensure it's fixed, but the root cause was exposed in [this run](https://github.com/canonical/mongo-single-kernel-library/actions/runs/24442954470/job/71413236589?pr=258#step:9:4636)

The second one can be triggered by changing the CA and checking that the CA update is taken into account.
```text
1. juju deploy <my-app>
2. …
```

## 🔬 Automated testing steps

TBD

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/managers/cluster.py
+++ b/single_kernel_mongo/managers/cluster.py
@@ -352,6 +352,29 @@ class ClusterRequirer(Object):
 
         self.dependent.share_connection_info()
 
+    def handle_secret_changed(self, secret_label: str | None) -> None:
+        """If the certificates are rotated for example, handle it immediately.
+
+        Changes in secrets do not re-trigger a relation changed event, so it is necessary to listen
+        to secret changes events.
+        """
+        if not secret_label:
+            return
+        if not (relation := self.state.mongos_cluster_relation):
+            return
+        # many secret changed events occur,only listen to the ones related to our interface
+        # with the config server.
+        cluster_extra_secret_label = f"{self.relation_name.value}.{relation.id}.extra.secret"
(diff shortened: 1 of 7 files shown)


single_kernel_mongo/managers/cluster.py
```diff
@@ -352,6 +352,29 @@ def update_mongos_and_restart(self) -> None:
...
+        cluster_extra_secret_label = f"{self.relation_name.value}.{relation.id}.extra.secret"
+        cluster_user_secret_label = f"{self.relation_name.value}.{relation.id}.user.secret"
+        if secret_label not in (cluster_extra_secret_label, cluster_user_secret_label):
+            logger.info(
+                f"Secret unrelated to this sharding relation {relation.id} is changing, ignoring event."
+            )
+            return
+
+        # This will take care of updating everything that needs updating
+        self.update_mongos_and_restart()
```
Mehdi-Bendriss, line 376: @patriciareinoso FYI for rolling restart

Mehdi-Bendriss (approved):
Thanks Neha!

patriciareinoso (approved):

PR #317, opened 2026-04-22
patch(DPE-9772): CI Green

Gu1nness:
## 🏷️ Type of changes

- [ ] Documentation update
- [x] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

 * Poetry dynamic versioning check: We broke releasing once by pushing an invalid pyproject.toml file, let's not reproduce that.
 * Race condition (exposed by release testing):
 	- TLS + Mongos => we were not checking properly if TLS and Mongos had the same CA and if mongos had received certificates. Mongos would start without internal certificate, preventing from building the cluster.
    - We were potentially running some relation-changed events too early.
 * Other bugs 
    -  Generating the config server URI for the local mongos running along with config-server would fail if there are more than 1 mongos integrated. This is fixed by remodelling the way we build the URI, ensuring we don't unnecessarily access the `mongos_cluster_relation` relation object.
    -  Integrating LDAP to Config server and config server to mongos would not show a status for invalid ldap integration because the check for the hashes would never run. This is now fixed by inverting a few instructions and a helper to set the right status on the cluster relation changed events. 
 * CI:
	- Move some tests to large runners
	- Vault is now deployed and waits allowing for errors as it ends up in error state quite often and is out of our control.

## 🧪 Manual testing steps

```text
1. Deploy config server + 1 shard + mongos.
2. Integrate config-server and shard.
3. Integrate config-server to tls and shard to tls (at least peer-certs).
4. Integrate mongos to config server and tls integration to mongos at the same time.
5. This should end up quickly.
```

## 🔬 Automated testing steps

TLS race condition test should not be flaky anymore.
New LDAP test with config-server integrated to ldap and mongos, but mongos not integrated to ldap.
Updated UT.

## ✅ Checklist

### DIFF
--- a/single_kernel_mongo/managers/config.py
+++ b/single_kernel_mongo/managers/config.py
@@ -657,13 +657,9 @@ class MongosConfigManager(MongoConfigManager):
         """The config server DB parameter."""
         # In case we are integrated with a config-server, we need to provide
         # it's URI to mongos so it can configure_and_restart to it.
-        if uri := self.state.cluster.config_server_uri:
+        if uri := self.state.config_server_uri:
             return {"sharding": {"configDB": uri}}
-        return {
-            "sharding": {
-                "configDB": f"{self.state.app_peer_data.replica_set}/{self.state.unit_peer_data.internal_address}:{MongoPorts.MONGODB_PORT.value}"
-            }
-        }
+        return {}
 
     @property
     @override
--- a/.github/workflows/ci-quick.yaml
+++ b/.github/workflows/ci-quick.yaml
@@ -14,7 +14,7 @@ jobs:
     timeout-minutes: 5
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           persist-credentials: false
       - name: Install tox & poetry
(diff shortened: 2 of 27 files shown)


single_kernel_mongo/managers/config.py
```diff
@@ -657,13 +657,9 @@ def config_server_db_parameter(self) -> dict[str, Any]:
...
         # it's URI to mongos so it can configure_and_restart to it.
-        if uri := self.state.cluster.config_server_uri:
+        if uri := self.state.config_server_uri:
             return {"sharding": {"configDB": uri}}
-        return {
-            "sharding": {
-                "configDB": f"{self.state.app_peer_data.replica_set}/{self.state.unit_peer_data.internal_address}:{MongoPorts.MONGODB_PORT.value}"
-            }
-        }
+        return {}
```
Mehdi-Bendriss, line 662: For my own understanding, how is this handled on the caller side?

Mehdi-Bendriss (approved):
Thanks Neha, good work!

single_kernel_mongo/managers/config.py
Gu1nness: It's a deep merge of dicts so this dict will not bring any modification in the generated configuration

patriciareinoso (approved):

PR #318, opened 2026-04-27
patch: use rolling ops during tear down

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

This PR integrates the synchronous locks of the rollingops library.

We use synchronous locks to remove one unit from the replica set at the time

## 🧪 Manual testing steps

Deploy a replica set with and remove units

```
unit-rs-3: 12:48:54 INFO unit.rs/3.juju-log charmlibs.rollingops._rollingops_manager:Acquiring sync lock backend stop-replset-member.
unit-rs-3: 12:48:54 INFO unit.rs/3.juju-log single_kernel_mongo.utils.rollingops:Mongo replica-set sync lock acquired.
unit-rs-3: 12:48:54 INFO unit.rs/3.juju-log single_kernel_mongo.managers.mongodb_operator:Removing 10.26.146.62 from replica set
unit-rs-0: 12:48:54 INFO juju.worker.uniter.operation ran "database-peers-relation-departed" hook (via hook dispatching script: dispatch)
unit-rs-3: 12:48:55 INFO unit.rs/3.juju-log charmlibs.rollingops._rollingops_manager:Sync lock backend stop-replset-member released.
```

## ✅ Checklist


single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -993,20 +1011,28 @@ def prepare_storage_for_shutdown(self) -> None:
...
-                    self.mongo_manager.remove_replset_member()
+                self.mongo_manager.remove_replset_member()
+
+        except TimeoutError:
+            logger.info(
+                "Timed out waiting to remove %s from replica set.",
+                self.charm.unit.name,
+            )
+        except RollingOpsSyncLockError as e:
+            logger.error("Failed to acquire sync lock to remove replica set. %s", e)
```
Mehdi-Bendriss, line 1050: Should we abandon or keep retrying?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1136,9 +1164,51 @@ def update_hosts(self) -> None:
...
 
+    def process_unremoved_units(self) -> None:
+        """Remove units from replica set."""
+        members: set[str] = set()
+        try:
+            with MongoConnection(self.state.mongo_config) as mongo:
+                members = mongo.get_replset_members() - mongo.config.hosts
+
+            if not members:
+                return
```
Mehdi-Bendriss: I think there is a bit of duplication in this method, why do we compute the diff between replset members and config hosts twice?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -182,6 +188,18 @@ def __init__(self, charm: AbstractMongoCharm[MongoDBCharmConfig, MongoDBOperator
             self.state,
             container,
         )
+
+        sync_lock_backend = MongoReplsetSyncLockBackend(self.state)
+        self.rollingops_manager = RollingOpsManager(
+            charm=charm,
+            peer_relation_name=PeerRelationNames.ROLLINGOPS_PEERS,
+            etcd_relation_name=None,
+            cluster_id="mongodb",
```
Gu1nness: It will matter for the async part, but I think this cluster_id should not be a constant.
I would vouch for a string that is generated (on the leader of the replica set or on the leader of the config server) and then shared to related applications so that they all have the same cluster id, without impacting other mongodb applications that could be related to the same etcd but that are not part of the same sharded cluster.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -993,20 +1011,28 @@ def prepare_storage_for_shutdown(self) -> None:
...
-                stop=stop_after_attempt(600),
-                wait=wait_fixed(1),
-                reraise=True,
+            self.charm.status_handler.set_running_status(
+                MongoDBStatuses.WAITING_SYNC_REMOVAL.value, scope="unit"
+            )
+            # When we remove member, to avoid issues when majority members is removed, we need to
+            # remove next member only when MongoDB forget the previous removed member.
+            with self.rollingops_manager.acquire_sync_lock(
+                backend_id="stop-replset-member",
```
Gu1nness: Nit: build a collection of backend-ids in the `single_kernel_mongo.config` module so we avoid hardcoded strings as much as possible.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1136,9 +1164,51 @@ def update_hosts(self) -> None:
         """Update the replica set hosts and remove any unremoved replica from the config."""
         if not self.state.db_initialised:
             return
-        self.mongo_manager.process_unremoved_units()
+        self.process_unremoved_units()
         self.update_related_hosts()
 
+    def process_unremoved_units(self) -> None:
```
Gu1nness, line 1185: Hum, there's many levels of nesting here which I'm not sure I enjoy.
Would it be possible to export some of the functions to the `MongoManager` ?
I would see the `get_replset_members` and the one that removes.
This way we would look like:

```
try:
    members = self.mongo_manager.get_replset_members()
	if not members:
		return
	<set_status>
	with acquire_lock:
		self.mongo_manager.remove_replset_member().
```

It also feels like we're getting twice the members, one for an early check and one in the lock.
Probably we want to do something a bit better here? Bringing it closer to the implem for `prepare_storage_for_shutdown`.

Gu1nness (changes requested):

Mehdi-Bendriss (commented):
Thanks Patricia, I have 2 comments

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: The concern was about the time that we may wait before getting the lock and the state of the replica set to potentially change in between. 

now I changed a little bit `remove_replset_member` to not update config version if the replica set member is no longer there

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: The origin of the duplication is related to possible unexpected change on the replica set due to the waiting time for the lock. 
Now `remove_replset_member` handles the case in which we try to remove the member that is no longer there

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: Indeed. I will do that when adding the integration with etcd

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: We should retry on `NotReadyError, RollingOpsSyncLockError, PyMongoError`

So we can handle a small race condition that may happen and transient errors reaching mongo.

single_kernel_mongo/managers/mongo.py
```diff
@@ -668,3 +668,21 @@ def _compute_auth_restrictions(
                 serverAddress=cidrs(self.state.peer_network().bind_addresses),
             ),
         ]
+
+    def get_replset_members(self) -> set[str]:
```
Gu1nness: Please rename, I don't think the name matches what the function means.
I think `get_unwanted_replicaset_members` is better ?

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Perfect, thanks a lot!

single_kernel_mongo/utils/rollingops.py
```diff
@@ -0,0 +1,70 @@
...
+
+class StopReplsetSyncLockBackend(SyncLockBackend):
+    """Mongo-backed sync lock for replica-set membership changes.
+
+    This does not create an external lock. It waits until MongoDB has no
+    member currently in REMOVED/REMOVING-like state before allowing the
+    critical section to proceed.
+    """
+
+    def __init__(self, state):
```
Gu1nness, line 24: nit: type hints

Gu1nness (approved):
LGTM except for the missing type hint, but already approving

single_kernel_mongo/utils/rollingops.py
patriciareinoso: ack. Already added the next PR. The CI is taking so long I do not want to push again just because of it

Mehdi-Bendriss (approved):

PR #319, opened 2026-04-27
patch: add async restarts using charmlibs-rollingops

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

### Rolling Ops Integration for MongoDB

This PR introduces rolling-ops based asynchronous restarts for mongod and mongos components.
It updates the dependencies to use https://pypi.org/project/charmlibs-rollingops/
Integration with etcd is added for VM charms

The main change is:

Before: components restarted immediately within hooks
Now: restarts are queued via rolling ops (async lock) and executed in a rolling fashion.

This ensures safer, ordered restarts across units.

- Most of the workflows that restart do not do meaningful work after the restart. In these cases, the call to `restart_charm_services` was just replace with `async_restart_charm_services`
- The `restart_charm_services` now detects changes on the TLS certificates and "forces" the restart if a change is detected. 
- When adding a member to the replica set, we do not check anymore if we can connect before adding it. This is problematic when TLS is enable and certificates has not been written to disk because restart hasn't happen. We can add hosts to the replica set without this check.
- For the complex workflows changes were made as follows:

### TLS

**Former behavior**
- Certificates are stored or removed from the secrets. 
- Certificates are stored or removed from disk.
- Immediate restart

**New behavior**
- Certificates are stored or removed from the secrets. 
- Trigger async restart
- Certificates are stored or removed from disk during restart

### SHARD MANAGER
On relation changed:
- On a keyfile change, we trigger the async restart and return the function
- Once the shard restarts, it sets the auth_updated = True
- Once it is added to the cluster by the config-manager a new config changed will be trigger and in that case we will restart PBM

### CLUSTER (mongos)
The whole `update_mongos_and_restart` now runs on a callback. 

### Cluster ID 

- Generated by the leader unit of a replica set or a config server.
- The config server shares the cluster id in the cluster relation for mongos and in the sharding relation for shards
- Stored in the `app_peer_data` as a secret

image

## 🧪 Manual testing steps

```text
1. juju deploy <my-app>
2. …
```

## ✅ Checklist


single_kernel_mongo/managers/tls.py
```diff
@@ -423,7 +487,7 @@ def get_statuses(self, scope: Scope, recompute: bool = False) -> list[StatusObje
         if tls_client_private_key_id := self.dependent.config.tls_client_private_key_id:
             if not self.update_private_key(tls_client_private_key_id, internal=False):
                 charm_statuses.append(TLSStatuses.INVALID_CLIENT_PRIVATE_KEY.value)
-
+        print(charm_statuses)
```
patriciareinoso: TO REMOVE

single_kernel_mongo/events/cluster.py
```diff
@@ -170,49 +167,14 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
...
-            self.dependent.state.statuses.add(
-                MongosStatuses.WAITING_FOR_SECRETS.value,
-                scope="unit",
-                component=self.charm.name,
-            )
-        except WorkloadServiceError:
-            # Some status was already set and a log was already displayed in
-            # `restart_charm_services`
-            return
+        self.manager.async_update_mongos_and_restart()
```
Gu1nness: Nit: update docs (requests lock to restart)
Nit-bis: Explain that all errors that could be raised are caught by the callback
Nit-ter: Can requesting a lock raise an error ? Shouldn't we catch it here?

single_kernel_mongo/managers/mongo.py
```diff
@@ -116,8 +116,13 @@ def mongod_ready(self, uri: str | None = None, direct: bool = True) -> bool:
 
         actual_uri = uri or f"mongodb://localhost:{port}"
         actual_uri = f"{actual_uri}/?{urlencode(params)}"
-        with MongoConnection(EMPTY_CONFIGURATION, actual_uri, direct=direct) as direct_mongo:
-            return direct_mongo.is_ready
+
+        try:
+            with MongoConnection(EMPTY_CONFIGURATION, actual_uri, direct=direct) as direct_mongo:
+                return direct_mongo.is_ready
+        except FileNotFoundError as e:
```
Gu1nness, line 123: This exception feels weird. Could we raise a more meaningful error ? Or at least explain what this exception corresponds to with a comment ?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1297,21 +1307,68 @@ def restart_charm_services(self, force: bool = False):
...
+
+        if self.ldap_manager.should_not_restart():
+            raise DeferrableError(
+                "Workload not allowed to restart: LDAP is in an inconsistent state."
+            )
+
+        if (
+            self.state.enable_encryption_at_rest
+            and (state := self.vault_manager.vault_state()) != VaultConfigurationState.ACTIVE
+        ):
```
Gu1nness: I'm wondering if we should move this check in a helper function in the vault manager because it's replicated in a few places.

single_kernel_mongo/config/literals.py
```diff
@@ -131,3 +131,11 @@ class RollingOpsBackend(StrEnum):
...
 
     STOP_REPLSET_MEMBER = "stop-replset-member"
+
+
+class RollingOpsCallbackId(StrEnum):
+    """Identifier of the callbacks used for rollingops."""
+
+    RESTART_CHARM_SERVICES = "restart-charm-services"
+    SHARD_RESTART_ON_KEYFILE_CHANGED = "shard-restart-on-keyfile-change"
+    UPDATE_MONGOS_AND_RESTART = "update-mongos-and-restart"
```
Mehdi-Bendriss, line 141: This is nice

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -240,6 +235,21 @@ def __init__(self, charm: AbstractMongoCharm[MongoDBCharmConfig, MongoDBOperator
             self, self.state, self.substrate, ExternalRequirerRelations.VAULT
         )
 
+        stop_replset_sync_lock = StopReplsetSyncLockBackend(self.state)
+        self.rollingops_manager = RollingOpsManager(
+            charm=charm,
+            peer_relation_name=PeerRelationNames.ROLLINGOPS_PEERS.value,
+            etcd_relation_name=RelationNames.ETCD.value,
+            cluster_id="mongodb",
```
Gu1nness: Please handle cluster id as we said if  you have time, otherwise, in a cleanup PR soon

single_kernel_mongo/managers/sharding.py
```diff
@@ -371,21 +375,23 @@ def add_shard(self, relation: Relation) -> None:
...
+                logger.error(
+                    f"{shard_name} shard does not have the same auth as the config server."
+                )
+                raise ShardAuthError(shard_name)
+            logger.warning(f"Unhandled Operation Error {e.code}: {e}")
+            raise
+        except PyMongoError as e:
+            logger.error(f"Failed to add {shard_name} to cluster")
+            raise e
+        except FileNotFoundError as e:
```
Gu1nness, line 392: Same comment on `FileNotFoundError`.

single_kernel_mongo/managers/sharding.py
```diff
@@ -639,6 +653,9 @@ def synchronise_cluster_secrets(self, relation: Relation, leaving: bool = False)
         if self.state.shard_state.shard_integrated:
             self.dependent.s3_backup_manager.configure_and_restart()
 
+        self._reconcile_shard_after_restart()
+
+    def _reconcile_shard_after_restart(self):
```
Gu1nness, line 659: Nit : docstring

single_kernel_mongo/managers/tls.py
```diff
@@ -150,6 +149,80 @@ def get_tls_file_contents(self, internal: bool) -> tuple[str | None, str | None]
 
         return ca_file, pem_file
 
+    def _tls_files_exist_on_workload(self, internal: bool) -> tuple[bool, bool]:
+        """Return whether both TLS files exist on the workload."""
```
Gu1nness: Update docstring to explain the two outputs.

single_kernel_mongo/managers/tls.py
```diff
@@ -150,6 +149,80 @@ def get_tls_file_contents(self, internal: bool) -> tuple[str | None, str | None]
...
+        ca = self.state.tls.get_secret(internal, SECRET_CA_LABEL)
+        chain = self.state.tls.get_secret(internal, SECRET_CHAIN_LABEL)
+        key = self.state.tls.get_secret(internal, SECRET_KEY_LABEL)
+        cert = self.state.tls.get_secret(internal, SECRET_CERT_LABEL)
+
+        return bool((chain or ca) and key and cert)
+
+    def _tls_files_differ_on_workload(self, internal: bool) -> bool:
+        """Return whether workload TLS files differ from the current secret state."""
+        expected_ca, expected_pem = self.get_tls_file_contents(internal)
```
Gu1nness, line 191: For safety I would strip here as well "just to be sure" in case storing format changes accidentally.

tests/charms/mongos_k8s_test_charm/charmcraft.yaml
Gu1nness, line 1: Missing etcdctl ?

tests/charms/mongodb_k8s_test_charm/charmcraft.yaml
Gu1nness, line 1: Missing etcdctl ?

Gu1nness (commented):
Looks very good to me with minor comments!

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -1297,21 +1307,68 @@ def restart_charm_services(self, force: bool = False):
...
+    @override
+    def async_restart_charm_services(self, force: bool = False) -> None:
+        self.charm.state.statuses.add(
+            MongoDBStatuses.WAITING_FOR_RESTART.value,
+            scope="unit",
+            component=self.name,
+        )
+        self.rollingops_manager.request_async_lock(
+            callback_id=RollingOpsCallbackId.RESTART_CHARM_SERVICES,
+            kwargs={"force": force},
```
Mehdi-Bendriss, line 1384: what is this flag for?

single_kernel_mongo/managers/sharding.py
Mehdi-Bendriss, line 392: nice

single_kernel_mongo/managers/tls.py
Mehdi-Bendriss: to be deleted

Mehdi-Bendriss (commented):
Thank you Patricia! Good work! 
Looks clean to me, only minor comment (pending CI to pass)

single_kernel_mongo/managers/mongo.py
patriciareinoso: I am not re raising this error. Comment added

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: This is pass along to the `restart_charm_services(self, force: bool = False):`

We will restart if we detect a change in the config file, a change in the certificates (new behavior) or if we pass the force flag. 

This is a legacy flag. With the add of the identification of changes in the certificates we are closer to get rid of  it but we are not yet.

single_kernel_mongo/managers/tls.py
patriciareinoso: they are strip below

tests/charms/mongos_k8s_test_charm/charmcraft.yaml
patriciareinoso: well I did not added because we cannot integrate with etcd

single_kernel_mongo/events/cluster.py
patriciareinoso: Improved docs and exception catching here and other lock request calls

single_kernel_mongo/managers/tls.py
Gu1nness: The `actual_*` are stripped, but not the `expected_*` from what I read ?

Gu1nness (commented):
Final nits on the TLS certs not being stripped and still the cluster_id to take care of

single_kernel_mongo/managers/ldap.py
```diff
@@ -78,14 +78,10 @@ def assert_pass_hook_checks(self) -> None:
...
-        if (
-            self.dependent.name == CharmKind.MONGOD
-            and self.dependent.state.enable_encryption_at_rest
-            and (state := self.dependent.vault_manager.vault_state())  # type: ignore[attr-defined]
-            != VaultConfigurationState.ACTIVE
-        ):
+        if self.dependent.name == CharmKind.MONGOD and self.dependent.vault_manager.is_degraded():
             logger.warning(
-                f"Encryption at rest may be degraded. Vault agent state: {state.value}. This must be fixed first."
+                "Encryption at rest may be degraded. Vault agent state: %s. This must be fixed first.",
```
Gu1nness: This computes the state twice which can be a tiny bit expensive. I think it would be great if it would return the status or None and we check on "not None"

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -829,6 +828,28 @@ def new_leader(self) -> None:
...
+        if not self.model.unit.is_leader():
+            return
+
+        if not (
+            self.state.is_role(MongoDBRoles.CONFIG_SERVER)
+            or self.state.is_role(MongoDBRoles.REPLICATION)
+        ):
+            return
+
+        if not self.state.app_peer_data.cluster_id:
```
Gu1nness: Is this safe to keep in a regular app databag ? Or should we keep it in a secret?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -829,6 +828,28 @@ def new_leader(self) -> None:
                 password = self.workload.generate_password()
                 self.state.set_user_password(user, password)
 
+    def generate_cluster_id(self):
```
Gu1nness: Maybe this function could also return the value after generating ? Or just generate and then it is stored by the caller ? I know we have a similar patterns in other places and I'm just realizing how confusing that could be for a newcomer that a function saying "generate" actually generate and stores but does not return the value.

single_kernel_mongo/state/charm_state.py
```diff
@@ -605,6 +605,19 @@ def config_server_uri(self) -> str | None:
             return None
         return f"{self.app_peer_data.replica_set}/{self.unit_peer_data.internal_address}:{MongoPorts.MONGODB_PORT.value}"
 
+    @property
+    def cluster_id(self) -> str | None:
+        """Returns the cluster ID."""
+        if self.charm_role.name == CharmKind.MONGOS:
+            return self.cluster.cluster_id
```
Gu1nness: Maybe mongos/shards should persist the cluster id in its app databag for safety reasons (nothing breaks if the relation is suspended). We do that already for a few values, I think it makes sense to do it here as well. So that from restart you can immediately be part of the cluster before even you get all the events from relations.

I think it would make this accessor way easier and that way all our configurations work the same way when they get/store the cluster id.

single_kernel_mongo/managers/cluster.py
```diff
@@ -128,7 +128,7 @@ def share_secret_to_mongos(self, relation: Relation, initial_event: bool = False
         if ldap_user_to_dn_mapping := self.state.ldap.ldap_user_to_dn_mapping:
             relation_data[ClusterStateKeys.LDAP_USER_TO_DN_MAPPING.value] = ldap_user_to_dn_mapping
 
-        relation_data[ClusterStateKeys.CLUSTER_ID] = self.state.cluster_id
+        relation_data[ClusterStateKeys.CLUSTER_ID] = self.state.cluster_id or ""
```
Gu1nness: Suggestion:
```
if self.state.cluster_id:
	relation_data[……] = self.state.cluster_id
```

Don't set to empty string that will end up in weird processing later and could break (we never know if it won't end up being an empty string in the databag if ops breaks)?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -841,6 +841,9 @@ def generate_cluster_id(self):
         if not self.model.unit.is_leader():
             return
 
+        if self.substrate == Substrates.K8S:
+            return
```
Gu1nness: Nit: move this above the leader check so that it is the first check to run (exclude from the broadest exclusion pattern to the smallest one)

single_kernel_mongo/managers/cluster.py
patriciareinoso: normally  ops is able to handle it but sure

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: it's a secret now

single_kernel_mongo/state/charm_state.py
patriciareinoso: done

Gu1nness (dismissed):

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #337, opened 2026-05-22
patch: improve rollingops integration test

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [x] Chores / refactoring

## 📝 Description

This PR improves the rollingops integration tests:

1. Integrate all cluster components so that all of them use the same cluster ID
2. Integrate all cluster components with etcd
3. Enable TLS so that a rolling restart is executed
4. Assert that the etcd rollingops log file is not empty

Uses `charmlibs-rollingops` v 1.1.1 after https://github.com/canonical/charmlibs/pull/481

## ✅ Checklist

### DIFF
--- a/tests/integration/mongodb/test_sharding_rollingops.py
+++ b/tests/integration/mongodb/test_sharding_rollingops.py
@@ -9,7 +9,7 @@ from pathlib import Path
 import pytest
 from pytest_operator.plugin import OpsTest
 
-from tests.integration.helpers.common import DEPLOYMENT_TIMEOUT, TIMEOUT
+from tests.integration.helpers.common import DEPLOYMENT_TIMEOUT, TIMEOUT, read_remote_file
 from tests.integration.helpers.rollingops import (
     ETCD_APP_NAME,
     deploy_etcd,
@@ -20,10 +20,7 @@ from tests.integration.helpers.sharding import (
     deploy_cluster_components,
     integrate_sharding_components,
 )
-from tests.integration.helpers.tls import (
-    PEER_TLS_RELATION_NAME,
-    TLS_CERTIFICATES_APP_NAME,
-)
+from tests.integration.helpers.tls import integrate_apps_with_tls
 from tests.integration.helpers.types import Substrate
 
 logger = logging.getLogger(__name__)
@@ -70,6 +67,7 @@ async def test_integrate_shard_with_etcd(ops_test: OpsTest) -> None:
     await integrate_shard_with_etcd(ops_test, CLUSTER_COMPONENTS)
 
     await ops_test.model.wait_for_idle(
+        status="active",
         apps=CLUSTER_COMPONENTS + [ETCD_APP_NAME],
         idle_period=20,
         timeout=DEPLOYMENT_TIMEOUT,
@@ -77,7 +75,9 @@ async def test_integrate_shard_with_etcd(ops_test: OpsTest) -> None:
 
 
 @pytest.mark.abort_on_fail
-async def test_enable_tls_in_shard_using_rolling_ops(ops_test: OpsTest) -> None:
+async def test_enable_tls_in_shard_using_rolling_ops(
+    ops_test: OpsTest, substrate: Substrate
+) -> None:
     """Enable TLS for shard components and verify rolling ops etcd lock acquisition.
 
     Integrates TLS certificates for all cluster components and verifies that
@@ -91,9 +91,7 @@ async def test_enable_tls_in_shard_using_rolling_ops(ops_test: OpsTest) -> None:
          owner=53a5e2f6-7c0e-4017-8a36-c61093fa17ab-config-server4-0]
         __main__: Lock granted using lease 278b9e4abfac811b.
     """
-    for app in CLUSTER_COMPONENTS:
-        await ops_test.model.integrate(TLS_CERTIFICATES_APP_NAME, f"{app}:{PEER_TLS_RELATION_NAME}")
-
+    await integrate_apps_with_tls(ops_test, CLUSTER_COMPONENTS, peer=True, client=False)
     tls_start_time = datetime.now()
 
     await ops_test.model.wait_for_idle(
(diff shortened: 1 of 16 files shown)


tests/integration/mongodb/test_sharding_rollingops.py
```diff
@@ -47,19 +50,44 @@ async def test_build_and_deploy(
...
+        status="active",
         idle_period=20,
         timeout=DEPLOYMENT_TIMEOUT,
-        raise_on_blocked=False,
     )
 
-    await integrate_sharding_components(ops_test)
+
+@pytest.mark.abort_on_fail
+async def test_integrate_shard_with_etcd(ops_test: OpsTest) -> None:
```
Mehdi-Bendriss, line 61: missing docstrings in this test and the following test

tests/integration/mongodb/test_sharding_rollingops.py
```diff
@@ -47,19 +50,44 @@ async def test_build_and_deploy(
...
+    await ops_test.model.wait_for_idle(
+        apps=CLUSTER_COMPONENTS, idle_period=20, timeout=TIMEOUT, status="active"
+    )
+
+    log_file_path = "/var/lib/rollingops/etcd_rollingops_worker.log"
+
+    for app_name in CLUSTER_COMPONENTS:
+        app = ops_test.model.applications[app_name]
+        for unit in app.units:
+            ssh_command = ["ssh", unit.name, "sudo", "cat", log_file_path]
```
Mehdi-Bendriss: i'm a bit unclear about what we're testing here, if the goal is to confirm that a rolling operation is happening in etcd, why not query etcd itself and confirm?

Mehdi-Bendriss (commented):
Thanks Patricia! I have a couple of comments

tests/integration/mongodb/test_sharding_rollingops.py
patriciareinoso: If the unit does not use etcd at all the etcd log file is never created. 
So a simple check is to see if the file exists.

Checking directly on etcd its way more complex.

To query etcd we need:
1. Extract the cluster ID from the application secrets
2. ssh to the unit machine to parse the etcd config file and export/extract the values needed to connect to etcd
4. Build the query path using the charm's directory

Then, we do not know the exact moment requests are made to ectd. So if we do not get the check right the test would be unstable.

We can query etcd from a single machine until it sees a lock request for each unit in the cluster
If it is not found -> keep trying until timeout. 
If times out -> the test fails.

So checking the log file is straight forward. What I would improve is to check the content to find something like "Lock granted". This way the test would stay valid if for any reason the behavior of the etcd log file changes in the library (for example, if it is created and has content even if etcd is not actually used).

Mehdi-Bendriss: Okay, let's add the "lock granted" grep + timestamp logic so we don't fetch older locks if granted?

Mehdi-Bendriss (dismissed):

tests/integration/mongodb/test_sharding_rollingops.py
```diff
@@ -47,19 +51,83 @@ async def test_build_and_deploy(
...
+
+    Establishes the etcd integration for all cluster components and verifies
+    that all components reach active status.
+    """
+    await integrate_shard_with_etcd(ops_test, CLUSTER_COMPONENTS)
+
     await ops_test.model.wait_for_idle(
-        apps=CLUSTER_COMPONENTS,
-        status="active",
+        apps=CLUSTER_COMPONENTS + [ETCD_APP_NAME],
```
Gu1nness, line 71: Why not waiting for active here?

tests/integration/mongodb/test_sharding_rollingops.py
```diff
@@ -47,19 +51,83 @@ async def test_build_and_deploy(
...
+    with a timestamp after TLS enablement began.
+
+    Log entry format example:
+        2026-05-21 13:36:24,692 [INFO] [77172]
+        [unit=config-server4/0 cluster=7sSW9jUx
+         owner=53a5e2f6-7c0e-4017-8a36-c61093fa17ab-config-server4-0]
+        __main__: Lock granted using lease 278b9e4abfac811b.
+    """
+    for app in CLUSTER_COMPONENTS:
+        await ops_test.model.integrate(TLS_CERTIFICATES_APP_NAME, f"{app}:{PEER_TLS_RELATION_NAME}")
```
Gu1nness: I think we have already a helper to do that.
```
from tests.integration.helpers.tls import integrate_apps_with_tls
await integrate_apps_with_tls(ops_test, CLUSTER_COMPONENTS, client=False)
```

tests/integration/mongodb/test_sharding_rollingops.py
```diff
@@ -47,19 +51,83 @@ async def test_build_and_deploy(
...
+        apps=CLUSTER_COMPONENTS, idle_period=20, timeout=TIMEOUT, status="active"
+    )
+
+    log_file_path = "/var/lib/rollingops/etcd_rollingops_worker.log"
+    lock_message = "Lock granted using lease"
+
+    for app_name in CLUSTER_COMPONENTS:
+        app = ops_test.model.applications[app_name]
+        for unit in app.units:
+            ssh_command = ["ssh", unit.name, "sudo", "cat", log_file_path]
```
Gu1nness: nit: for the future, we should make a helper out of this.
We have some for copying to the file system but since we do use this in a few tests, that should be a helper landing in common.py

Gu1nness (commented):
Small nitpicks only

tests/integration/mongodb/test_sharding_rollingops.py
patriciareinoso: Added the helper and a [ticket](https://warthogs.atlassian.net/browse/DPE-10257) so that we keep the refactor of the other tests using it in mind

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #351, opened 2026-06-08
patch(DPE-10317): Don't request certs for unix socket

Gu1nness:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

Mongos charm was requesting a certificate for a local unix socket.
While this was possible with self-signed-certificates operator, is will fail with real life operators.
This PR fixes this, removing the unix socket from the CSR.
It also updates the mongo config to remove the wrongly used mTLS configuration, and adds an exception for pymongo for the unix socket case (remove host validation).

## 🧪 Manual testing steps

```text
1. juju deploy config-server
2. juju deploy shard
3. juju deploy mongos
4. juju deploy continuous-writes --config external-connectivity=false
5. juju deploy self-signed certificates
6. juju integrate shard:sharding config-server
7. juju integrate mongos continuous-writes
8. Integrate everything with TLS
9. juju integrate mongos config-server
10. juju run continuous-writes start-continuous-writes
11. juju run continuous-writes stop-continuous-writes
```
The last step should return a valid number of writes.

## 🔬 Automated testing steps

test_tls_inverted.py has been updated to reflect that situation.

## ✅ Checklist

- [x] I have added or updated any relevant documentation.

### DIFF
--- a/single_kernel_mongo/utils/mongo_config.py
+++ b/single_kernel_mongo/utils/mongo_config.py
@@ -60,29 +60,34 @@ class MongoConfiguration:
         return self.hosts
 
     @property
-    def formatted_replset(self) -> dict:
+    def formatted_replset(self) -> dict[str, str]:
         """Formatted replicaSet parameter."""
         if self.replset:
             return {"replicaSet": quote_plus(self.replset)}
         return {}
 
     @property
-    def formatted_auth_source(self) -> dict:
+    def formatted_auth_source(self) -> dict[str, str]:
         """Formatted auth source."""
         if self.database != "admin":
             return ADMIN_AUTH_SOURCE
         return {}
 
     @property
-    def tls_config(self) -> dict:
+    def tls_config(self) -> dict[str, str]:
         """TLS Config."""
         if not self.tls_enabled:
             return {}
-        return {
+        config = {
             "tls": "true",
-            "tlsCertificateKeyFile": f"{self.tls_external_keyfile}",
             "tlsCaFile": f"{self.tls_external_ca}",
         }
+        # Edge case: MongoDB TLS with Unix Socket requires disabling client hostname verification.
+        # This is because the reported remote is the unix socket, which is not in the SANS list.
+        if len(self.hosts) == 1 and list(self.hosts)[0].endswith("27018.sock"):
+            config["tlsAllowInvalidHostnames"] = "true"
+
+        return config
 
     def _uri(self, tls: bool):
         if self.port == MongoPorts.MONGOS_PORT and self.replset:
--- a/tests/integration/applications/continuous_write_charm/src/charm.py
+++ b/tests/integration/applications/continuous_write_charm/src/charm.py
@@ -67,7 +67,7 @@ class ContinuousWritesApplication(CharmBase):
         # Database related events
         self.database = DatabaseRequires(self, "mongodb", self.database_name)
         # Database related events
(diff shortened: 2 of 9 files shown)


single_kernel_mongo/utils/mongo_config.py
```diff
@@ -60,29 +60,34 @@ def formatted_hosts(self) -> set[str]:
...
             return {}
-        return {
+        config = {
             "tls": "true",
-            "tlsCertificateKeyFile": f"{self.tls_external_keyfile}",
             "tlsCaFile": f"{self.tls_external_ca}",
         }
+        # Edge case: MongoDB TLS with Unix Socket requires disabling client hostname verification.
+        # This is because the reported remote is the unix socket, which is not in the SANS list.
+        if len(self.hosts) == 1 and list(self.hosts)[0].endswith("27018.sock"):
```
patriciareinoso, line 87: so this condition will only be true for the admin mongos user.

Gu1nness: There's only one mongos user in that scenario: Mongos VM without external connectivity.
Since mongos is subordinated, the config-server creates the user and sends the credentials over the network.
Then mongos vm creates the URI (which contains only one host, the socket) which is used by the charm to run the queries.

So yes, and no at the same time. It's not the admin user, it's all mongos users in that scenario (which amounts to 1 user)

patriciareinoso (dismissed):

patriciareinoso (approved):

single_kernel_mongo/utils/mongo_config.py
```diff
@@ -60,29 +60,34 @@ def formatted_hosts(self) -> set[str]:
...
+            config["tlsAllowInvalidHostnames"] = "true"
```
Mehdi-Bendriss, line 88: where do we put it back to false when multiple units?

tests/integration/applications/continuous_write_charm/src/charm.py
```diff
@@ -154,17 +154,29 @@ def _database_config(self) -> dict[str, str]:
...
-                        f"{hosts}/{quote_plus(parsed_uri['database'])}?"
-                        f"{urlencode(params)}"
-                )
+        parsed_uri = parse_uri(uris)
+        params = parsed_uri["options"]
+        params["tls"] = "true"
+        params["tlsCaFile"] = f"{CA_PATH}"
+
+        if any("27018.sock" in host for host, _ in parsed_uri["nodelist"]):
+            params["tlsAllowInvalidHostnames"] = "true"
```
Mehdi-Bendriss, line 163: where do we put it back to false when multiple units?

Mehdi-Bendriss (commented):
Thanks Neha, looks good to me. I have 1 question

single_kernel_mongo/utils/mongo_config.py
Gu1nness: Because we have only ever one host if we're using unix sockets.

tests/integration/applications/continuous_write_charm/src/charm.py
Gu1nness: Because we have only ever one host if we're using unix sockets.

Mehdi-Bendriss (approved):

PR #353, opened 2026-06-15
patch: bug fixing and stabilization

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

This PR includes:
1. TLS refactoring
2. Cluster-ID fixing
3. https://github.com/canonical/mongo-single-kernel-library/pull/354

### Refactoring

The only change on behavior is:
-> Async restart in TLS is force=False now. Since we reconcile the TLS files on the restart callback, we do not need to force it here.

1. `disable_certificates_for_unit()` → `disable_tls()`
4. Move the logic to enable TLS from the TLS event handle to the TLS manager in a function called `enable_tls`
5. `enable_certificates_for_unit()` → `request_charm_restart()`
6. `set_certificates()` → `_set_certificate_secrets()`

Methods taken from the charm state and moved to the TLS manager:

1. `_propagate_ca_secrets()` : centralized CA propagation logic across cluster, config-server, and client relations
2. `_propagate_ca_secret_as_config_server()`: handles CA updates for CONFIG_SERVER role with cluster/sharding relations
3. `_propagate_client_ca_as_replicaset()`: handles CA updates for client relations (REPLICATION role)

###  Cluster-ID fixing

We fix this ERR log:
```
single_kernel_mongo.core.secrets:Non-existing secret Scope.APP:cluster-id was attempted to be removed.
```

The issue was seen in k8s: given that in k8s we do not have cluster ID, the shards and mongos detected a clusterID = None so they tried to removed it. Since the cluster ID did not exist in the first place, we saw the ERR log.
 
Now if we detect a cluster ID = None , we just ignore it. The cluster ID is removed on relation broken.

## 🧪 Manual testing steps

```text
1. juju deploy <my-app>
7. …
```

## ✅ Checklist


single_kernel_mongo/managers/tls.py
```diff
@@ -241,26 +246,66 @@ def reconcile_tls_files(self) -> bool:
 
         return need_restart
 
-    def disable_certificates_for_unit(self, internal: bool):
-        """Disables the certificates on relation broken."""
+    def disable_tls(self, internal: bool):
+        """Disable TLS for the given scope and clear related state.
+
+        - Remove all TLS secrets for the chosen scope
+        - Propagate the removal of the CA secret to related relations
```
Gu1nness: I think this should be done upon restart.
That way, the other applications will always be in sync with the config-server.
Otherwise, the risk is that the client  / shard / mongos tries to connect without using TLS while the config server has not restarted yet.
IMHO: `reconcile_tls_files` should also send the certs to the clients/shards/mongos instead of having the propagation here.

single_kernel_mongo/managers/tls.py
```diff
@@ -282,16 +327,22 @@ def enable_certificates_for_unit(self):
             logger.info("Still waiting for a certificate, delaying restart.")
             return
 
-        self.dependent.async_restart_charm_services(force=True)
+        self.dependent.async_restart_charm_services()
 
+        # TODO: this is not the correct place for this. Now restart is async
         if self.state.is_role(MongoDBRoles.MONGOS):
             # After restarting, we update the certificates for all clients.
             self.dependent.share_connection_info()  # type: ignore[attr-defined]
```
Gu1nness: I'll take care of that but you are right, this should be done in the mongos restart.

Gu1nness (commented):
Good except that we should be sharing certs to clients / shards / mongos only after restart

single_kernel_mongo/events/sharding.py
```diff
@@ -187,3 +242,6 @@ def _on_relation_broken(self, event: RelationBrokenEvent):
             )
             self.dependent.remove_ca_cert_from_trust_store(TrustStoreFiles.PBM)
             logger.info(f"Skipping {str(type(event))}: {str(e)}")
+            return
+
+        self.manager.cleanup_cluster_id()
```
Gu1nness: This means that the cluster id is never cleaned up if the cluster relation was never fully set up,  is that correct ? Or do we want to clean it as well in the `NonDeferrableFailedHookChecksError` handler ?

single_kernel_mongo/managers/tls.py
```diff
@@ -234,26 +240,66 @@ def reconcile_tls_files(self) -> bool:
...
+            private_key,
+        ):
+            logger.error("Received certificate and private key do not match.")
+            return
 
-        self.dependent.async_restart_charm_services(force=True)
+        self._set_certificate_secrets(
+            internal=internal, provider_cert=provider_cert, private_key=private_key
+        )
+        self._propagate_ca_secrets(internal, new_ca=provider_cert.ca.raw)
```
Gu1nness: This should be called from the `finalize_after_restart` on the mongodb/mongos operator side so that it's propagated only after the workload has restarted.

single_kernel_mongo/managers/tls.py
Gu1nness: Sorry I think I forgot to add that, can you do it ?

single_kernel_mongo/managers/tls.py
patriciareinoso: done

single_kernel_mongo/events/sharding.py
patriciareinoso: I separated the exceptions raised to be able to identify the relation broken events so that the app can keep its cluster ID in case of scale down. Tested manually

single_kernel_mongo/managers/tls.py
```diff
@@ -130,9 +136,9 @@ def get_tls_file_contents(self, internal: bool) -> tuple[str | None, str | None]
         """
         scope = TLSType.PEER.value if internal else TLSType.CLIENT.value
         if not self.state.tls.is_tls_enabled(internal):
-            logging.debug(f"{scope} TLS disabled.")
+            logger.debug("%s TLS is disabled.", scope)
             return None, None
-        logging.debug(f"{scope} TLS *enabled*, fetching data for CA and PEM files ")
+        logger.debug("%s TLS *enabled*, fetching data for CA and PEM files ", scope)
```
Gu1nness:
```suggestion
        logger.debug("%s TLS is enabled, fetching data for CA and PEM files ", scope)
```

single_kernel_mongo/managers/tls.py
```diff
@@ -332,27 +395,32 @@ def push_tls_files_to_workload(self, internal: bool) -> None:
...
         )
-        self.state.tls.set_secret(internal, SECRET_KEY_LABEL, private_key)
-        self.state.tls.set_secret(internal, SECRET_CSR_LABEL, csr)
-        self.state.tls.set_secret(internal, SECRET_CERT_LABEL, certificate)
-        self.state.tls.set_secret(internal, SECRET_CA_LABEL, ca)
+        self.state.tls.set_secret(internal, SECRET_CERT_LABEL, provider_cert.certificate.raw)
+        self.state.tls.set_secret(internal, SECRET_CA_LABEL, provider_cert.ca.raw)
         logger.info(
-            f"{TLSType.PEER.value if internal else TLSType.CLIENT.value} certificate secrets updated."
+            "%s certificate secrets updated.",
```
Gu1nness:
```suggestion
            "%s TLS certificate secrets updated.",
```

Gu1nness (dismissed):
Two tiny feedback on the debug logs but that's all

Gu1nness (approved):

Mehdi-Bendriss (approved):
Thanks Patricia! There are too many moving pieces in this PR, understandly so (stabilisation) but it's difficult for me to have full context. 
Approved based on Neha's feedback/approval.

PR #354, opened 2026-06-15
patch(DPE-10361): Charm stabilization

Gu1nness:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

This PR is mostly refactoring and cleanup.

 * Statuses recomputation after restart for a limited set of managers (mongos on mongos, shard on mongod)
 * Mongos config-changed handler optionally restarts.
 * Mongos config-changed has proper error handling.
 * Charm state properly sets the tls_enabled flag when we have the secrets && the certificate is on the file system.
 * TLS keyfile is removed from mongodb configuration objects (we don't do mtls)

## 🧪 Manual testing steps

```text
1. juju deploy <my-app>
2. …
```

## ✅ Checklist


single_kernel_mongo/events/sharding.py
```diff
@@ -144,17 +152,52 @@ def _on_relation_created(self, event: RelationCreatedEvent):
...
+        try:
+            self.manager.update_config_server_certs()
+        except DeferrableFailedHookChecksError as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
+
+    def _synchronize_passwords(self, event: RelationChangedEvent):
+        """Upon receiving the operator and backup user passwords, we want to update them locally."""
+        try:
```
patriciareinoso, line 177: `self.manager.synchronize_user_passwords(event.relation)` -> `self.sync_cluster_passwords(operator_password, backup_password)` ->  `self.dependent.s3_backup_manager.configure_and_restart() `

and `s3_backup_manager.configure_and_restart` can raise a `WorkloadServiceError`

single_kernel_mongo/events/sharding.py
```diff
@@ -144,17 +152,52 @@ def _on_relation_created(self, event: RelationCreatedEvent):
...
+    def _synchronize_member_auth(self, event: RelationChangedEvent):
+        """When we receive a new keyfile /TLS CA we want to restart mongodb with the right files."""
         try:
-            self.manager.synchronise_cluster_secrets(event.relation)
+            self.manager.synchronize_member_auth(event.relation)
         except (
             DeferrableFailedHookChecksError,
+            NotReadyError,
             WaitingForSecretsError,
+            FailedToUpdateCredentialsError,
```
patriciareinoso, line 197: I believe `FailedToUpdateCredentialsError` is not raised here

single_kernel_mongo/managers/sharding.py
```diff
@@ -651,24 +685,30 @@ def synchronise_cluster_secrets(self, relation: Relation, leaving: bool = False)
...
-        """Reconcile shard state after a unit restart.
+        # Always try to restart, forced if the certs are updated.
+        self.dependent.s3_backup_manager.configure_and_restart(force=must_restart)
 
-        This method resets the shard status to active/idle and, if executed on
-        the leader unit, propagates the necessary shard metadata to allow the
-        config-server to add the shard to the cluster.
-        """
-        # By setting the status we ensure that the former statuses of this component are removed.
         self.state.statuses.set(ShardStatuses.ACTIVE_IDLE.value, scope="unit", component=self.name)
```
patriciareinoso: shouldn't this be in the `reconcile_shard_after_restart`?

single_kernel_mongo/events/sharding.py
```diff
@@ -144,17 +152,52 @@ def _on_relation_created(self, event: RelationCreatedEvent):
...
+            RollingOpsNoRelationError,
             WaitingForCertificatesError,
+        ) as e:
+            defer_event_with_info_log(logger, event, str(type(event)), str(e))
+        except NonDeferrableFailedHookChecksError as e:
+            logger.info(f"Skipping {str(type(event))}: {str(e)}")
+
+    def _handle_pbm_restarts(self, event: RelationChangedEvent):
+        """If everything is working and we're added to the cluster, we want to finally start PBM."""
+        try:
```
patriciareinoso, line 207: PBM restart also raises a `WorkloadServiceError`

single_kernel_mongo/events/sharding.py
patriciareinoso, line 220: PBM restart also raises a `WorkloadServiceError` which is not handled here

single_kernel_mongo/managers/sharding.py
```diff
@@ -745,29 +787,37 @@ def handle_secret_changed(self, secret_label: str | None) -> None:
                 )
             self.sync_cluster_passwords(operator_password, backup_password)
 
-        self.update_pbm_certificate_in_trust_store()
+        if self.update_pbm_certificate_in_trust_store():
+            try:
+                # after updating the password of the backup user, restart pbm with correct password
```
patriciareinoso: the restart after the backup user changed happens in `sync_cluster_passwords`, this one happens if the PBM certs changed.

patriciareinoso: Also, would it be worth it to find a way to just restart PBM once here?

single_kernel_mongo/state/charm_state.py
patriciareinoso, line 722: I do not see any use of `mongodb_config_for_user` that includes  `tls_external_ca`

single_kernel_mongo/managers/mongos_operator.py
```diff
@@ -510,16 +538,9 @@ def share_connection_info(self):
             return
         if not self.charm.unit.is_leader():
             return
-        try:
-            self._share_configuration()
-        except PyMongoError as e:
```
patriciareinoso, line 515: removed because here we just get information from the state, right ?

patriciareinoso (commented):
Thank you Neha! 
just a couple of nits

single_kernel_mongo/events/sharding.py
Gu1nness: True

single_kernel_mongo/managers/sharding.py
Gu1nness: No because that would not cover the case where the restart is triggered by TLS manager (which is not this callback but the other one.
Overall this is more a final cleanup than anything (after everything is integrated, let's go back to the proper status.
I'm not happy about this one however, maybe we should find a better way to do taht.

single_kernel_mongo/managers/sharding.py
Gu1nness: It's probably worth it,  but I'm careful: if we remove the automated restart from the synchronize cluster passwords, then it means that everytime we use that method, we have to take care to restart afterwards.
TBH I think this is probably the safest solution.

> the restart after the backup user changed happens in WorkloadServiceError

I think you mean "ends up in WorkloadServiceError" ?

single_kernel_mongo/state/charm_state.py
Gu1nness: Not yet (and hopefully won't happen, but I'd rather keep it for consistency (have the same interface as mongos_config_for_user)

single_kernel_mongo/managers/mongos_operator.py
Gu1nness: Yes excatly

single_kernel_mongo/managers/sharding.py
patriciareinoso: I made a mistake in what I wrote. I already corrected it.
I meant that the comment "after updating the password of the backup user, restart pbm with correct password" is not correct. Here we are restarting because the PBM certs changed, not because of the password update

single_kernel_mongo/utils/mongo_config.py
```diff
@@ -67,9 +67,10 @@ def formatted_replset(self) -> dict[str, str]:
     @property
     def formatted_auth_source(self) -> dict[str, str]:
         """Formatted auth source."""
+        result = {"authMechanism": "SCRAM-SHA-256"}
         if self.database != "admin":
-            return ADMIN_AUTH_SOURCE
-        return {}
+            result |= ADMIN_AUTH_SOURCE
+        return result
```
patriciareinoso, line 73: why do we need this change?

Gu1nness: This ensures that we're using SCRAM-SHA-256.
I got a case (probably a bug in MongoDB) where it incorrectly thought we where using X509

Issue #371, opened 2026-07-01
replica set members are `not reachable/healthy` when deployed in aws


patriciareinoso:
Replica set members become unhealthy just after deployment.

In a VM replica set deployment with encryption at rest enabled, two replica set units were added to `rs.conf()` but remained `(not reachable/healthy)`. TCP connectivity from the primary to the secondaries on port 27017 succeeded, and mongod/Vault agent were running. The actual failure was MongoDB internal cluster authentication: the secondaries rejected the primary’s __system authentication because the generated clientSource authentication restriction did not include the primary unit IP.

From the primary we can see the replica sets have been properly set:

```
mongodb [primary] admin> rs.status().members.map(m => ({
...   name: m.name,
...   stateStr: m.stateStr,
...   health: m.health,
...   lastHeartbeatMessage: m.lastHeartbeatMessage
... }))
```
```
[
  {
    name: '172.31.34.161:27017',
    stateStr: 'PRIMARY',
    health: 1,
    lastHeartbeatMessage: ''
  },
  {
    name: '172.31.7.205:27017',
    stateStr: '(not reachable/healthy)',
    health: 0,
    lastHeartbeatMessage: ''
  },
  {
    name: '172.31.29.59:27017',
    stateStr: '(not reachable/healthy)',
    health: 0,
    lastHeartbeatMessage: ''
  }
]
```

```
mongodb [primary] admin> rs.conf()
```
```
{
  _id: 'mongodb',
  version: 3,
  term: 1,
  members: [
    {
      _id: 0,
      host: '172.31.34.161:27017',
      arbiterOnly: false,
      buildIndexes: true,
      hidden: false,
      priority: 1,
      tags: {},
      secondaryDelaySecs: Long('0'),
      votes: 1
    },
    {
      _id: 1,
      host: '172.31.7.205:27017',
      arbiterOnly: false,
      buildIndexes: true,
      hidden: false,
      priority: 1,
      tags: {},
      secondaryDelaySecs: Long('0'),
      votes: 1
    },
    {
      _id: 2,
      host: '172.31.29.59:27017',
      arbiterOnly: false,
      buildIndexes: true,
      hidden: false,
      priority: 1,
      tags: {},
      secondaryDelaySecs: Long('0'),
      votes: 1
    }
  ],
  protocolVersion: Long('1'),
  writeConcernMajorityJournalDefault: true,
  settings: {
    chainingAllowed: true,
    heartbeatIntervalMillis: 2000,
[...]

PR #375, opened 2026-07-07
patch: fix IP restrictions for RS in different subnets (aws)

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

This PR is a fix for https://github.com/canonical/mongo-single-kernel-library/issues/371

If units of a replica set are in different subnets, secondary RS nodes were marked as `unreachable`. This problem is seen with a simple RS deployment in aws. 

The problem is the `clusterIpSourceAllowlist` field in the mongod.conf file which only contains the CIDR of the leader unit. If the other units are in another CIDR,  mongod rejects connections. 
We have the same issue with the auth parameters for the charmed users.

The IPs of all the RS units are included in the `clusterIpSourceAllowlist` now. 

In this PR we delay the start until all the units have published their IPs in the databag. These addresses are written in the `mongod.conf` file, like this all the units can start with the proper values.

In case of scale up:

- Dinamically update the [clusterIpSourceAllowlist](https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-security.clusterIpSourceAllowlist) using `setParameter`. We do not need to restart the unit after setting this value. This actions must be executed by every unit: leader and not-leader
- Update the `clusterIpSourceAllowlist` section in the `mongod.conf` file. This is a protection on unexpected restart
- Update the `authenticationRestrictions` of the `charmed-logrotate`, `charmed-backups` and `charmed-stat` users, to include a the peers IP

In case of scale down:

- Same as for scale up but we remove the IP of the departing unit.
These actions are also performed in self healing

We use the `setParameter` approach to avoid restarting on each peer changed.

## 🧪 Manual testing steps

```
Name   Space ID  Subnets       
alpha  0         172.31.0.0/20 
                 172.31.16.0/20
                 172.31.32.0/20
                 252.0.0.0/12  
                 252.16.0.0/12 
                 252.32.0.0/12
```

1. Deploy a RS with 3 units in AWS
2. All the units should be active/idle

check the content of `/var/snap/charmed-mongodb/168/etc/mongod/mongod.conf`

```
security:
  authorization: enabled
  clusterAuthMode: keyFile
  clusterIpSourceAllowlist:
  - 172.31.0.0/20
  - 172.31.31.30
  - 252.0.0.0/12
```

The `clusterIpSourceAllowlist` should contain the IP of the other units outside of it CIDR

Use `juju exec -u rs3/1 -- network-get database-peers --format yaml` to check the addresses

```
bind-addresses:
- mac-address: 06:62:e9:ea:6a:b3
  interface-name: enp39s0
  addresses:
  - hostname: ""
    value: 172.31.11.160
    cidr: 172.31.0.0/20
    address: 172.31.11.160
  macaddress: 06:62:e9:ea:6a:b3
  interfacename: enp39s0
- mac-address: 56:d8:22:68:bd:d8
  interface-name: fan-252
  addresses:
  - hostname: ""
    value: 252.11.160.1
    cidr: 252.0.0.0/12
    address: 252.11.160.1
  macaddress: 56:d8:22:68:bd:d8
  interfacename: fan-252
egress-subnets:
- 172.31.11.160/32
ingress-addresses:
- 172.31.11.160
- 252.11.160.1
```

juju ssh into the unit and get the pbm URI
```
 sudo snap get charmed-mongodb
```
```
pbm-uri      mongodb://charmed-backup:<password>@localhost:27017/?authMechanism=SCRAM-SHA-256&authSource=admin
```

Check the RS and user auth

```
sudo snap run charmed-mongodb.mongosh 'mongodb://charmed-backup:<password>@localhost:27017/?authMechanism=SCRAM-SHA-256&authSource=admin'
```

```
rs3 [direct: secondary] test> use admin
switched to db admin
rs3 [direct: secondary] admin> db.system.users.find({user:"charmed-backup"}).pretty()
[
  {
    _id: 'admin.charmed-backup',
    userId: UUID('5cf1bb4b-d193-430f-91c5-a411d211032e'),
    user: 'charmed-backup',
    db: 'admin',
    credentials: {
      'SCRAM-SHA-256': {
        iterationCount: 15000,
        salt: 'IbZcp1FSHYQp2YbZ+HIcpIspzftoBzjb5eEbMA==',
        storedKey: 'GEdW+1QhgEfXt5f07RtGPcveFbUj2zA/Et8/zldq/M8=',
        serverKey: 'JPgtMRtTa2mHTNOhTeK+XiOxPceFJYauwx0HqID1vm4='
      }
    },
    authenticationRestrictions: [
      { clientSource: [ '127.0.0.1' ], serverAddress: [ '127.0.0.1' ] },
      {
        clientSource: [ '172.31.0.0/20',  '172.31.31.30', '252.32.0.0/12' ],
        serverAddress: [ '172.31.0.0/20',  '172.31.31.30', '252.32.0.0/12' ]
      }
    ],
    roles: [
      { role: 'readWrite', db: 'admin' },
      { role: 'clusterMonitor', db: 'admin' },
      { role: 'backup', db: 'admin' },
      { role: 'restore', db: 'admin' },
      { role: 'pbmAnyAction', db: 'admin' }
    ]
  }
]
```

using `db.adminCommand({ getParameter: 1, clusterIpSourceAllowlist: 1 })` we can check the actual value for the `clusterIpSourceAllowlist`

## 🔬 Automated testing steps

Updated UTs
Integration tests must pass
More tests will be added in the sharded cluster fix

## ✅ Checklist


single_kernel_mongo/utils/network_helpers.py
```diff
@@ -52,6 +52,19 @@ def get_host_public_ip() -> set[str]:
     return {output.stdout.strip()}
 
 
+def merge_cidrs(cidr_list: Sequence[str]) -> list[str]:
```
Gu1nness: For this but maybe for others, I would rather use sets than, or ensure that those lists are always merged to be sure that we don't end up in a situation where we're asking for a restart because the list is presented in a different order.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -563,8 +564,8 @@ def prepare_for_startup(self) -> None:
 
         # Publish this unit's address before waiting for every planned unit.
         self.update_ips_in_databag()
-        if len(self.state.peer_database_addresses) < self.state.planned_units:
-            raise ContainerNotReadyError("Waiting for peer database addresses")
+        if len(self.state.peer_database_addresses) + 1 < self.state.planned_units:
+            raise NotReadyError("Waiting for peer database addresses")
```
Gu1nness: This check is misleading.
I would expect (from the name) that `self.state.peer_database_addresses` contains the addresses of all peers, INCLUDING the current unit. It's apparently not the case. 
Wouldn't it make sense that either : it contains all addresses OR we have one for "other_units" and one for us that we can combine (cf the update_local_auth_restrictions where we combine the same way) ?

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -973,23 +969,58 @@ def update_secrets_and_restart(self, secret_label: str, secret_id: str) -> None:
...
+                # upgrade is occurring.
+                logger.warning(
+                    "Removing replicas during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
+                )
+            self.update_hosts()
+            self.mongo_manager.update_users_local_auth_restrictions()
+
+        self._sync_cluster_ip_source_allowlist(departing_unit=departing_unit)
+
+    def _sync_cluster_ip_source_allowlist(self, departing_unit: Unit | None = None) -> None:
```
Gu1nness: Isn't this needed also for mongos charms ? Or they do not care ?

patriciareinoso: [it looks like](https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-security.clusterIpSourceAllowlist) it is only for mongod.

anyways I will be testing sharded clusters nexts in aws and we will see

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: `self.state.peer_database_addresses` initially returned all the IPs including its own. 

I am also using the function to build the allowed IP list, so I was having a CIDR + IPs included in the CIDR in the list, which was disturbing to me, so I wanted to remove the own's unit IP.

Turns out that if we are not in aws we will probably have the "CIDR + IPs included in the CIDR" behavior anyways so we can just keep the unit's own IP in the list

single_kernel_mongo/managers/config.py
```diff
@@ -544,6 +545,27 @@ def cluster_ips(self) -> dict[str, Any]:
...
+        current_security_config = current_config_file_content.get("security") or {}
+        current_allowlist = current_security_config.get("clusterIpSourceAllowlist")
+
+        if current_allowlist == new_allowlist:
+            return
+
+        logger.info("Cluster IP source allowlist changed. Writing the new config.")
+        current_config_file_content.setdefault("security", {})
+        current_config_file_content["security"]["clusterIpSourceAllowlist"] = new_allowlist
+        self.workload.write(self.file, safe_dump(current_config_file_content))
```
Gu1nness, line 561: I'm wondering if we need to check if it's different, or always overwrite the file?
It does not seem very expensive to always overwrite the file, nor does it seem a risky operation? We're not deciding if we should trigger a restart here.

single_kernel_mongo/managers/mongo.py
```diff
@@ -207,6 +207,28 @@ def initialise_user(self, user: MongoDBUser):
...
+        with MongoConnection(self.state.mongo_config) as mongo:
+            for user in (CharmedStatsUser, CharmedBackupUser, CharmedLogRotateUser):
+                if not self.state.app_peer_data.is_user_created(user.username):
+                    continue
+
+                config = self.state.mongodb_config_for_user(
+                    user, auth_restrictions=self.state.local_auth_restrictions
+                )
+                logger.info("Updating auth restrictions for %s user.", user.username)
+                mongo.update_user_auth_restrictions(config)
```
Gu1nness: Beware: if we're updating the auth restrictions of the operator user, I'm not sure if the connection could get closed ? Maybe it's easier here to open the connection once per user ?
I'm not sure but it would be great to know.

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness, line 1: I think on config-changed we should update the local auth user restrictions if it's not done yet.
This is because in case of IP change, a config-changed event is trigerred.

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -963,23 +969,58 @@ def update_secrets_and_restart(self, secret_label: str, secret_id: str) -> None:
...
+                # incorrect and lead to MongoDB reporting that the replica set is unhealthy,
+                # we should make an attempt to fix the replica set configuration even if an
+                # upgrade is occurring.
+                logger.warning(
+                    "Removing replicas during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
+                )
+            self.update_hosts()
+            self.mongo_manager.update_users_local_auth_restrictions()
+
+        self._sync_cluster_ip_source_allowlist(departing_unit=departing_unit)
```
Gu1nness: We need to be sure if this is only on VM or also on K8S.
Juju spaces don't apply on K8S and IIRC the config does not add the cluster ip list fields on K8S, but we need to double check

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -963,23 +969,58 @@ def update_secrets_and_restart(self, secret_label: str, secret_id: str) -> None:
...
+        if departing_unit:
+            try:
+                departing_address = self.state.peer_unit_data(departing_unit).database_address
+            except (KeyError, ModelError):
+                departing_address = ""
+
+            if departing_address:
+                allowlist = [entry for entry in allowlist if entry != departing_address]
+
+        self.config_manager.sync_cluster_ip_source_allowlist_to_file(allowlist)
```
Gu1nness, line 1032: Please ensure that all lists are sorted to ensure that we don't update unnecessarily the config file.
It's a good habit to keep with lists in charms, to ensure that they are sorted everytime they are written / compared as it will make it easier to ensure equality / comparison / immutability.

Gu1nness (commented):
I like this implementation.
Let's see how it holds up to sharded deployments, but for now it does seem to do the trick and to do it cleanly!

single_kernel_mongo/managers/config.py
patriciareinoso: updated: we always write the file

single_kernel_mongo/managers/mongo.py
patriciareinoso: Updated: we use one connection per user now. 
We only update `CharmedStatsUser`, `CharmedBackupUser`, `CharmedLogRotateUser`, the operator user does not have restrictions.

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: the IP list is added to the config file in k8s so we always update in VM and k8s

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: the list is always ordered and duplicates are removed

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: on config-changed the config file was updated during the restart. 
now we dynamically update both the file and the users, and we remove a restart

single_kernel_mongo/managers/mongodb_operator.py
Gu1nness: Oh yeah, of course.
Because we don't have the juju spaces we'll get the same binding everywhere and that will just work out

single_kernel_mongo/managers/mongodb_operator.py
```diff
@@ -709,9 +710,8 @@ def update_config_and_restart(self) -> None:
                 f"Migration of sharding components not permitted, revert config role to {self.state.app_peer_data.role.value}"
             )
 
-        # If we had an IP change, we must restart.
-        if self.state.db_initialised:
-            self.async_restart_charm_services(force=False)
+        # An IP change requires the cluster network access restrictions to be refreshed.
+        self._sync_cluster_network_access_restrictions()
```
Gu1nness: this needs to check if `self.state.db_initialised`.
Probably we'll never end up in the situation where it's not the case, but better safe than sorry.

Gu1nness (dismissed):
One nit but approved already on the rest of the implementation.

single_kernel_mongo/managers/mongodb_operator.py
patriciareinoso: inside `_sync_cluster_network_access_restrictions` we have that check

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #382, opened 2026-07-21
patch: (fix) update cluster IP source allowlist in sharded cluster (aws)

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [ ] Chores / refactoring

## 📝 Description

This PR fixes https://github.com/canonical/mongo-single-kernel-library/issues/381

This PR builds on top of https://github.com/canonical/mongo-single-kernel-library/pull/375 , which adds the replica set addresses to the `clusterIpSourceAllowlist` and `authenticationRestrictions` on config-changed (IP changed), application scale up and down).

Now we are solving the same problem but for a sharded cluster.

When deployed in aws, units can be deployed in separate subnets. So we need to properly update the [clusterIpSourceAllowlist](https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-security.clusterIpSourceAllowlist) :
- in the config-server to include the shard hosts
- in the shards to include all the mongos/config-server addresses.

We also need to properly update the `authenticationRestrictions` for the charmed users. 

This values are updated when the shard is added and when it is removed. and only for VM mongoDB.

## 🧪 Manual testing steps

1. Deploy a sharded cluster with
- config server (3 units)
- shard0 (3 units)
- shard1 (3 units)

check the content of 

`sudo cat /var/snap/charmed-mongodb/168/etc/mongod/mongod.conf`

it should include all the IPs of the 9 units for the config server and 6 units for each shard.

remove the relation between the config-server and a shard and check that now we the config server points to 6 units

## 🔬 Automated testing steps

UTs
- test_sync_cluster_network_access_restrictions_excludes_departed_addresses
- test_config_server_local_auth_restrictions_include_shard_hosts (charm user auth)
- test_shard_local_auth_restrictions_include_config_server_hosts  (charm user auth)
- test_config_server_cluster_ips_include_shard_rs_hosts (config file)
- test_shard_cluster_ips_include_config_server_hosts  (config file)
- test_reconcile_shards_syncs_cluster_network_access_when_adding
- test_reconcile_shards_removes_network_access_after_draining

Integration tests:
- tests/integration/mongodb/ha/test_ha.py : check that the replica set IPs are included in the mongod.conf (include scale up and down)
- tests/integration/mongodb/test_sharding.py : check that the config-server config file includes the shard's addresses and the the shard's config file includes the config-server addresses (include shard removal)

## ✅ Checklist


Gu1nness (commented):
Overall that looks good, but I would love to have that tested.
We could add one small test to the sharding tests to ensure that the list of IPs in the cluster source ip allow list is the one we expect.
We could also maybe add a check in the ha tests ?

Also: about the PR description, the values are updated also on config-changed I think

single_kernel_mongo/managers/sharding.py
```diff
@@ -102,7 +102,10 @@ def prepare_sharding_config(self, relation: Relation) -> None:
         """Handles the database requested event.
 
         It shares the different credentials and necessary files with the shard.
+
         """
+        if not self.charm.unit.is_leader():
+            raise NonDeferrableFailedHookChecksError("Not leader")
```
Gu1nness: This already can run only on the leader:
 * It is called from the handler of the database requested custom event
 * The database-requested custom event can only be fired on the leader: [see here](https://github.com/canonical/mongo-single-kernel-library/blob/8/edge/single_kernel_mongo/lib/charms/data_platform_libs/v0/data_interfaces.py#L3553-L3571)

single_kernel_mongo/managers/sharding.py
```diff
@@ -159,11 +162,19 @@ def reconcile_shards_for_relation(self, relation: Relation, is_leaving: bool = F
...
+                self.dependent.sync_cluster_network_access_restrictions(
+                    excluded_addresses=set(shard_hosts)
+                )
             else:
-                self.add_shard(relation)
+                logger.info("Adding shard %s to cluster.", relation.app.name)
+                self.dependent.sync_cluster_network_access_restrictions()
+                if self.charm.unit.is_leader():
+                    self.add_shard(relation)
         except NotDrainedError:
```
Gu1nness, line 170: Two things:
 * Either sync and then add/remove or add/remove and then sync but not a mix of both according to the branch.
 * I'm wondering if we shouldn't be a bit smarter to reduce the nesting: one method that adds/removes and one that synchronizes the cluster access restrictions? That would make a code that is less convoluted, and where the preconditions such as `self.charm.unit.is_leader()` are tested early in the function instead of deeply in the decision tree. Either the two functions called from the `reconcile_shards_for_relation` or directly from the relation_changed handler:
```python3
def _on_relation_event(self, event: RelationChangedEvent) -> None:
    """Handle relation changed and relation broken events."""
    is_leaving: bool = isinstance(event, RelationBrokenEvent)
    try:
        self.manager.state.statuses.delete(
            status=ConfigServerStatuses.MISSING_CONF_SERVER_REL.value,
            scope="unit",
            component=self.manager.name,
        )
        # Add / Remove shard from cluster
        self.manager.reconcile_shards_for_relation(event.relation, is_leaving)

        # Update local auth restrictions
        # Note: update the method to be public ? 
        shard_hosts = self.manager._get_shard_hosts_from_relation(relation) if is_leaving else None
        self.manager.sync_cluster_network_access_restrictions(excluded_addresses=shard_hosts)
    except (
        DeferrableFailedHookChecksError,
        ServerSelectionTimeoutError,
        ShardAuthError,
        NotDrainedError,
        NotReadyError,
        BalancerNotEnabledError,
        PyMongoError,
        OperationFailure,
    ) as e:
        self.manager.state.statuses.add(
            status=ConfigServerStatuses.MISSING_CONF_SERVER_REL.value,
            scope="unit",
            component=self.manager.name,
        )
        defer_event_with_info_log(logger, event, action=str(type(event)), message=str(e))
    except NonDeferrableFailedHookChecksError as e:
        logger.info(msg=f"Skipping {str(type(event))}: {str(e)}")
```

Gu1nness (changes requested):
I think the implementation can be reworked a bit to smoothen it out:
 * Strongest preconditions should be early (is unit leader ?)
 * Reduce nesting and split responsibilities on manager methods.

Gu1nness (approved):
All good for me, thanks !

Mehdi-Bendriss (approved):

PR #405, opened 2026-08-19
patch: fix has_file function to check on every unit instead of only checking the leader

patriciareinoso:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [x] Chores / refactoring

## 📝 Description

This PR make the function `has_file` to actually check the given unit instead of checking the leader unit.
This function is used by:
- test_sharded_backups_tls.py -> test_remove_integration
- test_encryption_valid_scenario.py -> test_integration_goes_to_active

which iterate over all the units of the application and call the `has_file` function

## 🧪 Manual testing steps

No need, this PR just modifies an check in the integration test

## 🔬 Automated testing steps

The integration tests using the `has_file` function still pass

## ✅ Checklist


tests/integration/helpers/common.py
Gu1nness, line 1: This change probably means we want to check that all calls to this function provide the leader as unit, but agreed !

patriciareinoso: the `has_file` function is called twice (test_sharded_backups_tls.py and test_encryption_valid_scenario.py)
Both times the test iterates over all the units of the application and called the `has_file` functions. 
So in fact we were checking multiple times that the leader unit has a file , now we check that each unit in the application has a file

Gu1nness: Oh right, perfect then.

Thanks!

Gu1nness:
Thanks a lot for that!

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #406, opened 2026-08-19
patch(DPE-10868): Better backup configuration

### LINKED ISSUE #403: Improve PBM initial setup
Use [pipeline config](https://docs.percona.com/percona-backup-mongodb/manage/configure-remotely.html) to configure PBM (all the time) so that it does not have to do a two step initial configuration, and it does not leak the creds.

Gu1nness:
## 🏷️ Type of changes

- [ ] Documentation update
- [ ] Tooling and CI
- [x] Chores / refactoring

## 📝 Description

Refactor PBM configuration and synchronisation:
 * Use [remote configuration](https://docs.percona.com/percona-backup-mongodb/manage/configure-remotely.html) to configure PBM, getting rid of the quirky initial setup of PBM.
 * Remove the force_resync_config method (not used anymore, and useless as PBM when needed.

Fix #403 

## 🧪 Manual testing steps

You can run locally the backup_tls tests that will install microceph and everything for you before running the tests.

## 🔬 Automated testing steps

No update needed.

## ✅ Checklist


single_kernel_mongo/managers/backups/common.py
```diff
@@ -552,19 +515,6 @@ def set_config_options(self, credentials: dict) -> None:
             )
             raise SetPBMConfigError(err.stderr)
 
-    def clear_pbm_config_file(self) -> None:
-        """Overwrites the PBM config file with the one provided by default."""
-        # Bootstrap the config with blackhole configuration.
-        self.workload.write(
-            self.workload.paths.pbm_config,
```
patriciareinoso, line 559: it looks like the `pbm_config` property is no longer used and can be removed

single_kernel_mongo/managers/backups/common.py
```diff
@@ -552,19 +515,6 @@ def set_config_options(self, credentials: dict) -> None:
...
-            "# this file is to be left empty. Changes in this file will be ignored.\n"
-            + EMPTY_CONFIG,
```
patriciareinoso, line 561: the `EMPTY_CONFIG` can also be removed

single_kernel_mongo/managers/backups/common.py
```diff
@@ -502,41 +478,28 @@ def get_main_status(self) -> StatusObject | None:
     def set_certificate(self, credentials: dict) -> None:
         """Sets the certificate on the file system if needed."""
         # Add certificate to trust store
+        file_on_disk = self.dependent.get_ca_cert_from_trust_store(TrustStoreFiles.PBM)
```
patriciareinoso: If  we move this line inside the IF we wouldnt spend time reading from the workload  unnecessarily

single_kernel_mongo/managers/backups/s3.py
```diff
@@ -189,16 +189,18 @@ def validate_config(self) -> bool:
...
+                logger.info("Missing S3 Config")
+                return False
 
-        if provided_configs.get("storage.type") == "s3" and not provided_configs.get(
-            "storage.s3.endpointUrl"
-        ):
-            logger.info("Missing S3 endpoint.")
-            return False
+            if not storage_s3.get("region"):
+                logger.info("Missing region - this is required for AWS")
```
Mehdi-Bendriss: missing return False?

single_kernel_mongo/managers/backups/common.py
```diff
@@ -503,40 +477,27 @@ def set_certificate(self, credentials: dict) -> None:
...
-                        *[
-                            ("--set", f"{pbm_key}={pbm_value}")
-                            for pbm_key, pbm_value in config.items()
-                        ],
-                    )
-                ),
-                environment=self.environment,
+                "config", ["--file=-"], environment=self.environment, input=safe_dump(config)
             )
+            time.sleep(2)
```
Mehdi-Bendriss, line 499: is this still needed now that we don't have a resync?

single_kernel_mongo/managers/backups/common.py
```diff
@@ -166,46 +161,25 @@ def create_bucket(self, credentials: dict[str, str]):
...
-        return self.BASIC_CONFIG | {
-            self.CONFIG_MAP[s3_option]: s3_value
-            for s3_option, s3_value in credentials.items()
-            if self.CONFIG_MAP.get(s3_option)
-        }
+        output_dict: dict[str, Any] = self.BASIC_CONFIG
+        # Iterate on all configuration values
+        for s3_option, s3_value in credentials.items():
+            # Create a ref to the dict on which we'll walk
+            tmp_dict = output_dict
```
Mehdi-Bendriss, line 170: (on the 2 occurrences) I'm not clear about why not work on output_dict directly and initialize it with a copy of Basic_config?

Mehdi-Bendriss (commented):
Thanks Neha, great work! only 3 comments / questions from my side

single_kernel_mongo/managers/backups/s3.py
Gu1nness: There's a default region (which is supported by PBM) so probably not "mandatory" but I agree that the log message is misleading.
I always prefer to specify everything including the region, however it might not be the best UX (people would expect us-east-1 by default).

WDYT?

single_kernel_mongo/managers/backups/common.py
Gu1nness: I think it's good to ensure that the command was taken into account before running the PBM status.
The resync is done by PBM automatically when configuring the metadata that do need a resync.
This was the reason why I left it as is.

single_kernel_mongo/managers/backups/common.py
Gu1nness: The dictionary is nested, and in this loop, we walk down the dictionary with the `setdefault` function.
So we create a moving reference `tmp_dict` that we walk on as we create the sub-dictionaries, and we still keep track of the whole dictionary in `output_dict`.
The `basic_config` contains the storage type which is the only part not provided by the configuration from the integrator.

single_kernel_mongo/managers/backups/common.py
```diff
@@ -503,40 +477,27 @@ def set_certificate(self, credentials: dict) -> None:
         """Sets the certificate on the file system if needed."""
         # Add certificate to trust store
         if cert_chain_list := credentials.get("tls-ca-chain", None):
+            file_on_disk = self.dependent.get_ca_cert_from_trust_store(TrustStoreFiles.PBM)
             self.dependent.save_ca_cert_to_trust_store(TrustStoreFiles.PBM, cert_chain_list)
             self.share_certificate_with_shards(cert_chain_list)
             # Restart after setting all configurations
-            self.configure_and_restart(force=True)
+            should_restart = file_on_disk.strip() == "\n".join(cert_chain_list).strip()
```
patriciareinoso: shouldn't they restart if they are different?

single_kernel_mongo/events/backups.py
```diff
@@ -261,7 +258,7 @@ def _on_credentials_changed(  # noqa: C901
             event.defer()
         except WorkloadServiceError:
             backup_state = BackupState.WAITING_PBM_START
-        except (ResyncError, PBMBusyError):
+        except ResyncError:
```
patriciareinoso, line 261: I do not understand why if we are removing thhe resync, we still catch the ResyncError

patriciareinoso: aaahh , because 
"The resync is done by PBM automatically when configuring the metadata that do need a resync." ?

single_kernel_mongo/managers/backups/s3.py
patriciareinoso: I would make the region mandatory.

But if we decide that `us-east-1` is used by default, that can be added to the log message

single_kernel_mongo/managers/backups/common.py
Gu1nness: Oh yeah you're right, I missed that…

single_kernel_mongo/events/backups.py
Gu1nness: Yeah exactly

patriciareinoso (approved):

Mehdi-Bendriss (approved):

